PR #1, opened 2017-02-01
add delete and list operations to manageSSLCertificate.sh script

kshychko:


ml019:
Looks good. In other scripts, Adding a list operation as well was a nice touch

For other scripts, I've started explicitly defining the known commands at the top of the file and using these in the usage/case statements e.g.

CERTIFICATE_OPERATION_LIST="list"
CERTIFICATE_OPERATION_DELETE="delete"
CERTIFICATE_OPERATION_UPLOAD="upload"
CERTIFICATE_OPERATION_DEFAULT="${CERTIFICATE_OPERATION_UPLOAD}"

Just seems better to have an explicit list of supported commands at the top of the file - particularly if the user actually wants to set the CERTIFICATE_OPERATION variables before calling the script.

Would you mind adding this to the script please?

Issue #3, opened 2017-02-10
Break up segment template


ml019:
Currently the segment/solution/application templates are getting quite large and unwieldy. 

In order to make them more manageable, I want to split the support for each slice/resource type into its own freemarker template, and assemble these dynamically, in the same way we are currently assembling the ECS container definitions.

This will also permit new AWS resource types to be dynamically added, allowing a new type to be developed and then migrated into the gen3 repo.

We'll do createSegmentTemplate.ftl as a test of the idea.

Under templates, we will have a directory for segment, solution, and application. We will out each slice into its own file "segment_*.ftl" and save in the segment directory. We will also 

Each file will handle one slice, and will generate the resulting outputs. Each will include a case structure with a "mode" to select between "definition" and "outputs".

If required, we can add a start/end file as well. Finally all the definitions/macros at the start can be moved to their own include as they are mostly reusable across the different template types. We may need to adjust this file slightly to check for variable presence in some cases.

setContext.sh will also need to be modified to assemble the segmentList in the same manner as it does for container definitions, with the presence of a definition at the segment, then solution level overriding the same file in the generation framework. This permits modifications to be tested in a project before incorporating them more generally into the framework.

PR #4, opened 2017-02-12
Break up segment template #3

kshychko:
First commit for #3 
segment_*.ftl files have been added
due to change in aws/createTemplate.sh  aws/templates/createApplicationTemplate.ftl and  aws/templates/createSolutionTemplate.ftl have been added, although at the moment they are just copies of existing template.


aws/templates/solutions/solution_ElastiCache.ftl
```diff
@@ -0,0 +1,122 @@
...
+                                        { "Key" : "cot:segment", "Value" : "${segmentId}" },
+                                        { "Key" : "cot:environment", "Value" : "${environmentId}" },
+                                        { "Key" : "cot:category", "Value" : "${categoryId}" },
+                                        { "Key" : "cot:tier", "Value" : "${tier.Id}" },
+                                        { "Key" : "cot:component", "Value" : "${component.Id}" },
+                                        { "Key" : "Name", "Value" : "${productName}-${segmentName}-${tier.Name}-${component.Name}" } 
+                                    ]
+                                }
+                            }
+                            [#assign count += 1]
```
kshychko: count variable is being set but there is no check for it in Resources(definition mode) section of the template for all solutions. at the same time there is a check for it in Outputs(output mode) section. Seems to be a bug.

aws/templates/createApplicationTemplate.ftl
```diff
@@ -323,209 +323,12 @@
...
-                                        [#list task.Containers?values as container]
-                                            [#if container?is_hash]
-                                                [#include containerList]
-                                            [/#if]
-                                        [/#list]
-                                        [#assign count += 1]
-                                    [/#if]
-                                [/#list]
-                            [/#if]
+
```
kshychko: 'ecs' part has been hardcoded in the file names, should we add an additional outer iteration?

PR #18, opened 2017-02-25
get the quotes back

kshychko:
@ml019 , please review and merge


ml019:
Looks fine. Obviously you are just putting back the ones you deleted. There are others, even in these files, where the parameter expansion is not quoted - I think we address them as we change other things  in the files. I have learnt a lot about bash scripting over the course of developing codeontap and the code reflects that learning journey.

PR #22, opened 2017-02-28
use formatId/formatName

### LINKED ISSUE #21: Refactor code to use formatId/formatName
Move all the id and name assembly boilerplate into these convenience methods. 

getKey does the same as formatId to its inputs so calls to it can be refactored in the same way.

kshychko:
resolves #21 
the first commit for #21


kshychko:
resolves #21 
@ml019 , please review

aws/templates/createAccountTemplate.ftl
```diff
@@ -9,7 +9,7 @@
     [#case "accountInDomain"]
         [#assign accountDomain = accountName + "." + accountDomainStem]
         [#assign accountDomainQualifier = ""]
-        [#assign accountDomainCertificateId = accountDomainCertificateId + "-" + accountId]
+        [#assign accountDomainCertificateId = formatId(accountDomainCertificateId, accountId)]
```
ml019: Think this should be formatName not formatId

aws/templates/createProductTemplate.ftl
```diff
@@ -17,7 +17,7 @@
     [#case "productInHost"]
     [#default]
         [#assign productDomain = productDomainStem]
-        [#assign productDomainQualifier = "-" + productName]
+        [#assign productDomainQualifier = "-" + productDomain]
```
ml019: I don't think this change should be made. Because the switch assumes you want to include the product in the host portion of the domain, it should be the productName not the productDomain that is included

PR #24, opened 2017-03-02
Add support for VPC Endpoints as part of vpc

### LINKED ISSUE #23: Add support for VPC Endpoints as part of vpc
Add support for vpc endpoints as part of the vpc deployment unit for segments.

It will be added automatically by default. We can add a control latter to turn it off if required.

kshychko:
resolves  #23


kshychko:
@ml019 , please review

ml019:
Support added for DIBP. PR no longer valid.

PR #27, opened 2017-10-26
cleanup redirector image name

kshychko:


ml019:
Looks ok. I'm assuming both the old and new name exist on dockerhub. I'll merge this.

PR #31, opened 2017-12-21
Added S3 website hosting to s3

roleyfoley:
Allows a new object to be passed in the definition of a S3 bucket

s3 solution has a new parameter 
website {
index - The root index file 
error - A pretty error page to serve
}

This would configure the S3 bucket as a S3 Web host


roleyfoley:
fixed up then action in the s3_solution and fixed the resource template details

ml019:
Think it might be useful to add a default for the index file - "index.html". That way all we need in the solution file is "Website : {}".

Otherwise this looks good to go.

We will want to look at the api gateway config in the same way. I think it needs a "Publish : {}" attribute that if present, causes the api definition to be published wherever the child config defines. And by default, we will create an s3 website bucket and push there. Suggest we do this as part of a separate PR.

roleyfoley:
Added defaults to the resource definition since, it would pretty much be the same defaults for all s3 website configurations
index defaults to index.html
error defaults to empty as the error document is used for pretty error pages and you might want to see the error pages.

roleyfoley:
Added IP Whitelisting Condition creation 
s3IPAccessCondition which takes a cidr(s) and an allow boolean which is used for switching between whitelisting (default) and blacklisting IP's

aws/templates/common.ftl
```diff
@@ -621,6 +621,18 @@
                         }
                     ]
                 },
+                { 
+                    "Name" : "Website",
+                    "Children" : [
+                        {
+                            "Name": "Index"
```
ml019: Thinking we should put default here rather than on the macro. Given that we will be using this definition to build the "model", better that the defaults are here. 

Can leave on macro and add here as well but thinking the macro ust expect te index to be provided and the default here provides the default value

PR #32, opened 2017-12-29
Add API Gateway Documents Hosting

roleyfoley:
Adds support for a document hosting location for API Gateway API Documentation. Currently a static documentation HTML file is being generated from the swagger spec as part of the build process, but it is only available to the operations S3 buckets. 

This pull requests adds an S3 Static Website to each API Gateway application instance which can be used to host the static documentation. 

The name of each bucket will either be the DNS name of the API Endpoint with "docs." appended to the front or the bucket resource name with docs. appended to the front. 

To ensure that only the sources that can invoke the API can read the Documentation an IP Based condition is added to the bucket policy based on the IP Address Groups in the API Gateway WAF Configuration


aws/templates/application/application_apigateway.ftl
```diff
@@ -424,6 +424,125 @@
...
+                        )
+                    dependencies=docsS3BucketId
+                /]
+
+                [@createS3Bucket
+                    mode=listMode
+                    id=docsS3BucketId
+                    name=docsS3BucketName
+                    websiteConfiguration=docsS3WebsiteConfiguration
+                    outputId=docsS3BucketId
```
ml019: outputId is really only there for legacy support of some situations where the id used for outputs was different to that used to define the resource. Should always have been the same. In this case they are so the outputId isn't needed

aws/templates/application/application_apigateway.ftl
```diff
@@ -424,6 +424,162 @@
...
+                        "   ./temp_apidoc || return $?",
+                        "  #",
+                        "  # Sync to the API Doc bucket",
+                        "  copy_apidoc_file" + " " +  
+                            formatComponentS3Id(
+                                        tier,
+                                        component,
+                                        occurrence,
+                                        "docs") + " " +
+                        " ./temp_spa/apidoc.html",
```
ml019: should this be ./temp_apidoc/apidoc.html ?

PR #35, opened 2018-01-06
EFS support for ECS Cluster Shared storage

roleyfoley:
Elastic File System can be used to provide persistent storage for tasks and services running on a ECS cluster. 
The EFS Volume is mounted to each of the cluster instances as part of their cloud formation init. 
Services and tasks can then use this as a mount point in their configuration and the data will be available to all nodes in the cluster. 

This change contains two EFS usage implementations 

- Provisioning as part of ECS setup 
  -  ECS -  property  _ClusterWideStorage : true_
  - This will create a new EFS volume, add mount points in the same subnets as the ECS cluster, add the mount configuration to the ECS instances as part of their cfn init setup and setup up appropriate security groups

- Dedicated EFS solution 
  - a standalone EFS component can be added to a solution. This will provision the EFS volume and add mount points to the tiers subnets, and create a placeholder security group attached to the mount point. 
  - Mounting of the EFS volume will be added as part of storage configuration.


roleyfoley:
Redoing pull request for refresh

PR #67, opened 2018-02-06
add schedule event rule resource support

kshychko:
a scheduled event rule will be created along with lambda function in order to keep it warm.

an example - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html#w2ab2c21c10d637c13

### DIFF
--- a/aws/templates/application/application_lambda.ftl
+++ b/aws/templates/application/application_lambda.ftl
@@ -207,6 +207,20 @@
                                 component,
                                 fn,
                                 occurrence)]
+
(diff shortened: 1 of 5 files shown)


roleyfoley:
Could we make this a configurable option within the application_lambda?
* Enabled True/False
* Schedule = ScheduleExpression 

We have a requirement for events that trigger a reporting function that needs to be scheduled at a daily level, this would fit that requirement as well.

PR #82, opened 2018-02-12
Add Unit level appsettings lookup for Solution Components

roleyfoley:
Some Solution level components, like userpools need environment and component level properties. These are set in the AppSettings but aren't being applied at the moment.

This PR adds an AppSettings lookup for deployment unit level Appsettings on Solutions

### DIFF
--- a/aws/setContext.sh
+++ b/aws/setContext.sh
@@ -248,6 +248,19 @@ if [[ -n "${PRODUCT}" ]]; then
                 "${SEGMENT_APPSETTINGS_DIR}/${BUILD_DEPLOYMENT_UNIT}"/build*.json \
                 "${SEGMENT_APPSETTINGS_DIR}/${BUILD_DEPLOYMENT_UNIT}"/build*.ref
         fi
+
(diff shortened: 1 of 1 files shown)


aws/setContext.sh
```diff
@@ -248,6 +248,19 @@ if [[ -n "${PRODUCT}" ]]; then
                 "${SEGMENT_APPSETTINGS_DIR}/${BUILD_DEPLOYMENT_UNIT}"/build*.json \
                 "${SEGMENT_APPSETTINGS_DIR}/${BUILD_DEPLOYMENT_UNIT}"/build*.ref
         fi
+
+        # Some solution units require environment/unit level settings.
+        if isValidUnit "solution" "${DEPLOYMENT_UNIT}"; then 
+            export BUILD_DEPLOYMENT_UNIT="${DEPLOYMENT_UNI}"
```
ml019: Think there is a letter missing here?

roleyfoley: Fixed up, ready for review

PR #83, opened 2018-02-12
Add EFS storage support to ECS Clusters

roleyfoley:
Allows for an EFS volume to be created and attached to each node in an ECS cluster. Data is then persisted and shared in the cluster.


aws/templates/solution/solution_ecs.ftl
```diff
@@ -53,6 +53,47 @@
...
+        [#assign ecsEFSVolumeId = formatEFSId(tier, component)]
+        [#assign ecsEFSVolumeName = formatName( tier, component, "efs")]
+        [#assign ecsEFSSecurityGroupId = formatComponentSecurityGroupId( tier, component,"efs")]
+        [#assign ecsEFSIngressSecurityGroupId = formatDependentSecurityGroupIngressId(ecsEFSSecurityGroupId) ]
+
+        [@createComponentSecurityGroup
+            mode=listMode
+            tier=tier
+            component=component 
+            extensions="efs"
```
ml019: not sure we need "efs" extension - more for when we use getOccurrences

aws/templates/solution/solution_ecs.ftl
```diff
@@ -53,6 +53,47 @@
             name=formatComponentLogGroupName(tier, component) /]
     [/#if]
 
+    [#if deploymentSubsetRequired("efs", true) && 
+            ecs.ClusterWideStorage?has_content && ecs.ClusterWideStorage ]
+
+        [#assign ecsEFSVolumeId = formatEFSId(tier, component)]
+        [#assign ecsEFSVolumeName = formatName( tier, component, "efs")]
```
ml019: Doesn't need efs in the name I think - it will be shown in the console under EFS - but it should be the fulkl component name including product/segment

aws/templates/solution/solution_ecs.ftl
```diff
@@ -53,6 +53,47 @@
...
+        [#assign ecsEFSIngressSecurityGroupId = formatDependentSecurityGroupIngressId(ecsEFSSecurityGroupId) ]
+
+        [@createComponentSecurityGroup
+            mode=listMode
+            tier=tier
+            component=component 
+            extensions="efs"
+            /]
+        
+        [@createSecurityGroupIngress
```
ml019: DO we need this? Maybe ingress rules added when the efs is mounted - so by defualt the efs itself has no ingress rules

aws/templates/solution/solution_ecs.ftl
```diff
@@ -53,6 +53,47 @@
...
+            name=ecsEFSVolumeName
+            component=component
+        /]
+
+        [@createEFSMountTarget
+            mode=listMode
+            tier=tier
+            efsId=ecsEFSVolumeId
+            securityGroups=ecsEFSSecurityGroupId
+            dependencies=[ecsEFSVolumeId,ecsEFSSecurityGroupId]
```
ml019: do we explicitly need dependencies given both are parameters to macro anyway so CF will likely figure them out for itself...

aws/templates/solution/solution_ecs.ftl
```diff
@@ -198,14 +240,25 @@
...
-                                    "01Fluentd"
+                                    "01Fluentd",
                                     defaultLogDriver == "fluentd",
                                     {
                                         "command" : "/opt/codeontap/bootstrap/fluentd.sh",
                                         "ignoreErrors" : "false"
                                     }) +
+                                attributeIfTrue(
+                                    "02ConfigureEFSClusterWide",
+                                    ecs.ClusterWideStorage?has_content && ecs.ClusterWideStorage == true,
```
ml019: don't check has_content - it always will

aws/templates/solution/solution_efs.ftl
```diff
@@ -0,0 +1,44 @@
+[#-- EFS --]
```
ml019: Review in light of comments above

aws/templates/solution/solution_ecs.ftl
roleyfoley: Since the security group is created as part of the ECS creation the component details are for the ECS instance instead of EFS. Since this security group is for access to the EFS instance created as part of the ECS component I added it, otherwise the group name would have matched the ecs cluster not the efs instance

aws/templates/solution/solution_ecs.ftl
roleyfoley: Name updated to formatFullComponentName and "efs" removed

aws/templates/solution/solution_ecs.ftl
roleyfoley: Two security group components are added, one which creates a security group that controls the EFS eni interfaces and the Ingress update is for the ECS cluster security group. 

Since the two are created as part of the same component then they are added together.

aws/templates/solution/solution_ecs.ftl
roleyfoley: Removed

aws/templates/solution/solution_ecs.ftl
roleyfoley: Fixed

aws/templates/resource/resource_efs.ftl
```diff
@@ -0,0 +1,74 @@
...
+
+[#assign EFS_OUTPUT_MAPPINGS =
+    {
+        REFERENCE_ATTRIBUTE_TYPE : {
+            "UseRef" : true
+        },
+        ARN_ATTRIBUTE_TYPE : { 
+            "UseRef" : true
+        },
+        NAME_ATTRIBUTE_TYPE : {
```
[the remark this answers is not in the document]
roleyfoley: Removed Name attribute as it doesn't actually have one

aws/templates/resource/resource_efs.ftl
```diff
@@ -0,0 +1,74 @@
...
+    /]
+[/#macro]
+
+[#macro createEFSMountTarget mode tier efsId securityGroups dependencies="" ]
+
+    [#if getSubnets(tier)?has_content ]
+
+        [#list getSubnets(tier) as subnet ]
+
+            [#assign efsMountTargetId = formatDependentEFSMountTargetId(
```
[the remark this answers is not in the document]
roleyfoley: Fixed

aws/templates/solution/solution_ecs.ftl
```diff
@@ -53,6 +53,47 @@
             name=formatComponentLogGroupName(tier, component) /]
     [/#if]
 
+    [#if deploymentSubsetRequired("efs", true) && 
+            ecs.ClusterWideStorage?has_content && ecs.ClusterWideStorage ]
```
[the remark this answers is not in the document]
roleyfoley: Fixed

aws/templates/resource/resource_efs.ftl
```diff
@@ -0,0 +1,71 @@
...
+                "PerformanceMode" : "generalPurpose",
+                "FileSystemTags" : getCfTemplateCoreTags(name, tier, component)
+            } 
+        outputs=EFS_OUTPUT_MAPPINGS
+    /]
+[/#macro]
+
+[#macro createEFSMountTarget mode tier efsId securityGroups dependencies="" ]
+
+    [#if getSubnets(tier)?has_content ]
```
ml019: if not necessary as getSubnets always returns a (possibly empty) array

aws/templates/resource/resource_efs.ftl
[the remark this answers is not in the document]
ml019: Doesn't look like zone is initialised at this point? 

Either we copy logic from getSubnets here (as it iterates through the zones) or we enhance getSubnets to support optionally passing the required zone(s) in as a third parameter?

aws/templates/solution/solution_ecs.ftl
ml019: ok - my bad

aws/templates/solution/solution_ecs.ftl
ml019: ok - makes sense

roleyfoley:
Added an includeZone option to GetSubnets. 
This returns an object which lists the subnet Id ( either getReference or Id ) depending on asReferences 

Updated the EFS mount targets to use the new GetSubnets options

aws/templates/solution/solution_ecs.ftl
```diff
@@ -198,14 +238,25 @@
...
-                                    "01Fluentd"
+                                    "01Fluentd",
                                     defaultLogDriver == "fluentd",
                                     {
                                         "command" : "/opt/codeontap/bootstrap/fluentd.sh",
                                         "ignoreErrors" : "false"
                                     }) +
+                                attributeIfTrue(
+                                    "02ConfigureEFSClusterWide",
+                                    ecs.ClusterWideStorage == true,
```
ml019: If it is a boolean so no need to compare to true

PR #88, opened 2018-02-12
update AMIs, userdata and add nightly crons

kshychko:
the updates for v4


ml019:
Github is showing a lot of differences in the masterData.json file but I think these are probably simply formatting differences.

Would you mind confirming that the only material differences are the changes of the AMIs (and maybe addition of extra regions)?

kshychko:
@ml019 
I checked the diff with the same formatting - looks good, only AMIs have changed

PR #94, opened 2018-02-13
Add RDS Snapshot on as part of deployment process

roleyfoley:
Quite a few RDS configuration changes trigger a replacement of the RDS instance. When this happens the data in the RDS instance is lost and needs to be restored from a snapshot. 

This adds the ability to snapshot the RDS instance each time a deployment is run

### DIFF
--- a/aws/templates/solution/solution_rds.ftl
+++ b/aws/templates/solution/solution_rds.ftl
@@ -51,11 +51,24 @@
     [#assign processorProfile = getProcessor(tier, component, "RDS")]
 
     [#if deploymentSubsetRequired("prologue", false)]
-        [#if db.SnapShotOnDeploy ]
+        [#if db.SnapShotOnDeploy!false ]
             [@cfScript
                 mode=listMode
                 content=
                 [
+                    "function create_deploy_snapshot() {",
+                    "# Create RDS snapshot",
+                    "snapshot_arn=$(create_snapshot" + " \"" + 
+                    region + "\" \"" + 
+                    rdsFullName + "\" " + 
+                    "\"pre-deploy\" ) || return $?",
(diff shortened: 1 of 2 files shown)


aws/templates/solution/solution_rds.ftl
```diff
@@ -1,164 +1,192 @@
...
-            [#assign engineVersion =
-                db.EngineVersion?has_content?then(
-                    db.EngineVersion,
-                    "9.4"
-                )
-            ]
-            [#assign family = "postgres" + engineVersion]
+    [#list getOccurrences(component, deploymentUnit) as occurrence]
+        [#assign engine = occurrence.Engine]
+        [#switch engine]
```
ml019: Engine may not be provided so suggest using occurrence.Engine!""

aws/templates/solution/solution_rds.ftl
```diff
@@ -1,164 +1,192 @@
...
+            [#case "postgres"]
+                [#assign engineVersion =
+                    occurrence.EngineVersion?has_content?then(
+                        occurrence.EngineVersion,
+                        "9.4"
+                    )
+                ]
+                [#assign family = "postgres" + engineVersion]
+                [#break]
+        [/#switch]
```
ml019: Add a default which adds an exception flagging the egine type is unknown

aws/templates/solution/solution_rds.ftl
ml019: Suggest adding default for port if no content provided
Also suggest default size per engine type

roleyfoley:
Add default port settings.
Defaulted the size in the common.ftl. Would you need engine specific sizing?

roleyfoley:
Issue #131

PR #97, opened 2018-02-15
Add Support for Alternative Template Processing based on changes

roleyfoley:
During cloudformation processing some property updates require that the resource they effect are replaced.

For resources that maintain state, such as RDS this means that a new, empty database is created. 

With Alternative templates a collection of passes are made during template generation and multiple template files are created. 

When they processed, one of the templates can be chosen and used to generate the infrastructure. 

In the initial stage a replace template is created, when ManageTemplate is called the standard(primary) template is submitted to cloudformation as a change set. If any resource in the ChangeSet is flagged for replacement Cloudformation will use the replace alternative template to perform the changes. 

In the case of RDS the template would contain the DBSnapShotIdentifier property which tells cloudformation to create the new database in the replacement process using the specified snapshot.


roleyfoley:
Started this again to integrate into overall updates.

roleyfoley:
Managed to move changes into this branch. All updates now use the change set process. 

As part of this process new files will be added to the infrastructure folder in the CMDB along with the template and output json files 

change - The output results of the change set listing what has changed between updates 
dryrun - The output results of a change set if manageStack is run in dry run mode.

aws/manageStack.sh
```diff
@@ -241,35 +306,102 @@ function process_stack() {
...
+                cat "${CHANGE}"; return ${exit_status}; 
+
+            else 
+
+              replacement=$( cat "${CHANGE}" | jq '[.Changes[].ResourceChange.Replacement] | contains(["True"])' )
+              REPLACE_TEMPLATES=$( for i in ${ALTERNATIVE_TEMPLATES} ; do echo $i | awk '/-replace[0-9]-template\.json$/'  ; done  | sort  )
+              
+              info "Replacement Status: ${replacement}"
+
+              if [[ "${replacement}" == "true" && -n REPLACE_TEMPLATES ]]; then
```
ml019: -n REPLACE_TEMPLATES will always be true - need to do -n "${REPLACE_TEMPLATES}"

aws/setStackContext.sh
```diff
@@ -142,21 +142,42 @@ esac
 STACK_NAME="${STACK_NAME:-${PRODUCT_PREFIX}${SEGMENT_SUFFIX}${LEVEL_SUFFIX}${DEPLOYMENT_UNIT_SUFFIX}${DEPLOYMENT_UNIT_SUBSET_SUFFIX}}"
 TEMPLATE="${LEVEL_PREFIX}${DEPLOYMENT_UNIT_PREFIX}${DEPLOYMENT_UNIT_SUBSET_PREFIX}${REGION_PREFIX}template.json"
 STACK="${LEVEL_PREFIX}${DEPLOYMENT_UNIT_PREFIX}${DEPLOYMENT_UNIT_SUBSET_PREFIX}${REGION_PREFIX}stack.json"
+CHANGE="${LEVEL_PREFIX}${DEPLOYMENT_UNIT_PREFIX}${DEPLOYMENT_UNIT_SUBSET_PREFIX}${REGION_PREFIX}change.json"
```
ml019: Do we need dryrun? Isn't change the same? So we always create the change set, but don't action it if DRYRUN is true...

aws/createTemplate.sh
```diff
@@ -292,62 +300,69 @@ function process_template() {
...
-            cat "${template_result_file}" | jq --indent 1 "${jq_pattern}" | sed "${sed_patterns[@]}" > "${template_result_file}-new"
-            cat "${output_file}" | jq --indent 1 "${jq_pattern}" | sed "${sed_patterns[@]}" > "${template_result_file}-existing"
-
-            diff "${template_result_file}-existing" "${template_result_file}-new" > "${template_result_file}-difference" &&
-              info "Ignoring unchanged ${file_description} file ...\n" ||
+    for pass_alternative in ${pass_alternatives["${pass}"]}; do
+
+      [[ "${pass_alternative}" == "primary" ]] && pass_alternative=""
+      pass_args+=("-v" "alternative=${pass_alternative}")
+      pass_alternative_suffix="${pass_alternative:+${pass_alternative}-}"
```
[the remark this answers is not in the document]
roleyfoley: Done

aws/manageStack.sh
```diff
@@ -204,18 +204,82 @@ function add_host_to_apidoc() {
   return 0
 }
 
-function process_stack() {
+function wait_for_stack_execution() { 
+  
+  info "watching Stack Execution"
```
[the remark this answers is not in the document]
roleyfoley: Done

aws/setStackContext.sh
roleyfoley: The idea I had was that CHANGE contains the changes that were made as part of the last run. DRYRUN shows the changes that would occur on the next run. Given that the DRYRUN changes are show echp'd maybe we don't need this?

roleyfoley: Removed the DRYRUN file output. Instead the Changes will just be displayed

roleyfoley:
Issue #131

PR #99, opened 2018-02-19
Add s3 link support

ml019:
I refactored the bucket support to use a couple of format routines for different bucket types and added s3 as a valid destination for a link.

I didn't add Cloudfront/Certificate support yet - figured I'd wait until we had a project that needed it.

### DIFF
--- a/aws/templates/common.ftl
+++ b/aws/templates/common.ftl
@@ -91,6 +91,14 @@
         value?has_content?then(value,content)) ]
 [/#function]
 
+[#function firstContent alternatives=[] otherwise={}]
+    [#list asArray(alternatives) as alternative]
+        [#if alternative?has_content]
+            [#return alternative]
+        [/#if]
+    [#return otherwise ]
+[/#function]
+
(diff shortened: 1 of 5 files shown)


aws/templates/common.ftl
```diff
@@ -1245,6 +1254,27 @@
...
+            [#local result +=
+                {
+                    "ResourceId" : id,
+                    "Attributes" : {
+                        "NAME" : getExistingReference(id, NAME_ATTRIBUTE_TYPE),
+                        "FQDN" : getExistingReference(id, DNS_ATTRIBUTE_TYPE),
+                        "INTERNAL_FQDN" : getExistingReference(id, DNS_ATTRIBUTE_TYPE),
+                        "WEBSITE_URL" : getExistingReference(id, URL_ATTRIBUTE_TYPE),
+                        "ARN" : getExistingReference(id, ARN_ATTRIBUTE_TYPE),
+                        "REGION" : regionId
```
roleyfoley: Is the regionId based on the set context of the current run? for cross region components like S3, should this be part of the getExistingReference?

PR #122, opened 2018-03-02
Add RDS Snapshot Restoration

roleyfoley:
Allows for the restoration of a pre deploy snapshot created during the deployment process. Using replacement templates the replace1 template performs the restoration, renaming the database Id to a temp name. Replace2 then renames and restores the data base back to the original name. 

This should allow for database content to be maintained when a change is requested that requires replacement. 

Adds the following: 

- Restoration from pre deploy snapshot if a replacement has been triggered on the cloudformation template 
- A runId parameter passed in at createTemplate.sh providing  a unique random string which can referenced within template runs ( used to create a unique Id for the pre deployment snapshot) 
- RDS resource template/macro 
- RDS at rest encryption using segment KMS key


roleyfoley:
Issue #131

roleyfoley:
Tested today
- Created database 
  - created database as normal run 
- Update with no changes
  - No changes detected or performed
- Added Snapshotting 
  - Single run using normal update process. Snapshot created and pseudo stack created 
- Updated with a replacement change 
  - Triggered replacement processing, database renamed and created from snapshot, 2nd run renamed database to original name 
- Updated with a non replacement change 
  - Normal update processing

PR #123, opened 2018-03-02
Contenthub Path Naming Fixes and Link Lookup

roleyfoley:
Reorder pathing to follow least specific to more specific flow.

Fix link lookup to use external lookup as the type instead of contenthub 

Added Path styling, allowing for the use of a nested folder structure or single folder containing each of the included properties

### DIFF
--- a/aws/templates/common.ftl
+++ b/aws/templates/common.ftl
@@ -1028,22 +1028,31 @@
     [#local pathObject = occurrence.Configuration.Path ]
     [#local includes = pathObject.IncludeInPath]
 
-    [#return
-        valueIfTrue(
-            pathObject.Host,
+    [#local path =  valueIfTrue(
+            [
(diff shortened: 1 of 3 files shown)


aws/templates/common.ftl
```diff
@@ -1028,22 +1028,31 @@
...
-                valueIfTrue(segmentName!"", includes.Segment),
-                valueIfTrue(environmentName!"", includes.Environment),
-                valueIfTrue(productName!"", includes.Product)
-            )
+                valueIfTrue(pathObject.Host, includes.Host)
+            ]
         )
     ]
+    
+    [#if pathObject.Style = "Single" ]
```
ml019: I tend to use all lower case for variable values i.e. "single". Change in componentConfiguration as well.

PR #127, opened 2018-03-02
Add SQS link role support

ml019:
Manage SQS permissions via link roles
Start refactoring use of occurrence/tier/component in generating names/ids.

### DIFF
--- a/aws/templates/id/id_sqs.ftl
+++ b/aws/templates/id/id_sqs.ftl
@@ -2,27 +2,6 @@
 
 [#assign SQS_RESOURCE_TYPE = "sqs" ]
 
-[#function formatSQSId ids...]
-    [#return formatResourceId(
-                SQS_RESOURCE_TYPE,
-                ids)]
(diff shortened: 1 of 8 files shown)


aws/templates/id/id_sqs.ftl
```diff
@@ -50,13 +29,35 @@
...
+                            formatName(
+                                core.Component.Name,
+                                occurrence),
+                            formatName(
+                                productName,
+                                segmentName,
+                                componentName,
+                                occurrence))]
+
+    [#local dlqId = formatDependentResourceId(SQS_RESOURCE_TYPE, id, "dlq") ]
```
roleyfoley: Added the resource type here would make it sqsXsqs wouldn't it?

ml019: yes.

The first sqs is the resource type, the next is the type of the resource it depends on - in this case an SQS queue as well.

PR #128, opened 2018-03-02
Support for credentials in Content Hub Git pulls

roleyfoley:
In order to push changes to Github credentials need to be provided when the repo is pulled. 

This adds the ability to lookup the github credentials using the env variable set by the Jenkins server.

### DIFF
--- a/aws/templates/id/id_contenthub.ftl
+++ b/aws/templates/id/id_contenthub.ftl
@@ -34,7 +34,7 @@
                 "Default" : "master"
             },
             {
-                "Name" : "Repoistory",
+                "Name" : "Repository",
                 "Default" : ""
             }
         ]
--- a/aws/templates/id/id_contenthub.ftl
+++ b/aws/templates/id/id_contenthub.ftl
@@ -27,14 +27,14 @@
             "Prefix",
             {
                 "Name" : "Engine",
-                "Default" : "git"
+                "Default" : "github"
             },
             {
                 "Name" : "Branch",
                 "Default" : "master"
             },
             {
-                "Name" : "URL",
+                "Name" : "Repoistory",
                 "Default" : ""
             }
         ]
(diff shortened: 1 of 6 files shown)


aws/templates/id/id_contenthub.ftl
```diff
@@ -27,14 +27,14 @@
...
-                "Default" : "git"
+                "Default" : "github"
             },
             {
                 "Name" : "Branch",
                 "Default" : "master"
             },
             {
-                "Name" : "URL",
+                "Name" : "Repoistory",
```
ml019: Looks like a typo

aws/templates/id/id_contenthub.ftl
```diff
@@ -56,7 +56,7 @@
             },
             "Attributes" : {
                 "ENGINE" : getExistingReference(id, ENGINE_ATTRIBUTE_TYPE),
-                "URL"    : getExistingReference(id, URL_ATTRIBUTE_TYPE),
+                "REPOSITORY" : getExistingReference(id, REPOSITORY_ATTRIBUTE_TYPE),
                 "BRANCH" : getExistingReference(id, BRANCH_ATTRIBUTE_TYPE),
                 "PREFIX" : getExistingReference(id, PREFIX_ATTRIBUTE_TYPE)
```
ml019: How does prefix differ from path?

aws/utility.sh
```diff
@@ -715,16 +715,23 @@ function create_snapshot() {
...
   local local_dir="$1"; 
 
-  [[ (-z "${repo_url}") ||
+  [[  (-z "${repo_provider}") ||
+      (-z "${repo_host}") ||
+      (-z "${repo_path}") ||
       (-z "${repo_branch}") ||
       (-z "${local_dir}") ]] && fatalMandatory && return 1
 
   trace "Cloning the ${repo_url} repo and checking out the ${repo_branch} branch ..."
```
ml019: is repo_url still defined?

aws/templates/id/id_contenthub.ftl
roleyfoley: Prefix was for the Hub side and could be a folder or appended to a name. Path is for the node side to identify who the node is. Since they are used together in the scripts I wanted to seperate them.

aws/utility.sh
roleyfoley: Its built in the function so that I can include the credentials, the trace was in the wrong spot.

aws/templates/id/id_contenthub.ftl
roleyfoley: Fixed

Issue #139, opened 2018-03-04
Add support for controlling RDS minor release upgrade policy


ml019:
Add a configuration setting to control whether minor version upgrades are permitted in the maintenance window.

Support a per environment default, with it turned off in production environments and enabled otherwise.

ml019:
hmm - ok just reread my issue and can see why you made the changes you did. I'll need to update my review

roleyfoley:
@ml019 and @kshychko 
Can this be closed off now?

ml019:
Yep.

PR #154, opened 2018-03-07
add support for controlling AutoMinorVersionUpgrade property

kshychko:
AutoMinorVersionUpgrade property is supported on the tier, segment, solution and environment levels.
per  #139

### DIFF
--- a/aws/data/masterData.json
+++ b/aws/data/masterData.json
@@ -680,6 +680,9 @@
       "Description": "Kind of obvious...",
       "Category": "prod",
       "MultiAZ": true,
+      "RDS": {
+        "AutoMinorVersionUpgrade": false
+      },
       "Operations": {
         "Expiration": 365,
         "FlowLogs": {
--- a/aws/templates/id/id_rds.ftl
+++ b/aws/templates/id/id_rds.ftl
@@ -70,7 +70,8 @@
                         "Default" : false
                     }
                 ]
-            }
+            },
+            "AutoMinorVersionUpgrade"
         ]
 }]
     
--- a/aws/templates/resource/resource_rds.ftl
+++ b/aws/templates/resource/resource_rds.ftl
@@ -40,6 +40,7 @@
     component=""     
     dependencies="" 
     outputId=""
+    autoMinorVersionUpgrade=true
 ]
     [@cfResource 
     mode=listMode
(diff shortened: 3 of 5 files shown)


aws/data/masterData.json
```diff
@@ -680,6 +680,9 @@
       "Description": "Kind of obvious...",
       "Category": "prod",
       "MultiAZ": true,
+      "RDS": {
```
ml019: Apply the default in id_rds not in the master data

aws/templates/id/id_rds.ftl
```diff
@@ -70,7 +70,8 @@
                         "Default" : false
                     }
                 ]
-            }
+            },
+            "AutoMinorVersionUpgrade"
```
ml019: Apply default here

aws/templates/resource/resource_rds.ftl
```diff
@@ -40,6 +40,7 @@
     component=""     
     dependencies="" 
     outputId=""
+    autoMinorVersionUpgrade=true
```
ml019: think default here should match the default in the component configuration

aws/templates/setContext.ftl
```diff
@@ -166,6 +166,7 @@
 [#if blueprintObject.Solution?has_content]
     [#assign solutionObject = blueprintObject.Solution]
     [#assign solnMultiAZ = solutionObject.MultiAZ!(environmentObject.MultiAZ)!false]
+    [#assign RDSAutoMinorVersionUpgrade = (segmentObject.RDS.AutoMinorVersionUpgrade)!(solutionObject.RDS.AutoMinorVersionUpgrade)!(environmentObject.RDS.AutoMinorVersionUpgrade)!true]
```
ml019: I don't think this belongs here - we are moving stuff out of here - don't want to be adding things right now. Better to be in solution_rds via component configuration.

aws/templates/solution/solution_rds.ftl
```diff
@@ -217,6 +217,7 @@
                     parameterGroupId=getReference(rdsParameterGroupId)
                     optionGroupId=getReference(rdsOptionGroupId)
                     securityGroupId=getReference(rdsSecurityGroupId)
+                    autoMinorVersionUpgrade = configuration.AutoMinorVersionUpgrade!RDSAutoMinorVersionUpgrade
```
ml019: just rely on configuration value

ml019:
@kshychko ignore my review - I forgot about needing a defualt difference between production and other environments. Changes look good. I'll merge

Issue #162, opened 2018-03-09
Add Action based lambda triggers to Cognito User pools


roleyfoley:
Lambda has the ability to trigger lambda functions based on a given event that occurs on a Cognito user pool 

https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools-working-with-aws-lambda-triggers.html 

This should be reasonably straight forward to add to the framework using links. When a cognito solution is processed, using the link names we whould be able to match the functions to the appropriate cognito action.

roleyfoley:
Available states are listed in the cloudformation configuration 
{
  "CreateAuthChallenge" : String,
  "CustomMessage" : String,
  "DefineAuthChallenge" : String,
  "PostAuthentication" : String,
  "PostConfirmation" : String,
  "PreAuthentication" : String,
  "PreSignUp" : String,
  "VerifyAuthChallengeResponse" : String
}

Each string value being the ARN of a lambda function that will be invoked on the action.

roleyfoley:
The thinking is that when a userpool is processed, if it has lambda links that match the names listed they are added to the lambda user pool configuration.

PR #174, opened 2018-03-11
Add Cogntio event lambda triggers

roleyfoley:
Issue #162 
- Converted userpool solution to an occurrence based component
- Fixed some configuration properties casing so that they worked with the occurrence default settings
- Added Trigger links to user pool. When a lambda is linked from a user pool using one of the named actions for userpool event, the specified function will be set in the LambdaConfig section of the user pool template 
- Added inbound link properties to allow the lambda function to grant permission for the userpool to invoke it using the standard linking methods 
- Added ARN and Name attributes to lambda functions so that they can be looked up easily as part of a link 
- Added option to stop an error if a link is not found, because lambda functions are created during the application level deployment phase they aren't available for the initial creation of the userpool. So the user pool actually needs to be deployed initially as a solution, then lambda, then the userpool needs to be updated to get it working completely. This means that the link will fail on the first round, adding allowNotFound on the getLinkTarget stops the preconditionException being raised.

### DIFF
--- a/aws/templates/application/application_lambda.ftl
+++ b/aws/templates/application/application_lambda.ftl
@@ -51,17 +51,16 @@
 
                     [#switch linkTargetCore.Type!""]
                         [#case "userpool"] 
-                            [@createPolicy 
-                                mode=listMode
-                                id=formatDependentPolicyId(fnId, "link", linkName)
-                                name=fnName
-                                statements=[getPolicyStatement(
-                                                "lambda:InvokeFunction",
-                                                formatLambdaArn(fnId))]
-                                roles=formatDependentIdentityPoolAuthRoleId(
-                                        formatIdentityPoolId(linkTargetCore.Tier, linkTargetCore.Component))
-                            /]
-                            [#break]
+                            [#if linkTargetResources["userpool"].Deployed &&
+                                    (linkDirection == "inbound")]
+                                [@createLambdaPermission
+                                    mode=listMode
+                                    id=formatLambdaPermissionId(fn, "link", linkName)
+                                    targetId=fnId
+                                    source=linkTargetRoles.Inbound["invoke"]
+                                /]
+                            [/#if]
+                        [#break]
 
                         [#case "apigateway" ]
                             [#if linkTargetResources["gateway"].Deployed &&
@@ -73,7 +72,7 @@
                                     source=linkTargetRoles.Inbound["invoke"]
                                 /]
                             [/#if]
-                            [#break]
+                        [#break]
                     [/#switch]    
                 [/#list]
             [/#if]
--- a/aws/templates/application/application_lambda.ftl
+++ b/aws/templates/application/application_lambda.ftl
@@ -51,7 +51,9 @@
 
                     [#switch linkTargetCore.Type!""]
                         [#case USERPOOL_COMPONENT_TYPE] 
-                            [#if linkTargetResources[USERPOOL_COMPONENT_TYPE].Deployed &&
+                        [#case "apigateway"]
+                        [@cfDebug listMode  linkTargetResources true /]
+                            [#if linkTargetResources[(linkTargetCore.Type)].Deployed &&
                                     (linkDirection == "inbound")]
                                 [@createLambdaPermission
                                     mode=listMode
@@ -60,19 +62,7 @@
                                     source=linkTargetRoles.Inbound["invoke"]
                                 /]
                             [/#if]
-                        [#break]
-
-                        [#case "apigateway" ]
-                            [#if linkTargetResources["gateway"].Deployed &&
-                                    (linkDirection == "inbound")]
-                                [@createLambdaPermission
-                                    mode=listMode
-                                    id=formatLambdaPermissionId(fn, "link", linkName)
-                                    targetId=fnId
-                                    source=linkTargetRoles.Inbound["invoke"]
-                                /]
-                            [/#if]
-                        [#break]
+                            [#break]
                     [/#switch]    
                 [/#list]
             [/#if]
--- a/aws/templates/application/application_lambda.ftl
+++ b/aws/templates/application/application_lambda.ftl
@@ -52,7 +52,6 @@
                     [#switch linkTargetCore.Type!""]
                         [#case USERPOOL_COMPONENT_TYPE] 
                         [#case "apigateway"]
-                        [@cfDebug listMode  linkTargetResources true /]
                             [#if linkTargetResources[(linkTargetCore.Type)].Deployed &&
                                     (linkDirection == "inbound")]
                                 [@createLambdaPermission
--- a/aws/templates/application/application_lambda.ftl
+++ b/aws/templates/application/application_lambda.ftl
@@ -50,8 +50,8 @@
                     [#assign linkDirection = linkTarget.Direction ]
 
                     [#switch linkTargetCore.Type!""]
-                        [#case "userpool"] 
-                            [#if linkTargetResources["userpool"].Deployed &&
+                        [#case USERPOOL_COMPONENT_TYPE] 
+                            [#if linkTargetResources[USERPOOL_COMPONENT_TYPE].Deployed &&
                                     (linkDirection == "inbound")]
                                 [@createLambdaPermission
                                     mode=listMode
--- a/aws/templates/common.ftl
+++ b/aws/templates/common.ftl
@@ -979,7 +979,7 @@
     [#return getOccurrencesInternal(component, tier) ]
 [/#function]
 
-[#function getLinkTarget occurrence link ]
+[#function getLinkTarget occurrence link]
 
     [#if link.Tier?lower_case == "external"]
         [#local targetOccurrence =
--- a/aws/templates/common.ftl
+++ b/aws/templates/common.ftl
@@ -979,7 +979,7 @@
     [#return getOccurrencesInternal(component, tier) ]
 [/#function]
 
-[#function getLinkTarget occurrence link allowNotFound=false ]
+[#function getLinkTarget occurrence link ]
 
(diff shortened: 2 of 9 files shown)


aws/templates/application/application_lambda.ftl
```diff
@@ -73,7 +72,7 @@
                                     source=linkTargetRoles.Inbound["invoke"]
                                 /]
                             [/#if]
-                            [#break]
+                        [#break]
```
ml019: This looks like the wrong level of indenting

aws/templates/application/application_lambda.ftl
```diff
@@ -51,17 +51,16 @@
...
-                                id=formatDependentPolicyId(fnId, "link", linkName)
-                                name=fnName
-                                statements=[getPolicyStatement(
-                                                "lambda:InvokeFunction",
-                                                formatLambdaArn(fnId))]
-                                roles=formatDependentIdentityPoolAuthRoleId(
-                                        formatIdentityPoolId(linkTargetCore.Tier, linkTargetCore.Component))
-                            /]
-                            [#break]
+                            [#if linkTargetResources["userpool"].Deployed &&
```
ml019: I think we can consolidate this code with the api gateway code. Freemarker allows case statements to flow through to one another so we can just add [#case "userpool"] above the apigateway code. Only other change would be to replace the hardcoded type with linkTargetCode.Type. then if we get another target that needs to invoke lambda, we just add its type. 

Also suggest we use the *_COMPONENT_TYPE variables here rather than fixed strings

aws/templates/id/id_cognito.ftl
```diff
@@ -90,33 +92,37 @@
                 "Default" : 30
             },
             {
-                "Name" : "allowUnauthIds",
+                "Name" : "allowUnauthenticatedIds",
```
ml019: Config items should  always start with an upper case letter.

aws/templates/id/id_cognito.ftl
```diff
@@ -90,33 +92,37 @@
...
                 "Default" : false
             }
             {
                 "Name" : "passwordPolicy",
                 "Children" : [
                     {
-                       "Name" : "MinimumLength",
-                       "Default" : "8"
+                       "Name" : "minimumLength",
```
ml019: Not sure why you reversed the case on these. Everywhere else we use Title case... ?

aws/templates/id/id_cognito.ftl
```diff
@@ -90,33 +92,37 @@
...
-                        "Name" : "SpecialCharacters",
-                        "Default" : false
+                        "Name" : "specialCharacters",
+                        "Default" : true
                     }
                 ] 
+            },
+            {
+                "Name" : "Links",
+                "Default" : {}
```
ml019: if we merge #175, this shuld be changed to use the explicit config for children with repeating content

aws/templates/id/id_lambda.ftl
```diff
@@ -90,16 +90,20 @@
...
                 "function" : {
-                    "Id" : formatResourceId(LAMBDA_FUNCTION_RESOURCE_TYPE, core.Id),
+                    "Id" : id,
                     "Name" : formatSegmentFullName(core.Name)
                 }
             },
             "Attributes" : {
-                "REGION" : regionId
+                "REGION" : regionId,
+                "ARN" : getExistingReference("id",ARN_ATTRIBUTE_TYPE)!"",
```
ml019: No need for default - getExistingReference returns "" if the id isn't found. Also id shouldn't be in quotes

aws/templates/id/id_lambda.ftl
```diff
@@ -90,16 +90,20 @@
...
+                "NAME" : getExistingReference("id",NAME_ATTRIBUTE_TYPE)!""
```
ml019: see previous comment

aws/templates/solution/solution_userpool.ftl
```diff
@@ -1,209 +1,301 @@
...
+    [#list requiredOccurrences(
+        getOccurrences(tier, component),
+        deploymentUnit) as occurrence]
 
+        [@cfDebug listMode occurrence false /]
+
+        [#assign core = occurrence.Core ]
+        [#assign configuration = occurrence.Configuration ]
+
+        [#assign userPoolId = formatUserPoolId(tier, component, occurrence)]
```
ml019: Add the Id and Name functions could be simplified to just accept  (occurrence extension...) couldn't they? See lambda and s3 for examples.

Means now having pass around tier and component everywhere

aws/templates/solution/solution_userpool.ftl
```diff
@@ -1,209 +1,301 @@
...
+                [#assign emailSchema = getUserPoolSchemaObject( 
+                                            "email",
+                                            "String",
+                                            true,
+                                            true)]
+                [#assign schema = schema +  [ emailSchema ]]
+        [/#if]
+
+        [#list configuration.Links?values as link]
+            [#if link?is_hash]
```
ml019: Links will only be hashes following #175 - can remove this [#if] block

aws/templates/common.ftl
```diff
@@ -889,7 +889,7 @@
     [#return getOccurrencesInternal(component, tier) ]
 [/#function]
 
-[#function getLinkTarget occurrence link]
+[#function getLinkTarget occurrence link allowNotFound=false ]
```
ml019: Not sure about this. not found is about the link target not being in the solution file - which it should be.

I think what you want is to check if the target is deployed.

aws/templates/solution/solution_userpool.ftl
```diff
@@ -1,209 +1,301 @@
...
+                [#switch linkTargetCore.Type!""]
+                    [#case LAMBDA_FUNCTION_COMPONENT_TYPE ]
+
+                        [#-- Cognito Userpool Event Triggers --]
+                        [#switch link.Name ]
+                            [#case "createauthchallenge" ]
+                                [#assign userPoolTriggerConfig =  userPoolTriggerConfig +
+                                    attributeIfContent (
+                                        "CreateAuthChallenge",
+                                        linkTargetAttributes.ARN
```
ml019: think we should be checking the Deployed property and remove the allowNotFound flag...

aws/templates/application/application_lambda.ftl
roleyfoley: Fixed

roleyfoley:
All suggestions should now be fixed up. Removed the allowNotFound. Think this was a bug in my solution.

aws/templates/application/application_lambda.ftl
```diff
@@ -50,18 +50,17 @@
...
+                            [#if linkTargetResources[USERPOOL_COMPONENT_TYPE].Deployed &&
+                                    (linkDirection == "inbound")]
+                                [@createLambdaPermission
+                                    mode=listMode
+                                    id=formatLambdaPermissionId(fn, "link", linkName)
+                                    targetId=fnId
+                                    source=linkTargetRoles.Inbound["invoke"]
+                                /]
+                            [/#if]
+                        [#break]
```
ml019: breaks should line up with other content of case - so in line with [#if]

aws/templates/application/application_lambda.ftl
```diff
@@ -50,18 +50,17 @@
...
-                                name=fnName
-                                statements=[getPolicyStatement(
-                                                "lambda:InvokeFunction",
-                                                formatLambdaArn(fnId))]
-                                roles=formatDependentIdentityPoolAuthRoleId(
-                                        formatIdentityPoolId(linkTargetCore.Tier, linkTargetCore.Component))
-                            /]
-                            [#break]
+                        [#case USERPOOL_COMPONENT_TYPE] 
+                            [#if linkTargetResources[USERPOOL_COMPONENT_TYPE].Deployed &&
```
ml019: if you change this to  linkTargetResources[linkTargetCore,Type].Deployed,

then the code for apigateway and userpool is identical and I thing the two case blocks can collapse to one;

[#case USERPOOL_COMPONENT_TYPE]
[#case "apigateway"]
....

aws/templates/id/id_cognito.ftl
```diff
@@ -1,30 +1,35 @@
...
 [#assign USERPOOL_RESOURCE_TYPE = "userpool"]
-[#assign USERPOOL_CLIENT_RESOURCE_TYPE = "userpoolclient" ]
-[#assign IDENTITYPOOL_RESOURCE_TYPE = "identitypool" ]
+[#assign USERPOOL_CLIENT_RESOURCE_TYPE = "userpoolclient"]
+[#assign IDENTITYPOOL_RESOURCE_TYPE = "identitypool"]
 
-[#function formatUserPoolId tier component extensions...]
+[#assign USERPOOL_COMPONENT_TYPE = "userpool"]
+
+[#function formatUserPoolId occurrence extensions...]
```
ml019: these could simplify further to

[#return formatResourceId(USERPOOL_RESOURCE_TYPE,occurrence.Core.Id, extensions)

since Core.Id is effectively tier, component and id extensions

aws/templates/name/name_cognito.ftl
```diff
@@ -1,24 +1,28 @@
 [#-- Cognito User Pool --]
 
-[#function formatUserPoolName tier component ]
+[#function formatUserPoolName occurrence extensions...]
```
ml019: Like ids, these could simply to 

these could simplify further to

these could simplify further to

[#return formatSegmentFullName(occurrence.Core.Name, extensions) ]

the formatName is redundant even in the existing form

roleyfoley:
Indentation Fixed up
To consolidate the lambda linking I had to change the APIGateway resource type from "gateway" to "apigateway"

PR #175, opened 2018-03-11
Component configuration enhancements

### LINKED ISSUE #153: Component configuration to support mandatory and frequency indications
* add Mandatory indicator on all attributes - default false
* add frequency indicator - so attribute rules then apply to each unique child of the parent, rather than being expected as direct child attributes of the parent

This will allow explicit configuration in a range of situations where currently we leave the parent model open (e.g. Containers and Links). It will also support aggregation of children e.g. where a common set of links is required, along with some specific ones.

ml019:
Fixes #153

### DIFF
--- a/aws/templates/commonApplication.ftl
+++ b/aws/templates/commonApplication.ftl
@@ -253,193 +253,188 @@
 
     [#local containers = [] ]
 
-    [#list (configuration.Containers!{})?values as container]
-        [#if container?is_hash]
-
-            [#local portMappings = [] ]
-            [#list (container.Ports!{})?values as port]
-                [#if port?is_hash]
-                    [#local targetLoadBalancer = {} ]
-                    [#local targetTierId = (port.LB.Tier)!port.ELB?has_content?then("elb", "") ]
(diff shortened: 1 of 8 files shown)


aws/templates/commonApplication.ftl
```diff
@@ -253,193 +253,188 @@
...
-                                [#if core.Instance.Id == targetOccurrence.Core.Instance.Id]
-                                    [#if core.Version.Id == targetOccurrence.Core.Version.Id]
-                                        [#local instanceAndVersionMatch = targetOccurrence]
-                                    [#else]
-                                        [#if (core.Version.Id?has_content) &&
-                                                (!(targetOccurrence.Core.Version.Id?has_content))]
-                                            [#local instanceMatch = targetOccurrence]
-                                        [/#if]
-                                    [/#if]
+    [#list configuration.Containers?values as container]
```
roleyfoley: With resources being decentralised more and more should the container configuration be moved into either its own files or into the ECS configuration?

ml019: Originally this routine was shared by lambda but its now only used by ecs - so I think it could move. However the application_* templates are designed to be built into a switch statement so the current location is the most logical place at the moment.

PR #183, opened 2018-03-12
Occurrence support improvements for rds and cache

### LINKED ISSUE #182: Refactor occurrence support for rds/cache
Cache should use occurrences. 

Both should use the Resources of the occurence.

ml019:
Fixes #182 

Convert the cache template to use occurrences. Improve the default
handling of rds and cache.

Add the component full name as an extra attribute of the Core object so
it can be used when defining resource names. Convert existing usage of
formatSegmentFullName to use this new attribute instead.

### DIFF
--- a/aws/templates/common.ftl
+++ b/aws/templates/common.ftl
@@ -668,6 +668,10 @@
                 [#local result = getAPIGatewayState(occurrence)]
                 [#break]
 
+            [#case "cache"]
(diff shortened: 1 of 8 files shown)


ml019:
@roleyfoley I need this update to align with my solution for visa, but please review and let meknow if you are not happy with any of the changes.

PR #186, opened 2018-03-13
Convert APIGateway to new Id Format and add attributes for ContentNode or SPA

### LINKED ISSUE #120: Add API Gateway Docs URL as an API Gateway Component Attribute
Allows other services to discover the published API Docs URL and as a result of this the swagger spec that the API Gateway uses

### LINKED ISSUE #184: ContentNode Path Attribute returning Path Config instead of path
Currently the ContentNode Path Attribute is returning the path configuration instead of the determined path which is the intention of the attribute details.

### LINKED ISSUE #185: Add SPA Endpoint Attributes
Add the URL and FQDN details for an SPA so that they can be referenced by documentation and other components.

roleyfoley:
Add key attribute details for: 

SPA - FQDN, URL 
APIGateway - Docs URL 
ContentNode - Node Side path 

Fixes #185 
Fixes #184 
Fixes #120


ml019:
I've already added docs URL for visa reform

aws/templates/id/id_apigateway.ftl
```diff
@@ -146,6 +146,7 @@
 
     [#local id = formatAPIGatewayId(core.Tier, core.Component, occurrence)]
     [#local name = formatComponentFullName(core.Tier, core.Component, occurrence)]
+    [#local docsBucketId = formatS3Id(core.Id, "docs")]
```
ml019: Should be included in the resources as well? And application_gateway changed to use the resource id?

aws/templates/id/id_spa.ftl
```diff
@@ -71,10 +71,26 @@
     }]
     
 [#function getSPAState occurrence]
+    [#local core = occurrence.Core]
+    [#local configuration = occurrence.Configuration]
+
+    [#assign cfId  = formatComponentCFDistributionId(core.Tier, core.Component,occurrence)]
```
ml019: should be added to the resources and used in solution/application_spa?

aws/templates/id/id_apigateway.ftl
```diff
@@ -185,30 +160,71 @@
...
-                "URL" : "https://" + fqdn + versionPath,
-                "DOCS_URL" : "http://docs." + fqdn,
+                "URL" : "https://" + fqdn,
                 "SIGNING_FQDN" : signingFqdn,
                 "SIGNING_URL" : "https://" + signingFqdn + versionPath,
                 "INTERNAL_FQDN" : internalFqdn,
-                "INTERNAL_URL" : "https://" + internalFqdn + versionPath,
-                "INTERNAL_PATH" : internalPath
+                "INTERNAL_URL" : "https://" + internalFqdn,
+                "DOCS_URL" : "http://" + getExistingReference(docsId, NAME_ATTRIBUTE_TYPE)
```
ml019: NAME_ATTRIBUTE looks wrong - should be a URL attribute?

aws/templates/id/id_spa.ftl
```diff
@@ -88,10 +89,42 @@
...
     [#return
         {
-            "Resources" : {},
-            "Attributes" : {}
+            "Resources" : {
+                "cf" : {
+                    "Id" : cfId,
+                    "Name" : cfName
+                },
+                "cforiginspa" : {
```
ml019: Should we be nesting these? I'd prefer we use hierarchy than compound names

aws/templates/id/id_apigateway.ftl
roleyfoley: The S3 website url contains the aws suffix on the end - ap-southeast-2.aws.com or something like that. The name is actually the FQDN that we want to use for the Docs site. Since thats the DNS entry we put in.

aws/templates/id/id_spa.ftl
roleyfoley: Perhaps it should be 2 cforigin resources with a diferent Id?

PR #193, opened 2018-03-13
Roll back URL for RDS due to password encryption.

roleyfoley:
Since we encrypt RDS passwords using KMS, passing the URL straight through to the application poses potential issues as the KMS decryption process is usually done as part of a dedicated app startup process.

Instead this PR adds the database engine as an attribute so that all of the components to build a database URL are available.


roleyfoley:
No longer required. URL will be encrypted to allow for applications to get the benefit of the URL properties

PR #238, opened 2018-03-26
Add userpool lookup based on Account

roleyfoley:
To support userpool authorisation on api Gateway the ARN needs to be configured as part of the authoriser configuration. 

This needs to be updated for each account so instead of a sinlge value for userPoolARN it is now expected to have userPoolARNs hash which contains a key value pair with the account name as the key and the value as the ARN


roleyfoley:
Branch created from wrong source branch

PR #243, opened 2018-03-28
Convert Application and Solution Components to Occurrences and add Type information

roleyfoley:
Within a given component a resource can be used in different ways so the logical name of a resource might not reflect the type of resource that is being provisioned. 

This means that you can't identify what is being deployed based on the resources in a component. 

This PR adds a type field to each resource to provide this information.


aws/templates/id/id_apigateway.ftl
```diff
@@ -1,32 +1,33 @@
...
     [#return formatDependentResourceId(
-                APIGATEWAY_APIKEY_RESOURCE_TYPE,
+                AWS_APIGATEWAY_APIKEY_RESOURCE_TYPE,
                 resourceId,
                 extensions)]
 [/#function]
 
+[#-- Component --]
+[#assign APIGATEWAY_COMPONENT_TYPE = "apigateway"]
+[#assign APIGATEWAY_COMPONENT_DOCS_EXTENSION = "docs"]
```
ml019: Think this is a resource extension

aws/templates/id/id_apigateway.ftl
```diff
@@ -1,32 +1,33 @@
...
 [#function formatDependentAPIGatewayAPIKeyId resourceId extensions...]
     [#return formatDependentResourceId(
-                APIGATEWAY_APIKEY_RESOURCE_TYPE,
+                AWS_APIGATEWAY_APIKEY_RESOURCE_TYPE,
                 resourceId,
                 extensions)]
 [/#function]
 
+[#-- Component --]
+[#assign APIGATEWAY_COMPONENT_TYPE = "apigateway"]
```
ml019: maybe move this to the top before the resource definitions. Suspect this will eventually move out to its own file for the "component" which has the logic to work out which provider is to be used, and then it would call into the provider specific file to do the work.

aws/templates/id/id_contenthub.ftl
```diff
@@ -1,9 +1,11 @@
...
-[#assign CONTENTHUB_NODE_COMPONENT_TYPE = "contentnode"]
 
-[#assign CONTENTHUB_RESOURCE_TYPE = "contenthub"]
-[#assign CONTENTNODE_RESOURCE_TYPE = "contentnode"]
+[#-- Resources --]
+[#assign COT_CONTENTHUB_HUB_RESOURCE_TYPE = "contenthub"]
+[#assign COT_CONTENTHUB_NODE_RESOURCE_TYPE = "contentnode"]
+
+[#-- Components --]
+[#assign CONTENTHUB_HUB_COMPONENT_TYPE = "contenthub"]
```
ml019: same comment re posiitioning at the top

aws/templates/id/id_contenthub.ftl
```diff
@@ -47,6 +50,9 @@
     ]
 [/#function]
 
+
+[#assign CONTENTHUB_NODE_COMPONENT_TYPE = "contentnode"]
+
 [#assign componentConfiguration +=
     {
         CONTENTHUB_NODE_COMPONENT_TYPE : [
```
ml019: Index needs provider in value I think - its the specific component config IF you are using AWS.

Don;'t know if we would ever end up with the same component type in two different providers in a single solution, but for consistency anyting related to a specific provider should have AWS_ on the front I think?

aws/templates/id/id_apigateway.ftl
roleyfoley: The Docs are part of the component deployment though. They aren't part of API Gateway itself. The S3 bucket is the resource and in the API Gateway component it gets a docs extension.

aws/templates/id/id_apigateway.ftl
roleyfoley: Yeah I guess the component is worked on more than the resources. Or perhaps the resource Id sections should move to the resource file. The Id's/names will shrink as we convert everything so I think ID/name could fit in the resource section ok.

aws/templates/id/id_contenthub.ftl
roleyfoley: So the solution would be the point that the provider component is selected? And the component is the codeontap definition of a function for a provider?

aws/templates/common.ftl
```diff
@@ -743,6 +751,10 @@
                 [#local result = getFunctionState(occurrence)]
                 [#break]
 
+            [#case "ec2"]
+                [#local result = getEc2State(occurrence)]
```
ml019: Maybe EC2 instead of Ec2 for consistency?

aws/templates/id/id_ecs.ftl
```diff
@@ -27,6 +26,12 @@
                 component)]
 [/#function]
 
+
+[#-- Components --]
+[#assign ECS_COMPONENT_TYPE = "ecs" ]
+[#assign SERVICE_COMPONENT_TYPE = "service" ]
```
ml019: I think maybe subcomponent types should include the parent e.g. ECS_SERVICE_COMPONENT_TYPE?

aws/templates/solution/solution_ElasticSearch.ftl
```diff
@@ -1,125 +1,129 @@
...
-    [#-- In order to permit updates to the security policy, don't name the domain. --]
-    [#-- Use tags in the console to find the right one --]
-    [#-- "DomainName" : "${productName}-${segmentId}-${tierId}-${componentId}", --]
+        [#assign storageProfile = getStorage(tier, component, "ElasticSearch")]
+        [#assign volume = (storageProfile.Volumes["codeontap"])!{}]
+        [#assign esCIDRs =
+                    getUsageCIDRs(
+                        "es",
+                        configuration.IPAddressGroups)]
+        [#list zones as zone]
```
ml019: This code should be converted to use links to the IP component in other products. Links would make it much cleaner

PR #245, opened 2018-03-29
Generate Encrypted Password on RDS creation

### LINKED ISSUE #207: Add automated password encryption during deployment
For services like RDS, the master password has to be provided in a clear text format within the Cloudformation template. Given that we store cloudformation templates in a repository this is not following security best practices. 

Instead we should look at an alternate way to do the following: 
- Generate a Password and encrypt it using KMS
- Set a dummy password in the cloudformation template 
- Update the password in using the generated password

roleyfoley:
Adds support for a generated password to be created for an RDS instance. With this configuration the password is always encrypted and stored as a stack output.

When generateCredentials is enabled a dummy password is supplied to cloudformation and on the initial creation of the instance a password is generated, encrypted and set on the RDS instance as an epilogue script.

Since cloudformation resets the password to a dummy password each time on each deploy the password will be retrieved decrypted and reset as part of the deployment process

**This process will create some downtime during the deployment as the password will be set to the dummy password in between the cloudformation deployment and the epilogue script reseting the password back to the generated password**

Credentials are encrypted using the the segment KMS key. 

Configuration ( under rds solution configuration ) 

```
{
   "GenerateCredentials" : {
              "Enabled" : true|false,
              "CharacterLength" : Int (defaults to 20),
              "MasterUserName" : string (defaults to root)
   }
}
```
Username moves into config as the credentials will now be a state based configuration. 

To reset a password the epilogue pseudo stack can be removed. 

Fixes #207

### DIFF
--- a/aws/templates/id/id_rds.ftl
+++ b/aws/templates/id/id_rds.ftl
@@ -34,6 +34,23 @@
                 "Name" : "Encrypted",
                 "Default" : false
             },
+            {
+                "Name" : "GenerateCredentials",
+                "Children" : [
+                    {
+                        "Name" : "Enabled",
+                        "Default" : false
+                    },
+                    {
+                        "Name" : "MasterUserName",
+                        "Default" : "root"
+                    }
+                    {
+                        "Name" : "CharacterLength",
+                        "Default" : 20
+                    }
+                ]
+            },
             { 
                 "Name" : "Size",
                 "Default" : "20"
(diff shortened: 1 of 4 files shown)


aws/templates/solution/solution_rds.ftl
```diff
@@ -117,23 +132,22 @@
...
-                            "}",
-                            "convert_plaintext_snapshot || return $?"
-                        ],
-                        []
-                    )
-                /]
-            [/#if]
+                    ],
+                    []) +
+                (configuration.Backup.SnapshotOnDeploy && configuration.Encrypted)?then(
```
ml019: Thought this was already merged?

aws/utility.sh
```diff
@@ -540,6 +554,25 @@ function addJSONAncestorObjects() {
   runJQ "${pattern}" < "${file}"
 }
 
+# -- KMS --
+function decrypt_kms_string() { 
+  local value="$1"; shift
+
+  local tmpdir="$( getTempDir "kms_XXX" )"
```
ml019: Suggest using the new pushTempDir function - it tries to keep all the temp dirs under a single root dir

aws/templates/id/id_rds.ftl
```diff
@@ -34,6 +34,23 @@
                 "Name" : "Encrypted",
                 "Default" : false
             },
+            {
+                "Name" : "GenerateCredentials",
+                "Children" : [
+                    {
+                        "Name" : "Enabled",
+                        "Default" : false
```
ml019: Thinking default should be true? Just means we'd need to force a redeploy of anything that needs the password

aws/templates/solution/solution_rds.ftl
roleyfoley: Moved it around, the whole prologue subset was wrapped in an if loop for the snapshot on deploy. Felt better that the deployment subset was always there and the necessary sections were added.

aws/templates/id/id_rds.ftl
roleyfoley: Maybe we could leave it disabled now and enable it on the first train release. Then we can do the password refreshes as part of our hardware refreshes?

aws/utility.sh
roleyfoley: Couldn't find pushTempDir...?

ml019: Ah - yes. Added it in the stuff I'm working on. Dno't worry about it for now. But basically rather than a heap of tmpdirs in the top level, I've added a pushTempDir command like getTempDir but it basically maintains a stack of tmpdirs. If its used, the routine then PopTempDir's when its done. Just means we only get one top level dir in tmp for a given build rather than a million of them

aws/templates/id/id_rds.ftl
ml019: Sounds like a plan. Good to start with it off so we can deploy without breaking anything

PR #252, opened 2018-04-05
Add Support for epilogue based Lambda triggers in Cognito User Pools

roleyfoley:
Cognito user pools have the ability to trigger a lambda function based on a specific event that has occurred. The majority of the events are available through cloudformation except for some that have been added recently (Jan 2018).

This adds support for the cognito events which are available via the cli instead of cloudformation. 

The lambda event mappings are added to two seperate arrays: 
- userPoolTriggerConfig - is the array for cloudformation 
- userPoolManualTriggerConfig - is the array for cli based updates

### DIFF
--- a/aws/templates/solution/solution_userpool.ftl
+++ b/aws/templates/solution/solution_userpool.ftl
@@ -1,5 +1,5 @@
 [#-- Cognito User Pool --]
-[#if (componentType == USERPOOL_COMPONENT_TYPE) && deploymentSubsetRequired("userpool", true)]
+[#if (componentType == "userpool") ]
     
     [#list requiredOccurrences(
         getOccurrences(tier, component),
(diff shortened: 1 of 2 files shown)


aws/templates/solution/solution_userpool.ftl
```diff
@@ -190,125 +173,189 @@
                                     )
                                 ]
                                 [#break]
+                            [#case "pretokengeneration"]
+                                [#assign userPoolManualTriggerConfig = userPoolManualTriggerConfig +
```
ml019: FYI there is the += syntax when adding to an existing varialbe

aws/templates/solution/solution_userpool.ftl
```diff
@@ -190,125 +173,189 @@
...
+                authenticatedRoleArn=getReference(identityPoolAuthRoleId, ARN_ATTRIBUTE_TYPE)
+                unauthenticatedRoleArn=getReference(identityPoolUnAuthRoleId, ARN_ATTRIBUTE_TYPE)
+            /]
+        [/#if]
+        
+        [#if deploymentSubsetRequired("epilogue", false)]
+            [@cfScript
+                mode=listMode
+                content=
+                [] +
```
ml019: Is this empty array needed? The following function returns an empty array if there isn't any content

PR #255, opened 2018-04-06
Add custom domain for api gateway signing

ml019:
Support addition of a base path mapping when using cloudfront in front
of the api gateway so the fqdn used in the signature doesn't vary when
the api gateway is recreated.

Also remove the need for tier/component when generating a host name, as
these are already present in the occurrence structure.

### DIFF
--- a/aws/templates/id/id_apigateway.ftl
+++ b/aws/templates/id/id_apigateway.ftl
@@ -105,8 +105,26 @@
                 ]
             },
             {
-                "Name" : "IncludePathInUrls",
-                "Default" : true
+                "Name" : "Mapping",
+                "Children" : [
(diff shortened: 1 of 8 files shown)


aws/templates/id/id_apigateway.ftl
```diff
@@ -105,8 +105,26 @@
...
+                        "Default" : true
+                    }
+                ]
+            },
+            {
+                "Name" : "Service",
+                "Children" : [
+                    {
+                        "Name" : "Name",
+                        "Default" : "execute-api"
```
roleyfoley: Should this be configurable? I would have thought this was a fixed attribute?

PR #260, opened 2018-04-07
Feature appsettings in occurrences

ml019:
This is a pretty major hack to move the management of settings and credentials on to the occurrence.

It has involved reworking how appsettings, credentials and files are provided to the template engine.

In the process, I've tried to move away from the use of deployment units for settings management and more towards occurrences, so potentially a single deployment unit deploys multiple occurrences.

It also means container fragments should be much simpler, if needed at all as all settings are available by default to the application.

There's a bit more work to manage permissions automatically,however I think it worth a review.

Along the way, I tweaked the logging framework so it correctly handles only logging messages required by a particular level.

Hopefully I've rebased everything ok but let me know if you think I've dropped some critical changes.

This change also lays the groundwork for multi-account/multi-product processing, at least as far as settings management goes.

Build references are now integrated with settings, but there is migration to be performed in the cmdb.
I've made it non-destructive so it should be safe to run it each time.

I've added a routine to check if the cmdb needs to be upgraded with the intention that we can evolve this going forward whenever a cmdb change can be automated.


ml019:
@roleyfoley only thing missing now is the addition of permissions to decrypt and access the operations bucket to pick up asFiles if provided. 

I'm wondering if its simpler to always permit a component to access its part of the data and operations buckets. While I have the asFile stuff to use of a flag for accessing the ops bucket, there's nothing equivalent on the data bucket side.

aws/templates/commonApplication.ftl
```diff
@@ -211,6 +211,23 @@
...
+        s3ReadPermission(operationsBucket, getSettingsFilePrefix(occurrence)) +
+        s3AllPermission(dataBucket, getAppDataFilePrefix(occurrence)) +
+        valueIfContent(
+            s3AllPermission(dataBucket, getAppDataPublicFilePrefix(occurrence)),
+            getAppDataPublicFilePrefix(occurrence),
+            []
+        )
+    ]
+[/#function]
+
```
roleyfoley: Should cloudwatch be a default permission as well? If everything is going to log to a Cloudwatch prefix then its available if anything can use it.

PR #275, opened 2018-04-12
Update Cognito epilogue processing to include all custom configuration

roleyfoley:
When we run the epilogue script to set lambda triggers any non default configuration is lost that was set by cloudformation. 

After contacting AWS it appears that when the update-user-pool command is run an property not set in the command is treated as a null value which reverts everything back to its default configuration. So to get around this we need to reaply all configuration on the poll as part of the update. 

The update-user-pool supports the client skeleton json formatting so we can reuse some of the cloudformation functions for the templates to generate the json skeleton.


roleyfoley:
Updated PR to introduce the idea of a cli pass. The cli pass is intended to work with the AWS skeleton cli json objects. These objects allow you to pass parameters to a aws cli command using a json object. 

The cli pass generates a new json object which contains a collection of objects, at the root of the tree is an object with the resource name and at the next level is a command name, within the command name is the content generated in the macro. 

This structure allows for a resource to be configured with multiple commands using different config sets.

aws/setStackContext.sh
```diff
@@ -162,6 +162,15 @@ ALTERNATIVE_CONFIGS=$(findFiles \
     "${CF_DIR}/${LEVEL_PREFIX}${DEPLOYMENT_UNIT_PREFIX}${DEPLOYMENT_UNIT_SUBSET_PREFIX}${REGION_PREFIX}*-config.json" \
     "${CF_DIR}/${LEVEL_PREFIX}${DEPLOYMENT_UNIT_PREFIX}${DEPLOYMENT_UNIT_SUBSET_PREFIX}${ACCOUNT_PREFIX}${REGION_PREFIX}*-config.json" )
 
+CLI="${LEVEL_PREFIX}${DEPLOYMENT_UNIT_PREFIX}${DEPLOYMENT_UNIT_SUBSET_PREFIX}${REGION_PREFIX}cli.json"
```
ml019: As cli is new, there's no need for this conditional logic - it should always include the account id

aws/utility.sh
```diff
@@ -501,6 +501,18 @@ function addToArrayHead() {
   addToArrayHeadWithPrefix "${array}" "" "${elements[@]}"
 }
 
+# -- Cli file generation -- 
+function split_cli_file() {
+  local cli_file="$1"; shift
+  local outdir="$1"; shift
+
+  for resource in $(cat ${cli_file} | jq -r 'keys[]') ; do
```
ml019: better to use input redirect rather than cat.

aws/utility.sh
```diff
@@ -501,6 +501,18 @@ function addToArrayHead() {
   addToArrayHeadWithPrefix "${array}" "" "${elements[@]}"
 }
 
+# -- Cli file generation -- 
+function split_cli_file() {
+  local cli_file="$1"; shift
+  local outdir="$1"; shift
+
+  for resource in $( jq -r 'keys[]' <"${cli_file}" ) ; do
+    for command in $( cat ${cli_file} | jq -r ".$resource | keys[]" ); do
```
ml019: redirect here too

aws/utility.sh
```diff
@@ -501,6 +501,18 @@ function addToArrayHead() {
...
 }
 
+# -- Cli file generation -- 
+function split_cli_file() {
+  local cli_file="$1"; shift
+  local outdir="$1"; shift
+
+  for resource in $( jq -r 'keys[]' <"${cli_file}" ) ; do
+    for command in $( cat ${cli_file} | jq -r ".$resource | keys[]" ); do 
+        cat "${cli_file}" | jq ".${resource}.${command}" > "${outdir}/cli-${resource}-${command}.json"
```
ml019: and here. Basically void pipes where possible

PR #276, opened 2018-04-14
EFS  Improvements, NLB Support, Jenkins Container

roleyfoley:
Epic https://github.com/codeontap/codeontap/issues/14

A collection of changes introduced to support Jenkins as a product 

- New 
  - Adds a new container fragment to build a jenkins master on ECS/Docker. The container expects an ECS cluster with clustershared storage (EFS) to be enabled.
  - Adds Linking support for EFS Volumes to ensure that the EFS volume will be retained. Previously the EFS volume was created as part of the ECS deployment. This removes this functionality and instead requires you to link to an existing EFS volume
  - Adds EFS Mounts as a subcomponent which list a directory on the EFS volume to mount to 
  - Adds EFS Encryption at rest support and encryption in transit using the amazon-efs-utils package (wraps the EFS mount and traffic in a TLS tunnel 
  - Adds NLB support for ALB - The switch between the two is based on the protocols listed for the ports configured on an ALB
- Minor fixes 
  - Freemarker formats numbers with , for values over 1000 (1,000). This causes issues with Ids. the , value will be replaced with nothing in the formatId Function 

Depends On: https://github.com/codeontap/gen3-startup/pull/11

### DIFF
--- a/aws/templates/container/container_jenkins.ftl
+++ b/aws/templates/container/container_jenkins.ftl
@@ -17,9 +17,9 @@
                     description="Login Details not provided"
                     context=component
                     detail={
-                        "Login" : {
-                            "Username" : "",
-                            "Password" : ""
+                        "Jenkins" : {
+                            "User" : "",
+                            "Pass" : ""
                         }
                     }
                 /]
@@ -32,10 +32,10 @@
                     description="Github oAuth Credentials not provided"
                     context=component
                     detail={
-                        "Github" : {
+                        "GithubAuth" : {
                             "ClientId" : "",
                             "Secret" : "",
-                            "Administrators" : ""
+                            "Admin" : ""
                         }
                     }/]
             [/#if]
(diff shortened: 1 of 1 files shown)


aws/templates/container/container_jenkins.ftl
```diff
@@ -0,0 +1,72 @@
+[#case "jenkins"]
+    [@Attributes image="jenkins-master" /]
+   
+    [#assign credentials = credentialsObject]
```
ml019: This will break when we merge the settings stuff. I've left the credentialsObject in place but ideally it should be picking stuff up from the settings. Also the credentials will automatically be in the context so no need to use @Variables

aws/templates/container/container_jenkins.ftl
```diff
@@ -0,0 +1,72 @@
...
+            [#if (credentials.Login.Username)?has_content || (credentials.Login.Password)?has_content ]
+                [@Variables
+                    {
+                        "JENKINS_USER"          : credentials.Login.Username,
+                        "JENKINS_PASS"          : credentials.Login.Password,
+                        "JENKINS_SECURITYREALM" : credentials.SecurityRealm
+                    }
+                /]
+            [#else]
+                [@cfException
```
ml019: Could just use a value of "COTException: ..." as this will also be detected

PR #286, opened 2018-04-19
EC2 Occurrence Conversion, ALB Load Balancing Refactor, Deploy Script

roleyfoley:
Updates the S3 Solution component to align with use some of our new features 

- Registration with ALB based load balancers using linking/port definitions
- Converting ID/naming generation to support occurrence structure adds the idea of Zone level resources which individual resources for each zone that could potentially be deployed to 
- Adds the ability to deploy scripts to an Ec2 Instance
  
**To Deploy a script** 

1. create a code repo with an init.sh script as an entry point with other scripts or content. 
2. The script repo content will be copied to the ec2 instances under /opt/codeontap/scripts 
3. The /opt/codeontap/scripts/init.sh script will then be run on the inital setup of the instance as part of the cfn-init process 
4. The instance will be recreated each time the code is updated. Make sure that the data does not require persistence.

### DIFF
--- a/aws/templates/common.ftl
+++ b/aws/templates/common.ftl
@@ -783,7 +783,7 @@
                 [#local result = getContentNodeState(occurrence)]
                 [#break]
 
-            [#case "ec2"]
+            [#case EC2_COMPONENT_TYPE]
                 [#local result = getEC2State(occurrence)]
                 [#break]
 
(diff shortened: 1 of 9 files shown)


aws/templates/solution/solution_ec2.ftl
```diff
@@ -151,26 +232,32 @@
...
-                                                        "echo \"cot:component="     + componentId            + "\"\n",
-                                                        "echo \"cot:zone="          + zone.Id                + "\"\n",
-                                                        "echo \"cot:name="          + formatName(ec2FullName, zone) + "\"\n",
-                                                        "echo \"cot:role="          + component.Role         + "\"\n",
-                                                        "echo \"cot:credentials="   + credentialsBucket      + "\"\n",
-                                                        "echo \"cot:code="          + codeBucket             + "\"\n",
-                                                        "echo \"cot:logs="          + operationsBucket       + "\"\n",
-                                                        "echo \"cot:backups="       + dataBucket             + "\"\n"
-                                                    ]
+                                                        "#!/bin/bash\\n",
```
ml019: these need adjustment. there was a change to getJSON so that you don't need all the extra backslashes

aws/templates/common.ftl
```diff
@@ -1767,6 +1767,92 @@
 
 [/#function]
 
+[#function getLBLink occurrence port ]
```
ml019: Is this being used by getContainers as well?  If adding a generic function, shoul dupdate its use in the other places its needed to make sure the function signature is reusable.

aws/templates/solution/solution_ec2.ftl
roleyfoley: Fixed up

aws/templates/common.ftl
roleyfoley: Added to the containers function as well.

PR #288, opened 2018-04-20
Filter LogMetrics using Name instead of dimension

roleyfoley:
The intention of metric dimensions is to filter a metric to the source that generated it. This allows you to have the same metric but reported from different locations. 

Unfortunately dimensions are not available for LogMetrics. After raising a support case with AWS they have advised that at the moment you can not add dimensions to LogMetrics and instead to filter the metric the metric name should include a unique Id. 

This PR updates the metric name to make it unique for each function and then updates the alarm Metric to include the function name. 

The intention would be to move this back to dimensions when they are supported by AWS.

### DIFF
--- a/aws/templates/application/application_lambda.ftl
+++ b/aws/templates/application/application_lambda.ftl
@@ -189,7 +189,7 @@
                             [@createLogMetric
                                 mode=listMode
                                 id=formatDependentLogMetricId(fnId, metric.Id)
-                                name=metric.Name
+                                name=formatName(metric.Name, fnName) 
                                 logGroup=logGroupName
                                 filter=metric.LogPattern
                                 namespace=formatProductRelativePath()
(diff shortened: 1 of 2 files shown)


aws/templates/application/application_lambda.ftl
```diff
@@ -204,19 +204,16 @@
...
-                                        "Value" : logGroupName
-                                    }
-                                ]
-                            ]
-                        [#break]
-                    [/#switch]
+                            [#-- TODO: Ideally We should use dimensions for filtering but they aren't available on Log Metrics --]
+                            [#-- feature requst has been reaised... --]
+                            [#-- Instead we name the logMetric with the function name and will use that --]
+                            [#assign metricName = formatName(alert.Metric.Name, fnName) ]
```
ml019: metricName = formatName(metricName, fnName)

Also shouldn't fnNAme came first? Not sure how this affects sorting in the console...

roleyfoley: IT would let you see the same metrics grouped together for each environment. This would be the way you look for them with dimensions, you go to the metric then filter it based on dimension to get the details you are after

PR #291, opened 2018-04-20
Correct syntax error

ml019:
Also spelling of variable

Not sure how this one made it through - framework basically didn't run


ml019:
Its ok - I then thought I'd corrected it but it still didn't run  - lol

PR #301, opened 2018-04-30
Fix LB Healthcheck lookup

roleyfoley:
The health check configuration is on the LB component instead of the listener subcomponent. This fix moves the health check lookup to the LB component so that its properties can be looked up properly

### DIFF
--- a/aws/templates/solution/solution_lb.ftl
+++ b/aws/templates/solution/solution_lb.ftl
@@ -164,11 +164,11 @@
                 
                 [#case "classic"]
                 
-                    [#if !healthCheckPort?has_content ]
-                        [@cfPreconditionFailed listMode "solution_lb" {} "No health check port provided" /]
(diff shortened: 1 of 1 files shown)


aws/templates/solution/solution_lb.ftl
```diff
@@ -20,6 +20,13 @@
 
             [#assign engine = solution.Engine]
 
+            [#assign healthCheckPort = "" ]
+            [#if engine == "classic" && solution.HealthCheckPort?has_content ]
+                [#assign healthCheckPort = ports[solution.HealthCheckPort]]
+            [#else]
+                [@cfPreconditionFailed listMode "solution_lb" {} "No health check port provided" /]
```
ml019: Is this logic right? I  thought the "HealthCheckPort was only added to support the classic lb but this code will catch all lb types?

PR #305, opened 2018-05-01
Support environment and segment in names

ml019:
Move to using the environment by default the segment only if it differs
from the environment and does not have the value of "default".

The next stage will be to change the cmdb layout to explicitly include
the environment in the tree to allow for multi-segment environments.

NOTE: This change may need adjustment to segment.json files where the
environment is not the same as the segment.

### DIFF
--- a/aws/templates/segment/segment_dns.ftl
+++ b/aws/templates/segment/segment_dns.ftl
@@ -8,10 +8,10 @@
         properties=
             {
                 "HostedZoneConfig" : {
-                    "Comment" : formatName(productName, segmentName)
+                    "Comment" : formatSegmentFullName()
                 },
                 "HostedZoneTags" : getCfTemplateCoreTags(),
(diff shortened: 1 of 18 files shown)


aws/templates/segment/segment_dns.ftl
```diff
@@ -8,10 +8,10 @@
         properties=
             {
                 "HostedZoneConfig" : {
-                    "Comment" : formatName(productName, segmentName)
+                    "Comment" : formatSegmentFullName()
                 },
                 "HostedZoneTags" : getCfTemplateCoreTags(),
-                "Name" : segmentName + "." + productName + ".internal",
+                "Name" : concatenate(fullNamePrefixes?reverse + ["internal], "."),
```
roleyfoley: should internal have a closing "?

ml019: yep

PR #308, opened 2018-05-11
fix for IP Address Group action lookup

roleyfoley:
Minor fix for IPAddress Group action lookup. 

The lookup at the moment is currently using the raw group array instead of the IP Address Group that has been looked up


ml019:
I have a fix for this coming with the environment+segment support. The included code won't fix the issue - it has to do with handling different array entry types (string/object) in the IPAddressGroups config parameter. I'd suggest we close this PR.

PR #317, opened 2018-05-17
add proxy+ path support

kshychko:
generate an extended swagger specification with a "/{proxy+}" path to pass through the requests to invalid urls and get 404 instead of 403
see https://forums.aws.amazon.com/thread.jspa?threadID=216684 for details


kshychko:
-p option shall be removed, flag value should be obtained from apigw.json

ml019:
It seems unnecessary to extra the proxy setting from the INTEGRATION_FILE and pass as a separate parameter given that the entire integration file is pass to the template anyway.

I think the only change needs to be to the ftl - unless I've missed something?

kshychko:
@ml019 you made a point. I rolled back the changes for the script and add an integration variable read in the template. Forced pushed the PR, please review.

PR #319, opened 2018-05-19
Add Compute Cluster component and refactor init configuration

roleyfoley:
Fixes https://github.com/codeontap/codeontap/issues/16

#- **New Component: Compute Cluster**
  - A compute cluster is an autoscaling ec2 level application deployment it allows for the deployment of code at the ec2/virtual machine level. 
  - Deployment: the root directory of the code must contain a script called init.sh which is called on startup as root of an ec2 instance in the auto scale group. Any other code that is saved in the same location of the init.sh or subdirectires will be zipped and copied to the ec2 instance in /opt/codeontap/scripts
  - Code Deployment: A rolling update or replacement (depending on configuration of compute cluster)  will be triggered on each code update. This works by updating the launch config logical cloudformation Id with the build reference, this forces cloudformation to start a refresh of the ASG 
  - Configuration: 
    - MinUpdateInstances - int -  The minimum number of instances that must always be available during an update ( if you set to 0 a single node cluster will go down during update) 
    - ReplaceOnUpdate - boolean -  If enabled, when the launch config of the group is updated a new ASG will be created and when up and running the existing cluster will be destroyed 
    - DockerHost - boolean - Allows for the use of the ECS optimised instance 
    - Links - As per standard Link configuration
    - Ports - As per standard Link configuration 
    - Added Processor Profile and Storage Profile for ComputeCluster
- **Refactor of Ec2 instance Cloudformation Init configuration** 
  - Since cloudformation init scripts are used in a number of locations now they have been refactored into functions for each configKey that we use. They can then be turned into a configSet using the getInitConfig function 
- **Utility function for string cleanup -replaceAlphaNumericOnly** 
  - To make sure that the logical Id used in the compute cluster launch config matches the logical id cloudformation requirements i.e. alphanumeric only a regex based replace has been added which will replace all no alphanumeric characters with a given character ( defaults to empty)

### DIFF
--- a/aws/templates/application/application_computecluster.ftl
+++ b/aws/templates/application/application_computecluster.ftl
@@ -25,8 +25,7 @@
         [#assign loadBalancers = [] ]
         [#assign environmentVariables = {}]
 
-        [#assign configSetName = componentType]
-        [#assign configSets = {} ]
+        [#assign configSetName = componentType]        
         [#assign configSets +=  
                 getInitConfigDirectories() + 
                 getInitConfigBootstrap(component.Role!"") ]
@@ -47,10 +46,16 @@
             )
         ]
 
+        [#assign containerId =
+            solution.Container?has_content?then(
+                solution.Container,
+                getComponentId(component)
+            ) ]
+
         [#assign environmentContext =
             {
-                "Id" : core.Id,
-                "Name" : core.Name,
+                "Id" : containerId,
+                "Name" : containerId,
                 "Instance" : core.Instance.Id,
                 "Version" : core.Version.Id,
                 "DefaultEnvironment" : defaultEnvironment(occurrence),
--- a/aws/templates/application/application_computecluster.ftl
+++ b/aws/templates/application/application_computecluster.ftl
@@ -23,6 +23,8 @@
         [#assign computeClusterLogGroupName = resources["lg"].Name ]
 
         [#assign targetGroupPermission = false ]
+        [#assign targetGroups = [] ]
+        [#assign loadBalancers = [] ]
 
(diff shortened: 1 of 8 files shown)


aws/templates/solution/solution_ec2.ftl
```diff
@@ -24,22 +24,14 @@
         [#assign targetGroupRegistrations = {}]
         [#assign targetGroupPermission = false ]
 
-        [#assign efsMountPoints = {}]
+        [#assign configSetName = componentType ]
+        [#assign configSets = {} ]
```
ml019: Redundant - combine with next line

aws/templates/solution/solution_ecs.ftl
```diff
@@ -24,6 +24,13 @@
         [#assign ecsLogGroupName = resources["lg"].Name ]
         [#assign defaultLogDriver = solution.LogDriver ]
         [#assign fixedIP = solution.FixedIP ]
+
+        [#assign configSetName = componentType ]
+        [#assign configSets = {} ]
```
ml019: Redundant - combine with next line

aws/templates/application/application_computecluster.ftl
```diff
@@ -0,0 +1,265 @@
...
+                getRegistryPrefix("scripts", occurrence),
+                productName,
+                buildUnit,
+                buildReference
+                ) ]   
+
+        [#assign scriptsUrl = 
+            "https://" +
+            formatRelativePath(
+                getExistingReference(formatAccountS3Id("registry"),DNS_ATTRIBUTE_TYPE),
```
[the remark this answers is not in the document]
roleyfoley: Shouldn't be there now. That was a mistake

PR #344, opened 2018-06-14
Add Cognito ElasticSearch integration support

roleyfoley:
With AWS elastic search you can now use IAM based permissions to authenticate with ElasticSearch. This is fine for services which can easily generate sig4 Credentials but for users this is a bit more complicated. So AWS also added integration with Cognito, which lets you create users in a userpool which can be given IAM credentials to access ElasticSearch. 

This PR adds support for this setup in Codeontap and includes a few things
- Add a cognito domain name to a cognito user pool
- Add link support to Elastic Search 
- When a user pool is linked to from an elasticsearch component cognito will be enabled. This creates a service role which allows elasticsearch to setup cognito, and adds an access policy which allows the default auth user role to access kibana 

User management is still manual


aws/templates/policy/policy_es.ftl
```diff
@@ -0,0 +1,13 @@
+[#-- API Gateway --]
```
ml019: Needs fixing

aws/templates/solution/solution_ElasticSearch.ftl
```diff
@@ -51,33 +63,120 @@
...
+                            "AWS": "*"
+                        },
+                        attributeIfContent(
+                            "IpAddress",
+                            esCIDRs,
+                            {
+                                "aws:SourceIp": esCIDRs
+                            }) + 
+                        attributeIfTrue(
+                            "Null",
```
ml019: This looks a bit odd ... "Null" ???

PR #357, opened 2018-06-18
Add Encrypted variable prefix option to RDS Password Generation

roleyfoley:
Allows you to add a prefix to the encrypted strings created as part of the password generation process (e.g. base64:). This is used by some products to determine if the environment variable is encrypted or not.

### DIFF
--- a/aws/templates/id/id_rds.ftl
+++ b/aws/templates/id/id_rds.ftl
@@ -47,10 +47,14 @@
                     {
                         "Name" : "MasterUserName",
                         "Default" : "root"
-                    }
+                    },
                     {
                         "Name" : "CharacterLength",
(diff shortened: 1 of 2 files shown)


aws/templates/id/id_rds.ftl
```diff
@@ -47,10 +47,14 @@
...
                         "Name" : "MasterUserName",
                         "Default" : "root"
-                    }
+                    },
                     {
                         "Name" : "CharacterLength",
                         "Default" : 20
+                    },
+                    {
+                        "Name" : "VariablePrefix",
```
ml019: Can we make this a bit more specific? It only affects encrypted material so maybe 

"EncryptedVariablePrefix"  ?

PR #363, opened 2018-06-21
Add Support for Cognito Auth with Application Load Balancers

roleyfoley:
Adds support for a load balancer to authenticate clients using a cognito user pool (https://aws.amazon.com/blogs/aws/built-in-authentication-in-alb/) 

To support this a number of changes are required 
- UserPools
  - Configuration
     - OAuth - Scopes - The oAuth scopes presented to the client - Default openid
     - OAuth - Flows - The oAuth flows used for authentication - Default code (authorisation code grant) 
- LoadBalancer - Port
    - Configuration 
         - Path - The client side path which will match to a specific target group or action - Default default
         - Priority - If a rule needs to be created the priority of the rule against other rules ( lowest wins) - Default 100 
         - Links - Adds standard links
         - Authentication - SessionCookieName - The cookie used to determine if authentication is required - Default AWSELBAuthSessionCookie
         - Authentication - SessionTimeout - Default 7 days
- Consolidation of  all load balancing configuration into the LB solution. Paths can now be defined using multiple Port Mappings that use the same mapping. A default port mapping must be created in order to create additional  paths 
Example: 
```
                    "LB" : {
                        "Logs" : true,
                        "Engine" : "application",
                        "PortMappings" : {
                            "http" : {
                                "TargetType" : "ip"
                            },
                            "https" : {
                                "Links" : {
                                    "feed-userpool" : {
                                        "Tier" : "app",
                                        "Component" : "feed-userpool",
                                        "Instance" : "",
                                        "Version" : ""
                                    }
                                }
                            },
                            "httpsadmin" : {
                                "Mapping" : "https",
                                "Path" : "/admin"
                            }
                        },
                        "Instances" : {
                            "default" : {
                                "DeploymentUnits" : ["redirector-elb"]
                            }
                        },
                        "IPAddressGroups" : ["global"]
                    }
```
- In this LB configuration the http and httpsadmin both use the same mapping (the https gets its mapping from its key name). The httpsadmin has a path set and a Mapping set. The Path rule doesn't create a listener or security group and uses the default ports listener and security group.

- A target (ec2, computecluster, container) needs to link to the LB Port, this link will provide the target group that the target needs to register to. The target is responsible for registering to the target group and the client side configuration is looked after by the LB. 

- To enable Cognito Authentication two links are required, one from the user pool to the LB and one from the LB to the userpool. The Userpool to LB link gets the callbackURLs required by cognito and the LB to Userpool link configures the listener rule to enable the authentication. 

- When using LB user pool must have "ClientGenerateSecret" set to true

### DIFF
--- a/aws/templates/application/application_computecluster.ftl
+++ b/aws/templates/application/application_computecluster.ftl
@@ -159,7 +159,7 @@
                             id=formatDependentSecurityGroupIngressId(
                                 resources["securityGroup"].Id
                                 link.Id)
-                            port=link.DestinationPort
+                            port=linkTargetAttributes["DESTINATION_PORT"]
                             cidr=linkTargetResources["sg"].Id
                             groupId=computeClusterSecurityGroupId /]
                     [/#if]
--- a/aws/templates/application/application_computecluster.ftl
+++ b/aws/templates/application/application_computecluster.ftl
@@ -159,7 +159,7 @@
                             id=formatDependentSecurityGroupIngressId(
                                 resources["securityGroup"].Id
                                 link.Id)
-                            port=link.Port
+                            port=link.DestinationPort
                             cidr=linkTargetResources["sg"].Id
                             groupId=computeClusterSecurityGroupId /]
                     [/#if]
--- a/aws/templates/application/application_computecluster.ftl
+++ b/aws/templates/application/application_computecluster.ftl
@@ -169,37 +169,9 @@
                         [#case "application"]
                         [#case "network"]
                             [#if link.TargetGroup?has_content ]
-                                [#assign targetId = (linkTargetResources["targetgroups"][link.TargetGroup].Id) ]
-                                [#if targetId?has_content]
-
-                                    [#if deploymentSubsetRequired(COMPUTECLUSTER_COMPONENT_TYPE, true)]
-                                        [#if isPartOfCurrentDeploymentUnit(targetId)]
-
-                                            [@createTargetGroup
-                                                mode=listMode
-                                                id=targetId
-                                                name=formatName(linkTargetCore.FullName,link.TargetGroup)
-                                                tier=link.Tier
-                                                component=link.Component
-                                                destination=ports[link.Port]
-                                            /]
-                                            [#assign listenerRuleId = formatALBListenerRuleId(occurrence, link.TargetGroup) ]
-                                            [@createListenerRule
-                                                mode=listMode
-                                                id=listenerRuleId
-                                                listenerId=linkTargetResources["listener"].Id
-                                                actions=getListenerRuleForwardAction(targetId)
-                                                conditions=getListenerRulePathCondition(link.TargetPath)
-                                                priority=link.Priority!100
-                                                dependencies=targetId
-                                            /]
-
-                                            [#assign componentDependencies += [targetId]]
-
-                                        [/#if]
-                                        [#assign targetGroups += [ getReference(targetId, ARN_ATTRIBUTE_TYPE) ] ]
-                                    [/#if]
-                                [/#if]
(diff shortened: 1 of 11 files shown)


roleyfoley:
Still needs a little more testing with our default cases. But seems to be working pretty well so far. Just wanted to get the PR out in the wild

roleyfoley:
@ml019 This should be now ready to go. The Id's of the components change around a fair bit with this update and with the move of the listener rules from the load balanced components to the load balancers we might need to shuffle a few deploys.

aws/templates/application/application_computecluster.ftl
```diff
@@ -159,7 +159,7 @@
                             id=formatDependentSecurityGroupIngressId(
                                 resources["securityGroup"].Id
                                 link.Id)
-                            port=link.Port
+                            port=link.DestinationPort
```
ml019: I'd prefer to see the destinationPort as an attribute on the target and obtained from there. From what I can see, we have the target occurrence at the point where we need the destination port.

aws/templates/common.ftl
```diff
@@ -1891,61 +1891,9 @@
         } +
         attributeIfTrue("Instance", port.LB.Instance??, port.LB.Instance!"") +
         attributeIfTrue("Version",  port.LB.Version??, port.LB.Version!"") +
-        attributeIfContent("PortMapping",  portMapping)
+        attributeIfContent("PortMapping",  portMapping) + 
+        attributeIfContent("DestinationPort" , port.Name)
```
ml019: Not convinced about this. We need to keep link semantics clean - Dest port is really an attribute of the thing the link points to, rather than something needed to follow the link

aws/templates/common.ftl
```diff
@@ -1891,61 +1891,9 @@
...
-
-            [#if targetPath?has_content]
-                [#-- target group name must be provided if path provided --]
-                [#if !targetGroup?has_content]
-                    [@cfException
-                        listMode "No target group for provided path" occurrence /]
-                    [#local targetGroup = "default" ]
-                [/#if]
-            [#else]
-                [#if !targetGroup?has_content]
```
ml019: Good to see this logic go. It was always kind of clunky - moving all the tg stuff to the LB feels better.

aws/templates/id/id_lb.ftl
```diff
@@ -126,34 +149,42 @@
...
                 "LB" : lbId,
                 "ENGINE" : engine,
                 "FQDN" : fqdn,
-                "URL" : scheme + "://" + fqdn,
+                "URL" : url + path,
                 "INTERNAL_FQDN" : internalFqdn,
-                "INTERNAL_URL" : scheme + "://" + internalFqdn,
-                "PORT" : sourcePort.Name
+                "INTERNAL_URL" : internalUrl + path,
+                "PORT" : sourcePort.Name,
```
ml019: Think this is maybe where we need DESTINATION_PORT, and maybe change PORT to SOURCE_PORT for consistency

aws/utility.sh
```diff
@@ -708,7 +708,16 @@ function update_cognito_userpool() {
   local userpoolid="$1"; shift
   local configfile="$1"; shift
 
-  aws --region ${region} cognito-idp update-user-pool --user-pool-id "${userpoolid}" --cli-input-json "file://${configfile}"
+  aws --region "${region}" cognito-idp update-user-pool --user-pool-id "${userpoolid}" --cli-input-json "file://${configfile}"
+}
+
+function update_userpool_client() {
```
ml019: Time to refactor I think - utility.sh is supposed to be just that.

Pick this up when we re-organise for deploymentment framework/cloud provider

PR #364, opened 2018-06-23
Codeontap task container configuration

roleyfoley:
A few changes required for the codeontap task container 

- Remove all standard environment variables to prevent issues with environment variables defined during codeontap process
- Add ManagedPolicy Container fragment macro to attach predefined polices to task role
- Add codeontap container configuration to use a comma separated list env variable AWS_AUTOMATION_POLICIES to attach managed policies which have been defined for switching roles to other accounts 
- Add new AWSProfile volume to codeontap container which is used for a persistent aws cli config file. Config file maps to /var/opt/awsprofile/config and sets AWS_CONFIG_PROFILE variable to match
- Add the ability to configure Privileged mode to containers so that they can access the host (required for docker in docker scenarios) 
- Add the ability to add run capabilities which are more fine grained ways to access the host OS.

### DIFF
--- a/aws/templates/application/application_ecs.ftl
+++ b/aws/templates/application/application_ecs.ftl
@@ -242,13 +242,13 @@
             [#if solution.UseTaskRole]
                 [#assign roleId = resources["taskrole"].Id ]
                 [#if deploymentSubsetRequired("iam", true) && isPartOfCurrentDeploymentUnit(roleId)]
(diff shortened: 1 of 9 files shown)


roleyfoley:
@ml019 The deployment is now in place and I can't think of any other template changes that are required. Ready for review

PR #370, opened 2018-06-29
Move encryption prefix handling to run time instead of psuedo stack

roleyfoley:
The current implementation of the encryption prefix setup for generated credentials stored the prefix in the psuedo stack file. 

This isn't actually required and instead this should be performed at template generation time. This fix allows for passwords and URLs that have been generated with the prefix in the pseudo stack file to be left as they are. Urls that have the prefix will be automatically updated but passwords would need to be updated manually or reset the password

### DIFF
--- a/aws/templates/solution/solution_rds.ftl
+++ b/aws/templates/solution/solution_rds.ftl
@@ -73,12 +73,12 @@
             [#assign rdsPasswordLength = solution.GenerateCredentials.CharacterLength]
             [#assign rdsPassword = "DummyPassword" ]
             [#assign rdsEncryptedPassword = (
-                getExistingReference(
-                    rdsId, 
-                    GENERATEDPASSWORD_ATTRIBUTE_TYPE)
-                )?remove_beginning(
-                    passwordEncryptionScheme
-                )]
+                        getExistingReference(
+                            rdsId, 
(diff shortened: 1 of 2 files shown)


aws/templates/solution/solution_rds.ftl
```diff
@@ -295,6 +295,9 @@
         [#if deploymentSubsetRequired("epilogue", false)]
 
             [#assign rdsFQDN = getExistingReference(rdsId, DNS_ATTRIBUTE_TYPE)]
+
+            [#assign passwordPsuedoStackFile = "\"$\{CF_DIR}/$(fileBase \"$\{BASH_SOURCE}\")-password-pseudo-stack.json\"" ]
```
ml019: Suggest spelling pseudo correctly

aws/templates/solution/solution_rds.ftl
```diff
@@ -295,6 +295,9 @@
...
+            [#assign urlPsuedoStackFile = "\"$\{CF_DIR}/$(fileBase \"$\{BASH_SOURCE}\")-url-pseudo-stack.json\""]
```
ml019: Suggest spelling pseudo correctly

roleyfoley:
@ml019 Spelling fixes added. Could you please review again

PR #371, opened 2018-07-03
Add IP Address group control to  container ports

roleyfoley:
Allows for IP address group rules to be specified on a container port without a load balancer. 

This is useful in scenarios where a container registers with another service and its IP address is included in the registration and two way communication is required (both initiating) 

Also adds a name attribute as a placeholder attribute on services. This allows you to link to a service and access its settings/config via a container fragment.

please merge before #363 as this PR will cause conflicts with it.


roleyfoley:
Technically it would work on containers that aren't using awsvpc mode. If you had statically mapped ports you could add rules without an LB. Thinking stuff like admin ports.

roleyfoley:
A per our discussion I've added an exception when you use IPAddressGroups without awsvpc mode. This is an outside case and really only required for the awsvpc network type at the moment.

PR #400, opened 2018-07-23
Add DBParameters to RDS solution

roleyfoley:
Allows for parameters to be set on an RDS Instance


roleyfoley:
Worked through this with @ml019

PR #404, opened 2018-07-27
Mobile Notification Proxy

roleyfoley:
Adds initial support for mobile notifiers which are used to send push notifications to multiple mobile platforms 

## Configuration

A mobilenotifier resource containers a list of platforms, each of these platforms send notifications to a specific mobile platform, APNS (Apple), Google etc. 

Each platform is configured with an engine, we currently support: 
- APNS
- APNS_SANDBOX
- GCM
- SMS 

With the exception of SMS each of them require credentials provided by the mobile platform. For apple this is a certificate and private key and for GCM this is a shared secret. These need to be encrypted using KMS and stored as credentials, with the name format 
```
<platformId>_Principal - APNS only - this is the certificate provided by apple  - not required for GCM
<platformId_Credential - APNS - private certificate - GCM - shared secret 
```

## Attributes

Linking to a mobile notifier platform returns 3 attributes
- ARN - the arn of the platform application in SNS
- TOPIC_PREFIX - when creating a topic to send notifications the topic name must be prefixed with this prefix to ensure that topics don't overlap between platforms 
- ENGINE - the platform that this platform uses 

## Deployment
For AWS SNS platform applications are only available via the CLI. However the logging role must be created before deploying the component.

### DIFF
--- a/aws/templates/id/id_mobilenotifier.ftl
+++ b/aws/templates/id/id_mobilenotifier.ftl
@@ -1,0 +1,91 @@
+[#-- MOBILENOTIFIER --]
+
+[#-- Components --]
+[#assign MOBILENOTIFIER_COMPONENT_TYPE = "mobilenotifier" ]
+[#assign MOBILENOTIFIER_PLATFORM_COMPONENT_TYPE = "mobilenotiferplatform" ]
+
+[#assign componentConfiguration +=
+    {
+        MOBILENOTIFIER_COMPONENT_TYPE : {
+            "Attributes" : [
+                { 
+                    "Name" : "Links",
+                    "Subobjects" : true,
+                    "Children" : linkChildrenConfiguration
+                },
+                {
+                    "Name" : "SuccessSampleRate",
+                    "Default" : 100
+                }
+            ],
+            "Components" : [
+                {
+                    "Type" : MOBILENOTIFIER_PLATFORM_COMPONENT_TYPE,
+                    "Component" : "Platforms",
+                    "Link" : [ "Platform" ]
+                }
+            ]
+        },
+        MOBILENOTIFIER_PLATFORM_COMPONENT_TYPE : [
+            {
+                "Name" : "Engine"
+            },
+            {
+                "Name" : "SuccessSampleRate"
+            },
+            { 
+                "Name" : "Links",
+                "Subobjects" : true,
+                "Children" : linkChildrenConfiguration
+            }
+        ]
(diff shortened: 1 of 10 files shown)


aws/templates/id/id_mobilenotifier.ftl
```diff
@@ -0,0 +1,116 @@
...
+                    "Link" : [ "Platform" ]
+                }
+            ]
+        },
+        MOBILENOTIFIER_PLATFORM_COMPONENT_TYPE : [
+            {
+                "Name" : "Engine"
+            },
+            {
+                "Name" : "SuccessSampleRate"
```
ml019: what's the logic for these being on the notifier and each platform? Presume it aligns with AWS ...

aws/templates/solution/solution_mobilenotifier.ftl
```diff
@@ -0,0 +1,206 @@
...
+        [#assign resources = occurrence.State.Resources]
+
+        [#assign successSampleRate = solution.SuccessSampleRate ]
+        [#assign encryptionScheme = solution.Credentials.EncryptionScheme?ensure_ends_with(":")]
+
+        [#assign deployedPlatformAppArns = []]
+
+        [#assign roleId = resources["role"].Id]
+
+        [#if deploymentSubsetRequired("iam", true) && isPartOfCurrentDeploymentUnit(roleId)]
```
ml019: Do we need to predefine the log groups like we did with lambda to be able to configure the monitors?

Is there any ability to control the log group name?

aws/templates/id/id_mobilenotifier.ftl
roleyfoley: Allows for a global mobilenotifier setting and overrides on the platforms. So if you have multiple platforms you can configure the sample rate individually

aws/templates/solution/solution_mobilenotifier.ftl
roleyfoley: Ahh yeah good point, no you can't change it. Problem I had was that I hadn't set a notification to find out what the log path was and can't find it documented. 

If we merge this PR in then I can add one after its been tested completely.

aws/templates/id/id_mobilenotifier.ftl
ml019: Normal override behaviour for settings should handle that though. So set on the component and it will be available for all subcomponents.

PR #411, opened 2018-08-03
support json input for cloudwatch events

kshychko:
an additional property for a schedule object to support json input

### DIFF
--- a/aws/templates/application/application_lambda.ftl
+++ b/aws/templates/application/application_lambda.ftl
@@ -183,6 +183,7 @@
                         targetId=fnId
                         enabled=schedule.Enabled
                         scheduleExpression=schedule.Expression
+                        input=schedule.Input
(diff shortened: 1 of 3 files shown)


ml019:
@kshychko did you need more than the existing path?

kshychko:
@ml019 the existing path is not enough when you need to specify json input for the schedule's rule event. New `Input` parameter allows providing complete json structure without limiting it to 
`{"path": "${path}"}`.
To send a request to a specific path in a lambda function more parameters are needed. I ended up with the the following input:
```
{
  "path": "/v1/healthcheck/",
  "httpMethod": "GET",
  "pathParameters": {},
  "queryStringParameters": {},
  "headers": {},
  "requestContext": {},
  "body": ""
}
```
to actually get `/v1/healthcheck/` requested

PR #414, opened 2018-08-07
Add System user support

roleyfoley:
Adds the ability to create users as a solution level component. 

Users have two types:
- system - generates an IAM Access/Secret key only
- user - generates and sets a Password only 

secrets and passwords are encrypted using the segment encryption key and stored in a pseudo stack

### DIFF
--- a/aws/templates/id/id_user.ftl
+++ b/aws/templates/id/id_user.ftl
@@ -1,4 +1,4 @@
-[#-- MOBILENOTIFIER --]
+[#-- USER --]
 
 [#-- Components --]
 [#assign USER_COMPONENT_TYPE = "user" ]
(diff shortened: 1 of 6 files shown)


aws/templates/id/id_user.ftl
```diff
@@ -0,0 +1,71 @@
+[#-- MOBILENOTIFIER --]
```
ml019: copy/paste error

aws/templates/solution/solution_user.ftl
```diff
@@ -0,0 +1,145 @@
...
+                    } + 
+                    attributeIfContent(
+                        "ManagedPolicyArns",
+                        context.ManagedPolicy
+                    ) + 
+                    attributeIfContent(
+                        "Policies",
+                        context.Policy
+                    )
+                    attributeIfContent("Variables", stageVariables)
```
ml019: This looks wrong - only reference to stageVariables?

PR #419, opened 2018-08-12
Use "fragment" in preference to "container"

ml019:
Expand the idea of a container fragment to apply to all components. We have been doing this progressively so time to get the terminology more generic.

So the idea of a "fragment" is a piece of template (typically just macro
invocations) that can be associated with any component to customise its
configuration, typically to add security information.

As some container fragments may explicitly use the containerId variable,
leave it defined for now. I've removed it everywhere I can.

A cmdb upgrade process also needed to be added to rename any existing
container_* files to fragment_* files.


roleyfoley:
Looks great! 
If we are expanding it across all application components would it be feasible to add it to solution components as well? Adding extra permissions on some solution level components, like ec2 could be handy.

ml019:
Yep - the idea of changing the terminology was to make it feel more natural/applicable anywhere. It then is a choice of the template designer for a level (segment, solution, application) to include support.

PR #420, opened 2018-08-12
Allow treating of runId as significant

ml019:
Include the runId as metadata in every template

Treate runId differences as not significant unless specifically
requested.

Update "config as file" support to include runId in the S3 object to
trigger code updates when the config file contents change (assuming the
request asks for them to be treated as significant).

Align lambda code with the config as file changes.

NOTE: the api-gateway template already uses the runId to trigger redeploys
so will automatically be affected by the same switch flagging runId
differences as significant.


ml019:
Yes - could add as part of core environment so option is there to force a restart. Reckon we add that now?

roleyfoley:
Yeah sounds good. I can't think of any reasons not to add it now

PR #435, opened 2018-08-15
Add SES SMTP Password Generation

roleyfoley:
The SES SMTP password is made from a hash of a users IAM SECRET Access key 

This PR adds this hashed smtp password into the attributes of a user application component.

This does not apply SES permissions just creates the password that is required


ml019:
Nice to not have to manage this manually given as you suggest, its a processed version of the standard access crdentials.

PR #442, opened 2018-08-16
Release enis fixes

kshychko:
fixed filter value and added wait for 5 seconds to get an ENI detached - not the best solution, probably needs to be replaced with a proper wait loop.

### DIFF
--- a/aws/utility.sh
+++ b/aws/utility.sh
@@ -1425,15 +1425,17 @@ function release_enis() {
     local requester_id="$1"; shift
     local eni_list_file="$( getTempFile eni_list_XXXX.json)"
     
-    aws --region "${region}" ec2 describe-network-interfaces --filters Name=requester-id,Values=*${requester_id} > "${eni_list_file}" || return $?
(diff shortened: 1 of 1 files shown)


ml019:
Probably better to put a loop around the sleep (limited number of iterations - 6 x 5s = 30s) and after each sleep, check to see if eni still attached.

This allows for the common case where it usually takes 5s, but accommodates any more extended processing delays

kshychko:
It will be easier to put the next step - delete ENI - into a loop, in this case - if ENI is in use, it will not be deleted. @ml019 do you have any objections on this?

ml019:
Sure. So shift sleep from detach to before delete, and put it in a loop. This then covers both a delay in detaching (delete will fail) and a delay in deleting.

roleyfoley:
You could use 
https://docs.aws.amazon.com/cli/latest/reference/ec2/wait/network-interface-available.html 

Which you can use to filter on the detached state 
``
attachment.status - The status of the attachment (attaching | attached | detaching | detached ).
``

kshychko:
@roleyfoley I didn't use a filter for attachment status - available status of an ENI is enough in our case

PR #443, opened 2018-08-16
remove underscore from config filename

kshychko:
Underscore symbol is escaped in the environment variable `SETTINGS_FILE` value, so a filename shall not contain it.

### DIFF
--- a/aws/templates/commonApplication.ftl
+++ b/aws/templates/commonApplication.ftl
@@ -342,7 +342,7 @@
                     context.DefaultCoreVariables || asFile
                 ) +
                 valueIfTrue(
-                    { "SETTINGS_FILE" : "config/config" + runId + ".json"},
+                    { "SETTINGS_FILE" : "config/config_" + runId + ".json"},
                     asFile,
                     valueIfTrue(
                         getSettingsAsEnvironment(occurrence.Configuration.Settings.Build) +


ml019:
There's a couple of spurious files in this PR - they need t be removed.

In regards to the underscore, not sure why it needs escaping? Most env vars have at least one underscore - that's what we convert the json hierarchy with.

aws/templates/application/application_lambda.ftl
```diff
@@ -295,7 +295,7 @@
                             getLocalFileScript(
                                 "configFiles",
                                 "$\{CONFIG}",
-                                "config_" + runId + ".json"
+                                "config" + runId + ".json"
```
roleyfoley: Would it work with a - instead of an _?

run.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
roleyfoley: I think these are accidental commits

tmp.json
```diff
@@ -0,0 +1,98 @@
+{
```
roleyfoley: I think these are accidental commits

aws/templates/application/application_lambda.ftl
kshychko: tried - didn't work

kshychko:
>In regards to the underscore, not sure why it needs escaping? Most env vars have at least one underscore - that's what we convert the json hierarchy with.

Underscore in the value is escaping somewhere in the code I believe. So in s3 we get `config/config_mk1g89kpgl.json` and in lambda environment - "SETTINGS_FILE"="config/configmk1g89kpgl.json"

kshychko:
@ml019 @roleyfoley  I force-pushed the branch - added a fix to a proper place, the underscore was simply missing in `SETTINGS_FILE` set

PR #455, opened 2018-08-19
Add resource policy support for api gateway

ml019:
Extend the existing fragment support to include policies.

Support IP address whitelisting via the resource policy to avoid the
need for CloudFront if all that is required is Ip whitelisting.

Separate out the creation of policy conditions to their own functions.


ml019:
@roleyfoley this is ready for review/merge. Now supports the same authentication models as ES

PR #457, opened 2018-08-21
ignore SEGMENT in DB_INSTANCE_IDENTIFIER if it is default

kshychko:
in RDS snapshot script

### DIFF
--- a/aws/snapshotRDSDatabase.sh
+++ b/aws/snapshotRDSDatabase.sh
@@ -97,7 +97,7 @@ status_file="$(getTopTempDir)/snapshot_rds_status.txt"
 # Ensure we are in the right place
 checkInSegmentDirectory
 
-DB_INSTANCE_IDENTIFIER="${PRODUCT}-${SEGMENT}-${TIER}-${COMPONENT}"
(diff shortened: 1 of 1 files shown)


aws/snapshotRDSDatabase.sh
```diff
@@ -97,7 +97,7 @@ status_file="$(getTopTempDir)/snapshot_rds_status.txt"
 # Ensure we are in the right place
 checkInSegmentDirectory
 
-DB_INSTANCE_IDENTIFIER="${PRODUCT}-${SEGMENT}-${TIER}-${COMPONENT}"
+DB_INSTANCE_IDENTIFIER="${PRODUCT}-${ENVIRONMENT}-${TIER}-${COMPONENT}"
```
roleyfoley: We might have add a bit of logic here. 
With the latest changes we actually now include SEGMENT and ENVIRONMENT in the name, but if the segment is "default" then its not included. 

Would you mind adding that in?

kshychko: @roleyfoley fixed

PR #464, opened 2018-08-26
Add fragment support for ecs solution

ml019:
For legacy containers that don't support task roles, allow permissions
to be added to the ecs instance service role.

The main target at this point is the rabbit container.

An alternative was to attach container policies to the ecs role rather than task roles, but this got messy in terms of ensuring everything was included into the iam subset and handling use of managed policies, so I just opted for adding the same policy via fragments at the solution level.

### DIFF
--- a/aws/createTemplate.sh
+++ b/aws/createTemplate.sh
@@ -208,7 +208,7 @@ function process_template() {
       ;;
 
     solution)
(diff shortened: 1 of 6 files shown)


ml019:
@roleyfoley your wish is my command....

PR #488, opened 2018-09-11
Add Support for User defined bootstrap scripts

roleyfoley:
Adds support for user defined bootstrap scripts. These can be used to perform additional hardening or configuration tasks on ec2/IaaS based components. 

It also introduces the idea of a scriptstore, which can be used to add extra locations for scripts that could be used by different components, currently it just applies to user boot strapping but could be used for infrastructure level lambda functions and containers.


aws/templates/solution/solution_ec2.ftl
```diff
@@ -101,6 +102,13 @@
             getInitConfigEnvFacts(environmentVariables, false) +
             getInitConfigDirsFiles(context.Files, context.Directories) ]
 
+        [#list bootstrapProfile.BootStraps as bootstrapName ]
+            [#assign bootstrap = bootstraps[bootstrapName]]
+            [@cfDebug listMode bootstrap true /]
```
ml019: Think this should be false

aws/templates/solution/solution_ecs.ftl
```diff
@@ -65,8 +68,18 @@
...
 
+        [#assign environmentVariables += getFinalEnvironment(occurrence, context).Environment ]
+
         [#assign configSets += 
+            getInitConfigEnvFacts(environmentVariables, false) +
             getInitConfigDirsFiles(context.Files, context.Directories) ]
+        
+        [#list bootstrapProfile.BootStraps as bootstrapName ]
+            [#assign bootstrap = bootstraps[bootstrapName]]
+            [@cfDebug listMode bootstrap true /]
```
ml019: Think this should be false

roleyfoley:
@ml019 debug statements removed and PR rebased

PR #490, opened 2018-09-13
LB Fixes and Improvements

roleyfoley:
This PR includes a number of changes and fixes for the LB component 

**Features** 
- Host Filtering - Adds a condition on a listener rule for the generated DNS hostname of the site. This prevents requests to the load balancer IP addresses which can be used for malicious probing. The option is applied on each port mapping and can only be used on ports with certificates configured 

**Changes**
- For Application Load balancers all listener rules are now created via CLI instead of cloudformation. This allows the rules to be managed easier and stops overalaps between cli and cloudformation which prevents updates 
- As part of this process all listener rules are removed before adding the new rules. This ensures that rules are cleaned up and only the configured rules will be applied. This will cause a brief (seconds) outage between the removal of the rules and the new ones being added 
- A default target group is created for all listeners and the default listener is always configured to use this rule 
- When a forwarding action is required for an application load balancer a separate target group is created along with a listener rule. This allows for conditions to be applied to all forwarding rules
- Network load balancers can only use a single target group per listener at the moment so applications will receive the default target group arn when linking to a network LB
- Targetgroup ARN look ups by linking components will receive the ARN via Attributes instead of a getReference lookup 

**Fixes** 
-  Align Security group name to the Id configuration which uses the parent details to create the property. This stops the security group from being recreated when a portmapping is renamed
- Move all listener rule creation to cli based config, this stops issues when a cli listener rule has been created and a cloudformation based run is added with the same priority
- Use the parent configuration details + priority for the listener rule Ids - if we move back to cloudformation this is required to ensure that rules are updated instead of trying to create a new rule with the same priority during updates 
- Add Health check configuration to all ports in master data 
- Update status code type for fixed response
- Cleanup extended link configuration that is no longer required for latest lb config 
- Make sure listener rule and target group config is only applied on application load balancers 
- Make sure cli commands error when they break 
- cleanup listener rules as part of deletion process

### DIFF
--- a/aws/templates/id/id_lb.ftl
+++ b/aws/templates/id/id_lb.ftl
@@ -61,6 +61,11 @@
                 "Default" : {}
             },
             {
+                "Name" : "HostFilter",
+                "Type" : BOOLEAN_TYPE,
+                "Default" : false
+            },
+            {
                 "Name" : "Mapping",
                 "Type" : STRING_TYPE
             },
@@ -145,8 +150,8 @@
                     },
                     {
                         "Name" : "StatusCode",
-                        "Type" : NUMBER_TYPE,
-                        "Default" : 404
+                        "Type" : STRING_TYPE,
+                        "Default" : "404"
                     }
                 ]
             },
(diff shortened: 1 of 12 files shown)


aws/utility.sh
```diff
@@ -861,21 +861,28 @@ function create_elbv2_rule() {
...
-function update_elbv2_rule() {
+function cleanup_elbv2_rules() { 
   local region="$1"; shift
-  local ruleid="$1"; shift
-  local configfile="$1"; shift
+  local listenerarn="$1"; shift
 
-  aws --region "${region}" elbv2 modify-rule --rule-arn "${ruleid}" --cli-input-json "file://${configfile}" || return $?
-}
+  pushTempDir "elbv2_listener_cleanup_XXXX"
```
ml019: There should be a matching popTempDir?

aws/templates/id/id_lb.ftl
```diff
@@ -223,8 +228,13 @@
 
     [#local sourcePort = (ports[portMappings[solution.Mapping!core.SubComponent.Name].Source])!{} ]
     [#local destinationPort = (ports[portMappings[solution.Mapping!core.SubComponent.Name].Destination])!{} ]
+
     [#local listenerId = formatResourceId(AWS_ALB_LISTENER_RESOURCE_TYPE, parentCore.Id, sourcePort) ]
 
+    [#local targetGroupId = formatResourceId(AWS_ALB_TARGET_GROUP_RESOURCE_TYPE, core.Id) ]
+    [#local defaultTargetGroupId = formatResourceId(AWS_ALB_TARGET_GROUP_RESOURCE_TYPE, "default", parentCore.Id, sourcePort ) ]
```
ml019: Shouldn't "default" be at the end rather than the beginning? In fact is "default" needed at all now?

roleyfoley:
@ml019 popTempDir added and also added error reporting for the listener rule delete

PR #503, opened 2018-09-25
Add finer grained control over environment serialisation

ml019:
Add the ability to select whether environment values should be serialised as strings or native JSON types
When everything is serialised as a string, add an optional prefix for objects and arrays so the consumer knows to process the contents as serialisable as JSON

### DIFF
--- a/aws/templates/setContext.ftl
+++ b/aws/templates/setContext.ftl
@@ -234,6 +234,31 @@
     ]
 ]
 
+[#assign settingsChildConfiguration = [
+        {
+            "Name" : "AsFile",
+            "Type" : BOOLEAN_TYPE,
+            "Default" : false
+        },
+        {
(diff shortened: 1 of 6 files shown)


ml019:
@roleyfoley ready for review

aws/templates/setContext.ftl
```diff
@@ -234,6 +234,31 @@
...
+            "Children" : [
+                {
+                    "Name" : "Escaped",
+                    "Type" : BOOLEAN_TYPE,
+                    "Default" : true
+                },
+                {
+                    "Name" : "Prefix",
+                    "Type" : STRING_TYPE,
+                    "Values" : ["json"],
```
roleyfoley: Should an empty value be a valid value? so that the Prefix isn't set?

PR #523, opened 2018-10-02
use CONFIG_FILE with s3 url value instead of SETTINGS_FILE

kshychko:
Set `CONFIG_FILE` with an s3 url value so it can be used in code instead of `OPSDATA_BUCKET`, `SETTINGS_PREFIX`, `SETTINGS_FILE` combination.


roleyfoley:
I've just added an update to make the full URL available in SETTINGS_FILE. This aligns with the codeontap naming

PR #527, opened 2018-10-04
Component reference and id properties

roleyfoley:
- Adds support for generate a markdown based reference document based on the component configuration defined in the *id.ftl files. 
- Adds a properties section to the component configuration objects which adds documentation or immutable attributes for a component 
Properties have the following types
    - Description - A description of the component and what it does  
    - Note - A informational note. Severity can also be set on a Note property to highlight the note appropriately 
    - Providers - The providers that support the component 
    - ComponentLevel - The level ( segment, solution, application, etc ) that the component is deployed with 
- Updated all Attribute name properties to **Names** instead of **Name**.  Single string is still accepted and will be converted to a single item list during processing
- Running createReference.sh requires the npm package remark-cli ( https://www.npmjs.com/package/remark-cli ) Which cleans up the markdown file and makes it happy for mkdocs to process 

To generate a reference you just need the GENERATION_DIR no cmdb is required 
```
${GENERATION_DIR}/createReference.sh -t component
``` 

The file will then be available in `${GENERATION_BASE_DIR}/dist/reference/component-reference.md`

An example of reference is available here 
https://codeontap.readthedocs.io/en/latest/reference/component-reference/ 

More documentation needs to be added but this provides the foundations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.iml
 dist/*
 build/*
+cache/*
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *~
 .idea/*
 *.iml
+dist/*
+build/*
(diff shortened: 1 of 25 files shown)


roleyfoley:
@ml019 ready for review when you are

PR #544, opened 2018-10-19
Move ssh component to bastion and make it an occurrence

roleyfoley:
This PR updates the SSH Segment component with the following features

- Converts ssh component to occurrence based
- Rename the component type to bastion, it now has an engine option so that in the future you could specify ssh, rdp or systemset manager ( at the moment just supporting ssh) 
- Adds a few of the standard ec2 level components that we have introduced across other ec2 based components 
   - Instance File logging to cloud watch 
   - BootStrap configuration 
- Adds support for Fragments, standard Permissions (all disabled by default) and links 

This would be used in cases where you might want to link the bastion component to an RDS instance to get the connection details. Then you would enable the decrypt permissions to get the RDS Password.


aws/templates/segment/segment_bastion.ftl
```diff
@@ -0,0 +1,242 @@
...
+                        arrayIfContent(
+                            [getPolicyDocument(_context.Policy, "fragment")],
+                            _context.Policy)
+                    managedArns=_context.ManagedPolicy
+                /]
+            [/#if]
+
+            [#if deploymentSubsetRequired("eip", true) &&
+                    isPartOfCurrentDeploymentUnit(bastionEIPId) &&
+                    sshStandalone]
```
ml019: sshStandalone was used when we were optionally using the nat servers as ssh hosts. Now that we are running bastion in a asg with default of 0, I think it would be cleaner to remove the option of combining the nat server and the ssh server. It is also easier operationally as there is only one ssh ip you have to worry about.

Probably cleaner to remove sshStandalone in a separate PR but if it would make this PR cleaner, please do so here.

roleyfoley:
@ml019 remove sshStandalone 

Also added a couple of other changes 
- cleaned up the autoscale config 
- moved multiAz config to standard setup for bastion, still 1 instance but it can be deployed to any availability zone 
- Fix for createPolicy on Autoscale 
- Disable Detailed monitoring for ssh scaling group

roleyfoley:
Added a few more things - System Manager Session Manager Support

- An Account level component - console - This enables session manager tracing logs to show the output of each session manager command in cloudwatch 
- An option in the Segment Object - ConsoleOnly - Which adds the iam policy to enable Session Manager to the bastion instance profile and disables the IPAdressGroup security group for ssh. 

This way you can either use the Console/Session Manager or use SSH via the IP Address Groups

aws/data/masterData.json
```diff
@@ -866,12 +866,13 @@
         },
         "ssh" : {
           "DeploymentUnits" : ["ssh"],
+          "MultiAZ" : true,
```
ml019: For SSH, I think we only ever want one bastion by default...

aws/templates/segment/segment_bastion.ftl
```diff
@@ -37,7 +37,12 @@
         [#assign storageProfile = getStorage(tier, component, BASTION_COMPONENT_TYPE)]
         [#assign logFileProfile = getLogFileProfile(tier, component, BASTION_COMPONENT_TYPE)]
         [#assign bootstrapProfile = getBootstrapProfile(tier, component, BASTION_COMPONENT_TYPE)]
-        [#assign processorProfile = getProcessor(tier, component, BASTION_COMPONENT_TYPE)]
+
+        [#assign processorProfile = (getProcessor(tier, component, "SSH")?is_string)?then(
```
ml019: Is this the right way around?

aws/data/masterData.json
```diff
@@ -864,10 +864,17 @@
             "dashboard"
           ]
         },
-        "ssh-ssh": {
-          "DeploymentUnits": [
-            "ssh"
-          ]
+        "ssh" : {
+          "DeploymentUnits" : ["ssh"],
+          "MultiAZ" : true,
```
ml019: Not sure we want multiAZ by default. What scenario are you envisaging...?

aws/templates/segment/segment_bastion.ftl
```diff
@@ -0,0 +1,249 @@
...
+            [#case "linux" ]
+                [#assign imageId = regionObject.AMIs.Centos.EC2]
+                [#break]
+        [/#switch]
+
+        [#assign storageProfile = getStorage(tier, component, BASTION_COMPONENT_TYPE)]
+        [#assign logFileProfile = getLogFileProfile(tier, component, BASTION_COMPONENT_TYPE)]
+        [#assign bootstrapProfile = getBootstrapProfile(tier, component, BASTION_COMPONENT_TYPE)]
+
+        [#assign processorProfile = (getProcessor(tier, component, "SSH")?is_string)?then(
```
ml019: Is this the correct way around?

PR #545, opened 2018-10-22
Multi-domain support

ml019:
Support possibility of multiple parents when getting ancestors

Add idea of "PRIMARY" domain - its the one we want in component
attributes, and "SECONDARY" domains - they are older domains we still
want to support.

Add basic Lambda@edge support for redirecting domains for SPAs.

Support multiple domains for API Gateway configs. Change bucket id
naming to be based on the domain so order doesn't affect the ids.

Remove dashboard support in sqs and api gateway in readiness of a more
complete implementation.

Add support for redirect on s3 websites.

Add an explicit child config for certificate objects.

Refactor the common pattern of checking for "Configured" and "Enabled"
children of a parent config object into the isPresent() function.

Provide finer grained control over the automatic addition of "Enabled"
attributes to parent config objects.


ml019:
@roleyfoley ready to review thanks

aws/templates/id/id_apigateway.ftl
```diff
@@ -1,5 +1,52 @@
...
+    . EDGE or REGIONAL
+    . signing based on any of the aliases
+    . API-KEY can be used for client metering
+
+If multiple domains are provided, the primary domain is used to provide the
+endpoint for the the API documentation and for the gateway attributes. For
+documentation, the others used to redirect to the primary.
+--]
+
+
```
roleyfoley: This would be handy in the Properties section as an array for the Description

aws/utility.sh
```diff
@@ -951,7 +951,16 @@ function deleteTreeFromBucket() {
   local optional_arguments=("$@")
 
   # Delete everything below the prefix
-  aws --region "${region}" s3 rm "${optional_arguments[@]}" --recursive "s3://${bucket}/${prefix}/"
+  aws --region "${region}" s3 rm "${optional_arguments[@]}" --recursive "s3://${bucket}/${prefix}${prefix:+/}"
```
roleyfoley: Will that put the prefix twice?

aws/templates/resource/resource_s3.ftl
```diff
@@ -88,12 +88,20 @@
...
             "IndexDocument" : index
         } +
-        attributeIfContent("ErrorDocument", error)
+        attributeIfContent("ErrorDocument", error) +
+        attributeIfContent(
+            "RedirectAllRequestsTo",
+            redirectTo,
+            {
+              "HostName" : redirectTo,
+              "Protocol" : redirectProtocol
```
roleyfoley: Might be worth making Protocol an optional configuration i.e. if_content add the protocol.

The default config is to redirect on the protocol that the request came in on which would be what we want for domain to new domain redirects sine you could either come via HTTP or HTTPS

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-websiteconfiguration-redirectallrequeststo.html

Its worded kind of weird though..

aws/utility.sh
ml019: No. It is using the alternate value operator to add a slash after the prefix if the prefix value is defined and not empty. The format has been used in other routines in this file.

aws/templates/id/id_apigateway.ftl
ml019: done

PR #548, opened 2018-10-23
Add Standard outputs to pseudo stack outputs

roleyfoley:
Add the standard stack outputs to pseudo stack files generated as part of the epilogue/prologue process. 

This ensures that when deploying the same environment to multiple accounts that the stack outputs align as they are supposed to be. 

Also adds some minor fixes for system users with console credentials


ml019:
Cleans things up a little as well as making stacks more consistent.

PR #549, opened 2018-10-23
replace Prefix with Paths variable to support an array of prefixes

kshychko:
and re-organise PublicAccess configuration per access details

### DIFF
--- a/aws/templates/id/id_s3.ftl
+++ b/aws/templates/id/id_s3.ftl
@@ -126,6 +126,7 @@
                 },
                 {
                     "Names" : "PublicAccess",
+                    "Subobjects" : true,
                     "Children" : [
                         {
                             "Names" : "Enabled",
@@ -144,9 +145,9 @@
                             "Default" : [ "_localnet" ]
                         },
                         {
-                            "Names" : "Prefix",
-                            "Type" : STRING_TYPE,
-                            "Default" : ""
+                            "Names" : "Paths",
+                            "Type" : ARRAY_OF_STRING_TYPE,
+                            "Default" : [ ]
                         }
                     ]
                 },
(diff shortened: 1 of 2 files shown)


aws/templates/id/id_s3.ftl
```diff
@@ -174,7 +175,13 @@
 
     [#local id = formatOccurrenceS3Id(occurrence)]
     [#local name = formatOccurrenceBucketName(occurrence) ]
-
+    [#local publicAccessEnabled = false ]
```
kshychko: introduce a local variable as we need to check configurations in a loop to find if any is enabled

aws/templates/solution/solution_s3.ftl
```diff
@@ -57,38 +53,46 @@
...
-                                                    "*",
-                                                    "*",
-                                                    publicIPWhiteList)]
-                    [#break]
-            [/#switch]
+        [#assign bucketPolicyId = resources["bucketpolicy"].Id ]
+        [#assign policyStatements = [] ]
+
+        [#list solution.PublicAccess?values as publicAccessConfiguration]
+            [#list asArray(publicAccessConfiguration.Paths) as publicPrefix]
```
kshychko: @ml019 I used `asArray` function to convert a single string value into an array, is it a correct approach?

kshychko: Also, this code produces a statement per path, but we could change it to a single statement with a Resource as an array instead of a single string. Shall we do it? It will require the changes in getS3Statement and getPolicyStatement functions.

ml019: Because of the type in the config, Paths will already be an array so no need to do asArray(). It won't hurt but is redundant.

https://github.com/codeontap/gen3/blob/master/aws/templates/common.ftl#L586
https://github.com/codeontap/gen3/blob/master/aws/templates/base.ftl#L328

For now I think a single statement per path is simpler/cleaner. If we find the need for a very long public access list then refactoring may be in order. I suspect there will typically not be a very large list.

PR #554, opened 2018-10-24
Ensure StackOutputs and SSH Keys reference their account and region

roleyfoley:
When deploying the same environment to a new account or to multiple regions we need to have separate encyrpted ssh keys for each account-region combination for the environment-segment. 

This PR adds this behaviour for the creation of new ssh keys and upgrades the existing CMDB update existing ssh keys 

Along with this, older stack output files didn't reference the account in the file name. This PR adds an upgrade step which renames all stack output files to include the account-region combination. 

It also checks that the minimum stack output values are available in the stack outputs file itself. and adds the default Account and Region to the files.


roleyfoley:
@ml019 Ready for Review

PR #563, opened 2018-10-25
ECS Service - Fix Port Mapping to Host and add daemon placement strategy

roleyfoley:
Fix: for Port Mapping configuration which allows for Host - Container mappings to be defined without a load balancer. 

Feature: ECS placement strategies - initially adds daemon support which ensures that all ECS instance have a single instance of the service running all the time

### DIFF
--- a/aws/templates/id/id_ecs.ftl
+++ b/aws/templates/id/id_ecs.ftl
@@ -247,7 +247,7 @@
                         {
                             "Names" : "Strategy",
                             "Type" : STRING_TYPE,
-                            "Values" : ["daemon"],
+                            "Values" : [ "", "daemon"],
                             "Description" : "How to place containers on the cluster",
                             "Default" : ""
                         }
--- a/aws/templates/id/id_ecs.ftl
+++ b/aws/templates/id/id_ecs.ftl
@@ -240,6 +240,18 @@
                     "Names" : "ContainerNetworkLinks",
                     "Type" : BOOLEAN_TYPE,
                     "Default" : false
+                },
+                {
+                    "Names" : "Placement",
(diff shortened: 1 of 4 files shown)


aws/templates/id/id_ecs.ftl
```diff
@@ -240,6 +240,18 @@
...
                     "Type" : BOOLEAN_TYPE,
                     "Default" : false
+                },
+                {
+                    "Names" : "Placement",
+                    "Children" : [
+                        {
+                            "Names" : "Strategy",
+                            "Type" : STRING_TYPE,
+                            "Values" : ["daemon"],
```
ml019: Need "" as one of the value if it is the default

aws/templates/resource/resource_ecs.ftl
```diff
@@ -159,7 +169,16 @@
                     }
                 },
                 networkMode == "awsvpc"
-            ) + 
+            ) +
+            valueIfTrue(
+                {
+                    "SchedulingStrategy" : "DAEMON"
+                },
+                (placement.Strategy?has_content) && (placement.Strategy == "daemon"),
```
ml019: Strategy has a default so should just be able to do straight comparison

aws/templates/id/id_ecs.ftl
roleyfoley: Fixed

aws/templates/resource/resource_ecs.ftl
roleyfoley: Fixed

roleyfoley:
@ml019 fixed up

PR #579, opened 2018-11-07
Lambda@Edge Support and CloudFront Redirection Policy

roleyfoley:
This PR adds Supports for Lambda@Edge Functions which can be attached to CloudFront Distributions 

A few changes: 
- Introduce a new tier global which has no network assigned and is intended for globally managed services
- Adds the ability to specify the region for a deployment unit. Under Product you can specify a deployment unit and set the region it should be deployed to 
```
  "Product": {
    "cfredirect" : {
      "Region" : "us-east-1"
    }
}
```
During processing of the template this will override the default product region and deploy the deployment unit to the specified region. This is required since Lambda@Edge Functions must be deployed to us-east-1
- Add configuration options to Lambda 
    - Versioned - adds a lambda version to the deployment. CloudFront only accepts versions of lambdas since it needs to replicate a fixed item across to all of the AWS regions 
    - DeploymentType - either REGIONAL or EDGE. In edge a few extra permissions are set for the lambda role and Log preprovisioning is disabled because EDGE lambdas log to the region they are invoked in which could be anywhere ( its the region closest to the client make the request to CloudFront) 
- Configuration options on SPA
    - eventHandlers - An extended link for EDGE lambdas only which include  an action property to specify where in the cloudfront processing it will be attached
- Added lambda Attributes fragment macro which can be used to override the lambda code location 
      - S3 Bucket and Prefix - get the code from a nominated bucket
      - zipFile - Pass the code to the function through the cloudformation template itself. This is available for minor functions as the code can only be a single file and 4096 bytes in size 
- Add cfredirect component which combines the other features into a predefined component. This is used to enable redirects from secondary domains to the primary domain on cloudfront distributions. 
- The cfredirect configuration is automatically added to SPA's if more than one domain alias is found.


roleyfoley:
@ml019  ready for review when you are. I've just added it to SPA for now but it would be easy to add to apigateway

PR #614, opened 2018-12-03
deploymentProfiles : Case sensitivity for Type and Deployment Profile Policies

roleyfoley:
types are treated as case insensitive when they are being processed. This change makes sure that deploymentProfiles follow the same convention.

This PR also adds support for using Deployment Profiles as enforced policy. Deployment profiles and Profile assignments can be configured on the tenant or account level object. When they are defined at this level they will override any deployment profiles assigned at lower levels ( product, environment, solution etc) 

Account level profiles override tenant level profiles which in turn override all other profiles. 

This approach allows for the implementation of tenant wide rules across all components. Tenant administrators can disable specific components, apply specific security policies or define a hibernate policy across all products and environments. 

This also adds a special deployment mode * which will apply the profile to all deploymentModes.


roleyfoley:
@ml019 ready when you are

PR #617, opened 2018-12-06
WIP: add support for global alert config and refactor logmetric

roleyfoley:
This includes two refactors which give us a flexible approach to adding any alarm to any resource.

- Alerts are now resource based and have dimension and namespcace look ups depending on the resource type. This ensures that alerts are filtered to only alert based on the component that they have been added to 

- LogMetrics have been re-factored to support the Alert changes above and adds them as proper resources in the component. 

When configuring an alert you can qualify the alert configuration in 3 different approaches 
- Resource Name - only applies the alert to a given resource 
- Resource Type - applies the alert to all resources in the component which match the specified type 
- Monitored Resources - applies to all resources with an attribute of Monitored set to true ( default )

### DIFF
--- a/aws/templates/application/application_ecs.ftl
+++ b/aws/templates/application/application_ecs.ftl
@@ -296,15 +296,16 @@
                         
                         [#list solution.LogMetrics as logMetricName,logMetric ]
 
+                            [#assign logMetricResource = resources[("lgMetric" + logMetricName)] ]
                             [#assign logFilter = logFilters[logMetric.LogFilter].Pattern ]
 
                             [@createLogMetric
                                 mode=listMode
-                                id=formatDependentLogMetricId(lgId, logMetric.Id)
-                                name=formatName(logMetricName, core.Name)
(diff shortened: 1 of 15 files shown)


roleyfoley:
Seems to be working as expected. An example of enabling an alert for an SQS dlq getting any messages would be 

```
"queue" : {
    "DeploymentUnits" : ["queue"],
    "sqs" : {
        "DeadLetterQueue" : {
        },
        "Alerts" : {
            "DLQMessagesReceived" : {
                "Description" : "The DLQ is receiving messages",
                "Name" : "DLQMessagesReceived",
                "Metric" :  "ApproximateNumberOfMessagesVisible",
                "Threshold" : 10,
                "Severity" : "ERROR",
                "Resource" :{
                    "Name" : "dlq"
                }
            }
        }
    }
}
```

roleyfoley:
This change is a breaking change though. Existing log metric alerts will need to be updated based on the new format of the alert configuration

PR #621, opened 2018-12-13
Extend alerting and Log Metric support

roleyfoley:
This PR adds support for Metric based alerting and Log Metrics for the following Components

- APIGateway - Metric Alerts and Log Metrics 
- Mobile Notifier - Metric Alerts and Log Metrics 
- Lambda - Metric Alerts and Log Metrics 
- ECS Hosts - Metric Alerts and Log Metrics 
    - ECS Services - Metric Alerts and Log Metrics 
    - ECS Tasks - Log Metrics  ( Metrics not available for Tasks ) 
       - ECS Containers - Log Metric ( When container based logs are enabled ) 

To support this all of the resources metrics and dimensions have been mapped to their appropriate values and outputs have been added for ECS hosts and ECS Tasks 

My brain hurts now...

PR #623, opened 2018-12-14
Add support for time based scheduled ECS tasks

roleyfoley:
Depends on #621 for ECS output mappings

ECS tasks can be triggered through CloudWatch events which provide a cron like schedule for running tasks.

We currently use this on lambda functions and this PR extends that functionality for ECS tasks. This can be used for long running tasks or using code already in a container for a scheduled task.

The command/entrypoint can not be overridden when it is invoked so the task must be able to run using the command/entry point the task is deployed with ( They can be overriden when defining the task using fragment files )

### DIFF
--- a/aws/templates/application/application_ecs.ftl
+++ b/aws/templates/application/application_ecs.ftl
@@ -234,6 +234,48 @@
                 [/#if]
             [/#if]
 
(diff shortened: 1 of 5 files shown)


ml019:
Neat! Only thing Jenkins gives you is locking to ensure only one are running at the same time. But for things like backups this is perfect.

PR #654, opened 2019-01-29
Add run lambda script

roleyfoley:
Adds support for invoking a lambda function from a Jenkins job

This uses the build blueprint process to locate the lambda ARN, and only requires the deployment unit to run the task. 

Payload is optional depending on how your function behaves and the last 4kb of the log can be provided optionally if required

### DIFF
--- a/aws/runLambda.sh
+++ b/aws/runLambda.sh
@@ -1,0 +1,151 @@
+#!/usr/bin/env bash
+
+[[ -n "${GENERATION_DEBUG}" ]] && set ${GENERATION_DEBUG}
+trap '. ${GENERATION_DIR}/cleanupContext.sh; exit ${RESULT:-1}' EXIT SIGHUP SIGINT SIGTERM
+. "${GENERATION_DIR}/common.sh"
+
+
+#Defaults 
+INCLUDE_LOG_TAIL_DEFAULT="false"
+
+tmpdir="$(getTempDir "cote_inf_XXX")"
+
+function usage() {
+    cat <<EOF
+
+Run an AWS Lambda Function
+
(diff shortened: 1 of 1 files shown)


roleyfoley:
@kshychko Would you mind having a look at this and see what you think?

aws/runLambda.sh
```diff
@@ -0,0 +1,151 @@
...
+
+    if [[ -n "${INPUT_PAYLOAD}" ]]; then 
+        INPUT_PAYLOAD="$( echo "${INPUT_PAYLOAD}" | jq -r '.' )"  
+        if [[ -z "${INPUT_PAYLOAD}" ]]; then 
+            fatal "Invalid input payload - must be in JSON format"
+            return 255
+        fi
+    fi 
+
+    LAMBDA_ARN="$( jq -r '.Occurrence.Occurrences[0].State.Attributes.ARN' < "${BUILD_BLUEPRINT}" )"
```
ml019: There can be more than one lambda function in a single lambda zip and hence multiple functions for one deployment unit. Is there a need for a function name parameter to delineate if > 1 function? Or am I missing something abut the way the blueprint is generated?

roleyfoley: This run command uses the occurrence to lookup the function to run rather than the component. So for any give runLambda there would only be one function.

ml019: All I can change is the deployment unit, so if I want to run the second lambda function defined for a deployment unit with three lambda functions, I can't? Use of the occurrence would just give me the first?

I'm missing something here...

roleyfoley: Nope, I am. Didn't realise you could have multiple lambdas under the same deployment unit. Thought that the deployment unit was defined at the function level rather than the lambda level.

roleyfoley: Fixed up with functionId parameter also added https://github.com/codeontap/gen3-automation/pull/169 to align the automation code

PR #657, opened 2019-02-04
Remove agent startupdelay

roleyfoley:
Removes a startup delay for dynamic agents. In jenkins this is a default behaviour for on demand agents which are persistent. Jenkins will wait a little bit to see if one of the other agents finishes a job and then starts the process of requesting a new instance. 

For container based jobs which only perform one task this is not a desirable feature and so it has been disabled as part of our container deployment.


roleyfoley:
Also adds a couple of extras

- Renames image from slave to agent ( inline with open source community best practices) 
- Makes the image name configurable through settings in order to use the extended product specific build agents

Issue #101, opened 2018-02-20
Add Support for SSM Patch Management


roleyfoley:
AWS System Set Manager can be used to automate tasks on Ec2 based or server instances within an environment. 

The key task that we want to support is for automated patch management

roleyfoley:
Key components to add: 

- Patch Baseline creation
- Tag based maintenance windows - Using the Codeontap default tags 
- Automated patching during maintenance window 
- Safe reboot support ( e.g. don't reboot all nodes in an ECS cluster together)

PR #703, opened 2019-03-22
<feat> Add automated snapshots with SSM Automation

roleyfoley:
Adds the ability to automatically backup datavolumes through AWS SSM Automation Documents 

AWS SSM Automation allows you to run Documents which can be used to perform tasks within an AWS environment. They can make calls to the AWS API, create cloudformation templates, invoke lambdas and run commands on Ec2 instances 

Combined with maintenance windows you can schedule a collection of tasks to be run against a collection of ec2 instances based on tags, or just run them against non ec2 resources

For data volumes we use two documents provided by AWS, 
  - AWS-CreateSnapshot - This makes a call directly to the Ec2 Api to create the snapshot
  - AWS-DeleteEbsVolumeSnapshots - Creates a lambda using cfn which lists the current snapshots and removes any older than the retention period

This method also lays the ground work for #101 which uses maintenance windows and tasks

### DIFF
--- a/aws/templates/id/id_datavolume.ftl
+++ b/aws/templates/id/id_datavolume.ftl
@@ -51,6 +51,35 @@
                 {
                     "Names" : "Profiles",
                     "Children" : profileChildConfiguration
+                },
+                {
+                    "Names" : "Backup",
+                    "Children" : [
+                        {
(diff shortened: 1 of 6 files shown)


roleyfoley:
@ml019 could you please review

PR #706, opened 2019-03-27
<feat>Refactor Segment S3 and S3 object notifications

roleyfoley:
This PR has two related changes 

**Re-factoring segment S3 into the baseline component** 
This moves the current S3 deploymentSubset into the baseline component. The S3 buckets are  defined as subcomponents of the baseline parent as the `BASELINE_DATA_COMPONENT_TYPE` which allows for them to be linked and treated as semi independent components from the baseline. 

You can now have multiple buckets under different roles, with two required as a minimum, `application` and `data` which are the standard appdata and ops buckets that we create now 

Backwards compatibility is supported and to do this the "s3" component type has been removed from segment and is now a deploymentSubset like iam and lg. This allows you to either run the `baseline` deployment unit and a `s3` unit to keep templates happy 

**Add S3 Notifications to SQS queue** 
Adds the ability to link an SQS queue to notifications on the S3 bucket. When this is applied S3 will add event notifications to the linked SQS queue(s) based on the filter applied to the Notification 

Example
```
                "stage" : {
                    "DeploymentUnits" : ["stage"],
                    "S3" : {
                        "Notifications" : {
                            "createLbLogs" : {
                                "Links" : {
                                    "queue" : {
                                        "Tier" : "app",
                                        "Component" : "queue"
                                    }
                                },
                                "Prefix" : "logs/",
                                "Actions" : [ "create" ]
                            }
                        }
                    }
                },
```

In this example objects created under the `logs/` prefix will have a notification event added to the SQS queue called `queue` the event has the s3 path so that something processing events on the queue can access the updated data.

This has been enabled for solution and segment level s3 buckets


roleyfoley:
@ml019 could you please review

PR #722, opened 2019-04-18
<feat> Config Stores 

roleyfoley:
This PR adds a new component called a configstore 

The idea of a configstore is to provide a storage location for multiple instances of a configuration value that can be queried dynamically through code. 

The config store has a collection of branch subcomponents which each hold a set of settings. Ideally the keys of the settings are the same across each branch and then the value changes. This doesn't have to be the case though 

From a deployment perspective the configstore is a Dynamo DB table with each branch as an item in the table. 

Item attributes are driven through our existing settings generation process. By default all standard settings are disabled except for links and settings are added through a fragment file. This ensures consistency with the setting names and attributes 

The table is read only for all components which link to it with the exception of the States which can be written to for components that link with the `managestate` role. An state manager can modify these to inform other components of changes to a state. States are configurable for each branch independently. 

In the case of blue/green switching you would have config branch for the blue config, one for the green config and a state branch which would manage the state of the environment in a central branch 

Exmaple definition - This would create a blue/green config store with a branch for storing state
```
                "dbswitch" : {
                    "configstore" : {
                        "Instances" : {
                            "default" : {
                                "DeploymentUnits" : ["walksdb-switch"]
                            }
                        },
                        "Fragment" : "_dbswitch",
                        "Branches" : {
                            "state" : {
                                "States" : {
                                    "activeEnvironment" : {
                                        "InitialValue" : "blue"
                                    }
                                }
                            },
                            "blue" : {
                                "Links" : {
                                    "DB" : {
                                        "Tier" : "db",
                                        "Component" : "walks-rds",
                                        "Instance" : "blue"
                                    }
                                }
                            },
                            "green" : {
                                "Links" : {
                                    "DB" : {
                                        "Tier" : "db",
                                        "Component" : "walks-rds",
                                        "Instance" : "green"
                                    }
                                }
                            }
                        }
                    }
```

Extra settings can the defined in a fragment file 
```
[#case "_dbswitch" ]

    [@Settings { "ENVIRONMENT" : _context.DefaultEnvironment.ENVIRONMENT } /]

    [#if _context.Branch == "green" ]
        [@Settings { "DATABASE_NAME" : "abc123" }/]
    [/#if]

    [#if _context.Branch == "blue" ]
        [@Settings { "DATABASE_NAME" : "def456" }/]
    [/#if]
    [#break]
``` 

The _context.Branch variable is used in the fragment to set the value for a given toggle.


ml019:
Looks good. Most of the code was getting the dynamodb support in place which I suspect we will reuse later.

As discussed, suggest adding support for the option of the state being held in a well known item e.g. "_state" for the common case of a switch with only one active position at a time. This makes changing the position only require update to a single item.

Suggest adding an "InitialPosition" attribute which if defined, results in an extra item being created with the value provided, which should be validated against the defined positions. Changing this value on an existing switch will have no effect as it will already have a state.

roleyfoley:
@ml019 was thinking about it and ended up renaming the components to `configstore` which has branches 

At the moment this component updates the configuration in a DynamoDb table but this same component could also provide the settings to other components as link Attributes which would allow for the creation of a component which manages configuration across a collection of different components. 

Also added the initial state value for states and updated the DynamoDb policy to work with table names or arns.

PR #723, opened 2019-04-23
<feat> RDS Template policy processing management

roleyfoley:
Adds the ability to define the deletion and Updatereplace policies on RDS instances. These policies define what to do when cloudformation deletes a resource or the resource is replaced as part of an update 

In RDS this defaults to `Snapshot` which will create a snapshot of the RDS instance before removing it. In some cases this is required, for example if you have a database which easy to recreate then you can disable this functionality

Also disables the snapshotting during RDS instance replacement template processing

### DIFF
--- a/aws/templates/resource/start.ftl
+++ b/aws/templates/resource/start.ftl
@@ -395,6 +395,7 @@
             dependencies=[]
             metadata={}
             deletionPolicy=""
+            updateReplacePolicy=""
             updatePolicy={}
             creationPolicy={}]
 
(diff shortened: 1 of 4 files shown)


aws/templates/resource/start.ftl
```diff
@@ -395,6 +395,7 @@
             dependencies=[]
             metadata={}
             deletionPolicy=""
```
ml019: Was there already some partial support for Deletion Policy?

PR #731, opened 2019-04-26
<fix>VPC security port lookup for ranges

roleyfoley:
Adds support for port ranges in ingress security groups
Adds default policy to pipeline ec2 instances

### DIFF
--- a/aws/templates/application/application_datapipeline.ftl
+++ b/aws/templates/application/application_datapipeline.ftl
@@ -83,6 +83,7 @@
                 "DefaultEnvironment" : defaultEnvironment(occurrence, contextLinks),
                 "Environment" : {},
                 "Links" : contextLinks,
+                "Policy" : standardPolicies(occurrence),
                 "DefaultCoreVariables" : true,
                 "DefaultEnvironmentVariables" : true,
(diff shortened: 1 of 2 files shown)


ml019:
Wonder if port ranges should be a separate object rather than a variant of a port. Lookup up could then use fallback to treat the port as an individual port and if not defined then look for a range.

roleyfoley:
> Wonder if port ranges should be a separate object rather than a variant of a port. Lookup up could then use fallback to treat the port as an individual port and if not defined then look for a range.

Woudln't you then need to ensure that you don't have overlapping names between the port objects and the portrange objects to make sure fallback works?

ml019:
You would - but when would you want to name a range the same as a port? Guess its whether they are really are the same or different concept.

Maybe we should be treating a single port as a port range of 1 - so change ports collection to a portranges collection...

roleyfoley:
Yeah that could work. What this has is backwards compatability for single ports or port ranges. 
Ports are defined in config quite a bit so it would be a breaking change

ml019:
Hmm. OK - I see why you went the way you did. Let's go with this as an interim solution. Can always add distinct portrange object later.

PR #732, opened 2019-04-29
<refactor> Regional WAF Support

ml019:
Our approach to WAF to date has been somewhat simplistic in that we only
supported IP filtering, and each IPAddressGroup was its own condition
and rule. Rules were also defined per group once at the account level.

With the introduction of support for the full set of conditions, and the
desire to mix IP filtering with OWASP support, account level WAF support
will be deprecated in favour of environment specific configuration.

The acl + rule cost model of WAF means that this change will likely be
cost neutral, as each acl will only have one rule for IP filtering,
rather creating a rule for every IP address group even if it wasn't
used.

The support for accounts has been left in place for now, but can be
removed once all uses have been redeployed.

Support has also been added for global and regional variants of WAF in
order to support direct WAF usage on regional API gateways and load
balancers.

WAF has also introduced less restrictive rules on supported CIDR masks.
These have been incorporated into this change.

Improvements to the documentation of CIDR processing routines
and the API Gteway configuration management have also been added.

The WAF section of the API Gateway support has also been pulled out of the
CloudFront section and made standalone in readiness for WAF directly
configured on an API Gateway.

Support has also been added to create rules and/or conditions as part of creating an acl. 

This commit updates the underlying WAF support. A separate PR will add
specific support on API gateway and load balance endpoints, as well as
default OWASP configurations.


aws/templates/resource/resource_waf.ftl
```diff
@@ -69,13 +236,37 @@
     /]
 [/#macro]
 
-[#macro createWAFAcl mode id name metric default rules]
+[#-- Rules should be ordered in decreasing order of importance --]
+[#macro createWAFAcl mode id name metric default rules=[] regional=false]
```
roleyfoley: Should we have priorities here?

aws/templates/setContext.ftl
```diff
@@ -253,6 +253,11 @@
             "Type" : ARRAY_OF_STRING_TYPE,
             "Mandatory" : true
         },
+        {
+            "Names" : "OWASP",
```
roleyfoley: Instead of hardcoding a security service maybe we should have security profiles which match to security standards. Then the rules defined in them can be seen easily in master data files and changed if requried

aws/templates/resource/resource_waf.ftl
ml019: Priorities set to reflect rules array ordering.  Each to add a Priority attribute to the rule object, but then all of the rules would need one to avoid ambiguity.

aws/templates/setContext.ftl
ml019: Yep - plan  to change this in next PR. For now this attribute isn't checked so its harmless.

aws/templates/resource/resource_waf.ftl
ml019: Happy to add if you have a use case in mind. Was waiting for one before adding the extra complexity. Might find it is important if mixing predefined rules with extras. We'd need to control priority ranges then.

ml019: Actually scratch that since use of rule is when priority added.

roleyfoley: Only a behaviour thing really. If i was adding a new rule my general thinking would be to add it to the end. which would make it the last in the list 

Since the priority here isn't going through to AWS just ordering the list based on priority would allow for overlaps but give you reasonable control over the rule order. 

ec2 init scripts have a similar function at the moment 
https://github.com/codeontap/gen3/blob/master/aws/templates/resource/resource_ec2.ftl#L39 

the idea here was that I can set priorities in groups to say, this rule must be processed befoer another one, but I'm not fussed of the order with scripts in the same prirotiy

ml019: So your idea is more of a "band" thn a specific priority

PR #749, opened 2019-05-09
<feat> SPA Path based routing to LB

roleyfoley:
Adds support for custom routing paths on SPAs. This allows specifc paths to be routed to a other components - currently LB ports are supported 

Example
```
                "www" : {
                    "spa" : {
                        "Instances" : {
                            "default" : {
                                "Versions" : {
                                    "v1" : {
                                        "DeploymentUnits" : ["www-v1"]
                                    }
                                }
                            }
                        },
                        "CloudFront" : {
                            "Paths" : {
                                "api" : {
                                    "PathPattern" : "/api/*",
                                    "Link" : {
                                        "Tier" : "elb",
                                        "Component" : "api-lb",
                                        "Port" : "httpsapi"
                                    }
                                }
                            }
                        }
                    }
                }
``` 

This example would route requests for `/api/*` to the api-lb httpsapi public endpoint

### DIFF
--- a/aws/templates/id/id_spa.ftl
+++ b/aws/templates/id/id_spa.ftl
@@ -139,6 +139,48 @@
                                     "Mandatory" : true
                                 }
                             ]
+                        },
+                        {
+                            "Names" : "Paths",
+                            "Subobjects" : true,
+                            "Description" : "Additional path based routes to other components",
+                            "Children" : [
+                                {
+                                    "Names" : "PathPattern",
+                                    "Type" : STRING_TYPE,
+                                    "Mandatory" : true
+                                },
+                                {
+                                    "Names" : "Link",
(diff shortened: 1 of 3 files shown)


aws/templates/id/id_spa.ftl
```diff
@@ -195,7 +249,8 @@
                     "Id" : formatDependentWAFAclId(cfId),
                     "Name" : formatComponentWAFAclName(core.Tier, core.Component, occurrence),
                     "Type" : AWS_WAF_ACL_RESOURCE_TYPE
-                }
+                },
+                "paths" : pathResources
```
ml019: needs to be conditional so component deployment correctly detected

roleyfoley: fixed

PR #762, opened 2019-05-12
<refactor> GEN3 directory tree to separate state

ml019:
Add support for semver range checking based on the syntax used by
node-semver.

Add the ability to constrain a cmdb upgrade to a particular version of
the GEN3 framework, mainly to allow a cmdb upgrade to be added to the
framework but not applied until the corresponding GEN3 upgrade is
active.

Add the upgrade to v2.0.0 of the cmdb, which separates state from
config/infrastructure. It has not yet been enabled to ensure the
refactoring still works with existing cmdb layouts.

Add functions to support switch to use of branches for releases.

Functions for context tree access now support the pre and post v2.0.0
cmdb directory formats. A few extra globals were added for locating
builds and state to assist with the transition between directory formats.

Note that once the v2.0.0 cmdb update is applied, build jobs will need
to be changed to reflect the new location, so they should use
SEGMENT_BUILDS_DIR when determining where in the repo to save build
references. 

There is an equivalent issue for gen3-automation that should
be applied with the issue to support the change of directory tree for
builds. See https://github.com/codeontap/gen3-automation/pull/176

The v1.1.0 cleanup did not correctly clean up accounts repos so a v1.1.1
cleanup has been added that can handle multiple config/infrastructure
pairs in a single repo. This should only affect accounts repos.

Add extra functions from gen3-automation/utility.sh so the two align.

### DIFF
--- a/aws/contextTree.sh
+++ b/aws/contextTree.sh
@@ -81,19 +81,19 @@ function findGen3AccountOperationsDir() {
   local root_dir="$1"; shift
   local account="$1"; shift
 
-  local infrastructure_dir="$(findGen3AccountInfrastructureDir "${root_dir}" "${account}")"
-  if [[ -n "${infrastructure_dir}" ]]; then
-    echo -n "${infrastructure_dir}/operations"
-    return 0
-  fi
-  return 1
+  # TODO(mfl): Remove infrastructure checks when all repos converted to >=v2.0.0
+  findDir "${root_dir}" \
+    "operations/**/${account}/settings" \
+    "${account}/operations/settings" \
+    "infrastructure/**/${account}/operations" \
+    "${account}/infrastructure/operations"
 }
 
-# TODO(mfl): Remove infrastructure checks when all repos converted to >=v2.0.0
 function findGen3AccountStateDir() {
   local root_dir="$1"; shift
   local account="$1"; shift
 
+  # TODO(mfl): Remove infrastructure checks when all repos converted to >=v2.0.0
   findDir "${root_dir}" \
     "state/**/${account}" \
     "${account}/state" \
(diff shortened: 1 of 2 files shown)


roleyfoley:
Should builds be part of the state instead of infrastructure? 
A build is really the codes current state in a given environment the same way the cloudformation is the current infrastructure state in a given environment

ml019:
@roleyfoley 

Builds is part of the config for a particular environment. They exist even if the environment doesn't so it can be created from scratch.

I've updated the v2.0.0 migration to move /infrastructure/operations -> /operations/settings the same as config.

Please review again.

aws/contextTree.sh
```diff
@@ -4,688 +4,832 @@
...
-  [[ ! ("segment" =~ ${location}) ]] && fatalDirectory "segment"
+  getJSONValue "${COMPOSITE_BLUEPRINT}" "${patterns[@]}"
 }
 
-function fatalProductOrSegmentDirectory() {
-  fatalDirectory "product or segment"
+# -- Buckets --
+
+function getOperationsBucket() {
+  getBucketName "s3XsegmentXops"
```
roleyfoley: This will need to account for the new bucket ids in the baseline data storage

roleyfoley:
@ml019 Is this still required as part of the latest round of changes? Its for a separate part of the process but just want to make sure

ml019:
@roleyfoley this is still needed. It rearranges the repo structure but doesn't affect dynamic loading etc.

aws/contextTree.sh
ml019: NOt sure if I actually changed this - maybe just missed with the baseline changes?

ml019:
Closing as most of functionality was included under https://github.com/codeontap/gen3/pull/1107.

Remaining changes associated with transition to branches will be done under a separate PR.

Switch to v2.0.0 as the default will also be done as a separate PR.

PR #778, opened 2019-05-21
<feature> Local build commit is preferred over shared

roleyfoley:
Allows for a shared deployment unit to have its own git commit reference which overrides the shared commit. 

This allows for different versions of a deployment unit while still using the same registry

### DIFF
--- a/aws/templates/common.ftl
+++ b/aws/templates/common.ftl
@@ -1126,6 +1126,11 @@ behaviour.
             alternatives
         ) ]
 
+    [#-- a local unit build commit takes preference over a shared one --]
+    [#if occurrenceBuild.COMMIT?has_content]
+        [#local occurrenceBuildCommit = occurrenceBuild.COMMIT ]
+    [/#if]
+
     [#-- Reference could be a deployment unit or a component --]
     [#if occurrenceBuild.REFERENCE?has_content]
         [#-- Support cross-segment references --]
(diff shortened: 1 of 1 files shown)


aws/templates/common.ftl
```diff
@@ -1126,11 +1126,17 @@ behaviour.
             alternatives
         ) ]
 
-    [#-- a local unit build commit takes preference over a shared one --]
+    [#-- a local unit build commit/tag takes preference over a shared one --]
+    [#local occurrenceBuildCommit = "" ]
     [#if occurrenceBuild.COMMIT?has_content]
         [#local occurrenceBuildCommit = occurrenceBuild.COMMIT ]
```
ml019: just add !"" to the end of the #local and skip the initialisation and the if

aws/templates/common.ftl
```diff
@@ -1126,11 +1126,17 @@ behaviour.
...
     [/#if]
 
+    [#local occurrenceBuildTag = ""]
+    [#if occurrenceBuild.TAG?has_content ]
+        [#local occurrenceBuildTag = occurrenceBuild.TAG ]
```
ml019: same as above

aws/templates/common.ftl
```diff
@@ -1151,9 +1162,23 @@ behaviour.
     [/#if]
 
     [#return
-        attributeIfContent(
+        attributeIfTrue(
```
ml019: could have stuck with attributeIfContent and dropped the ?has_content checks because attributeIfContent does that anyway.

roleyfoley:
@ml019  fixed up

aws/templates/common.ftl
roleyfoley: Can I check two attributes for content though?

PR #779, opened 2019-05-21
<fix> deployment profiles with instance version

roleyfoley:
* Fixes deployment profiles so that they can be applied at the instance and version level of a component and still get processed 
   They used to only be applied at the typeObject level but are now applied as part of the solution configuration generation process
* Adds Enable/Disable to links to allow for Deployment profile modes which might not want a specific link to apply

### DIFF
--- a/aws/templates/common.ftl
+++ b/aws/templates/common.ftl
@@ -1400,10 +1400,6 @@ behaviour.
         [#local typeObject = component ]
     [/#if]
 
-    [#--  Deployment Profile Configuration Overrides --]
-    [#local deploymentProfileComponents = getDeploymentProfile(typeObject, deploymentMode) ]
-    [#local typeObject = mergeObjects(typeObject, deploymentProfileComponents[type]!{} )]
-
     [#if tier?has_content]
         [#local tierId = getTierId(tier) ]
         [#local tierName = getTierName(tier) ]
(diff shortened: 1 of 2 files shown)


aws/templates/common.ftl
```diff
@@ -1669,140 +1682,141 @@ behaviour.
...
-                        "Name" : link.Tier
-                    },
-                    "Component" : {
-                        "Id" : link.Component,
-                        "Name" : link.Component
-                    },
-                    "Instance" : {
-                        "Id" : "",
-                        "Name" : ""
+    [#if (link.Enabled)!true ]
```
ml019: Rather than have most of the code in a massive if, suggest detecting disabled and return {}  immediately.

aws/templates/common.ftl
```diff
@@ -1943,12 +1957,12 @@ behaviour.
...
 [/#function]
 
-[#function getDeploymentProfile typeObject deploymentMode ]
+[#function getDeploymentProfile occurrenceProfiles deploymentMode ]
 
     [#local deploymentProfileNames = []]
 
-    [#if (typeObject.Profiles.Deployment)?? ]
-        [#list asArray(typeObject.Profiles.Deployment) as profileName ]
+    [#if (occurrenceProfiles)?? ]
```
ml019: Compiler won't let you call getDeploymenrProfile without occurrenceProfiles.

I think the intent is it must be there so just drop the if test.

Also the other asArrays could be modified with a missing default of ![] to remove the if test - I think it looks cleaner

aws/templates/setContext.ftl
```diff
@@ -115,6 +115,11 @@
             {
                 "Names" : "Type",
                 "Type" : STRING_TYPE
+            },
+            {
+                "Names" : "Enabled",
```
ml019: good idea

aws/templates/common.ftl
```diff
@@ -1499,12 +1495,29 @@ behaviour.
...
+                    ]
+
+                    [#local occurrenceProfiles = [] ]
+                    [#list occurrenceContexts as occurrenceContext ]
+                        [#if ((occurrenceContext.Profiles.Deployment)![])?has_content ]
+                            [#local occurrenceProfiles = occurrenceContext.Profiles.Deployment ]
+                        [/#if]
+                    [/#list]
+
+                    [#if occurrenceProfiles?has_content ]
```
ml019: I think you don't want this test as it will then not pick up profiles set at other levels like product or tenant.

I think you could just skip the if test and pass in the potentially empty array.

Effectively we want deployment profiles to trump everything.

aws/templates/common.ftl
roleyfoley: fixed

PR #785, opened 2019-05-26
WIP: <refactor> Dynamic template loading

ml019:
This is the first iteration of changes to support the dynamic loading of
templates, as part of the transition to supporting multiple providers.
The main aim was to determine the language constructs that would make this
possible.

The key innovations in this commit relate to dynamic loading of templates
and indirect macro invocation, both of which are to be found in the
start.ftl template in the resources directory. The core template loading is in base.ftl.

All segment, solution and application templates have been converted to
standalone ftl templates which define a single macro corresponding to the
level/component combination. Subcomponent processing is still done as
part of the parent component processing.

Because of variable scoping, all these macros currently accept the tier
and component, but in a next iteration, the logic to determining which
components are required can move out of each of the templates and the
macros can instead accept an occurrence instance. This will require
tweaking of a few of the templates to match the majority pattern, and to
replace references to tier, component or type to draw the required
values from the equivalent locations in the occurrence structure.

Before commencing any processing, a process is run to look at each
level/component combination and attempt to include the corresponding
template. It is not unusual for some combinations to be missing, so no
error is thrown if a template is not found. The net effect of this
process is to define the macros corresponding to valid level/component
combinations.

The core tier/component processing loop has then been modified to
dynamically work out which macro needs to be called, rather than working
its way through a long if list as was done previously. (This approach
has been left in place for accounts for the moment).

This logic relies heavily on strict alignment of file naming with
component naming. Basically the component type must match the filename,
so some remediation to remove legcy names (like es vs elasticsearch)
will need to be addressed in cmdbs as part of the transition.

The dynamic calling is achieved by using the .vars special variable
to get access to the appropriate macro (aka user directive), and then
passing this value in the macro invocation syntax as the name of
desired macro. Absence of the directive allows us to report an exception
as it means a component is being used at a level where it shouldn't be.
Equally, we could let it silently ignore the error as is done by the
current composite if structure.

More work is needed to fit account level processing to this pattern but
it does not seem unrealistic to do this.

Applying dynamic loading to policies and fragments would also seem to be
quite doable though framents are a bit trickier as they can be in
multiple locations in the cmdb.

A small change was made to setContext.sh to remove the composite
templates from the cache in order to confirm the operation of the new
logic, but more extensive changes can be expected once the account
template processing is aligned, and more generally after the dynamic
including of id, name and resource templates is worked through.

In parallel with these changes, an update has been made to the
freemarker wrapper to support a hierarchy of locations for
template loading. The intention is that this will form the basis for a plugin
mechanism, where additional templates can be added dynamically by
placing them in a known location and adding this to the freemarker
command via environment settings. Thinking a GENERATION_PLUGIN_DIRS
variable with a semicolon separated list of locations (ah la unix
paths) will be the first cut of this.


roleyfoley:
Looks pretty good to me, I think the approach seems reasonable and the naming convention mapping component type to file name works for me. 

How would resources work? would they match the type as well?

PR #789, opened 2019-05-27
<refactor> segment cmk into baseline

roleyfoley:
moves the segment cmk resources into the baseline component as a subcomponent 

- Adds the Key SubComponent which has different engines for the different keys that we have at the segment level ( ssh, cmk, oai) 
- a subcomponent with the same name as the engine is treated as the default key ( others can be added if required ) 
- Uses Cloudformation for any new OAI keys ( https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudfront-cloudfrontoriginaccessidentity.html ) 

**Still to do**
- [x] Add defaults to master data
- [x] Update lookup of baseline components in other components to not directly use Ids


roleyfoley:
@ml019 ready for you review when you are 

The CMK pseudo stack outputs are now created in the baseline component. So its a bit funny but it all seems to work

PR #804, opened 2019-06-07
<refactor> Add transitional assembly support

ml019:
Assembly process now looks in the new providers tree being introduced as part of multi-provider support. This will allow files to be incrementally moved across to the new structure. As a start, this PR includes segment, solution and application templates.

Once https://github.com/codeontap/gen3/pull/802 and https://github.com/codeontap/gen3/pull/803 has gone through, the next step will be to move the id files across, splitting them between the component aspects and the provider aspects.


ml019:
@roleyfoley ready for review

PR #809, opened 2019-06-08
<refactor> Dynamic loading of seg/sol/app templates

ml019:
Dynamically load component templates.

For now level is still being used until we resolve use of same component name in solution and application levels (ecs and spa)

Shift the occurrence processing loop outside of component templates and change them to be self-contained templates. As a result, change all variable usage in the component templates to use local variables.

With the change in indenting, github doesn't really show the first commit very well - better to try
```
git diff HEAD~~ HEAD~
```


ml019:
@roleyfoley this is one that shifts us to dynamic template loading. Please review.

PR #819, opened 2019-06-12
<feature> Static component dependencies

ml019:
Add ability to declare static dependencies of one component on
another.

In general, link processing should load all the component definitions
but this happens at run-time. As a result, when the code is loaded
and it is syntactically checked, any use of definitions from a link
component type will fail.

The suspicion is that such use needs to be reconsidered because it
implies an assumption on the part of the depending component as to
the provider of the dependent component. However, for now this feature
will get around this.


ml019:
@roleyfoley tested with repo-user in icl and worked ok.

PR #827, opened 2019-06-15
<feature> Resource Groups and placements

ml019:
Allow different subsets of a component's resources to be deployed
to different accounts/regions.

The subsets are called Resource Groups. The deployment details is called a placement.

For now, placements are assumed to provide provider and deployment
framework information. A subsequent change will look up the provider
based on the account. A default provider of AWS and deployment framework
of cloud formation are used for backwards compatibility.

The component designer decides which resources can be deployed to different
placements, thus deciding how many resource groups are needed.

Each group is then given a name in the component configuration, independent
of the provider. Each group is also assigned a name for the
placement to be used.

Every component has a "default" resource group where the bulk of its
resources are typically created. It uses "PRODUCT_PLACEMENT" as its
placement name.

Each resource group can have "shared" attributes which are expected for
all implementations. It can also have provider specific attributes which
will be expected depending on the placement of the resource group.

Placement names are mapped to actual placement details via a placement
profile. Placement profiles can be configured at the product and tenant
level, and support qualification.

When a component needs to be processed, the logic first needs to
determine the placement details to determine the provider to be used.
This in turn defines the set of attributes required, and the processing
needed to create the component and manage its state.

To implement the above, the macros for creating the component
configuration have been adjusted to create the default resource group
for each component.

The template loading logic was extended to work off the resource group
placement information, along with the process of determining the
processing macros to be used.

The handling of special attributes used for profile and placement
handling, along with the standard attributes all component have, was
modified such that they are added to the configuration provided by the
component configuration files. This permits documentation generated from
the component configuration to show the full range of configuration
options. Attributes are ordered so these framework injected attributes
appear last in the list of attributes, so that documentation can focus
the reader's attention on the unique attributes of the component.

A hook has also been added to permit each provider specific
implementation of a component to specify the provider specific services
it uses, and for these to then be loaded whenever that implementation is
required. A subsequent change will use this mechanism to move all of the
remaining service configuration in the original template tree across to
the new provider based tree.

### DIFF
--- a/providers/shared/component/component.ftl
+++ b/providers/shared/component/component.ftl
@@ -78,95 +78,294 @@
 [#assign USERPOOL_CLIENT_COMPONENT_TYPE = "userpoolclient" ]
 [#assign USERPOOL_AUTHPROVIDER_COMPONENT_TYPE = "userpoolauthprovider" ]
 
+[#-- Component configuration is extended dynamically by each component type --]
+[#assign componentConfiguration = {} ]
 
+[#-- Resource Groups --]
+[#assign DEFAULT_RESOURCE_GROUP = "default"]
+[#assign DNS_RESOURCE_GROUP = "dns"]
+
+[#-- Attributes are shared across providers, or provider specific --]
 [#assign SHARED_ATTRIBUTES = "shared"]
 
-[#function formatResourceGroupName componentName groupName]
-    [#return formatName(componentName, groupName)]
-[/#function]
-
+[#-- Placements --]
 [#assign PRODUCT_PLACEMENT = "product"]
 [#assign DNS_PLACEMENT = "dns"]
 
-[#-- Lookup shared component definition --]
-[#macro includeSharedComponentConfiguration component ]
-    [#-- Handle object or type string --]
-    [#local type = component ]
-    [#if component?is_hash]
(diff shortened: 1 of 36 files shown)


ml019:
@roleyfoley resource groups and placements are finally here after me talking about them for too long.

I'll start work on the next PR get the services configuration working as expected as well as pulling across the service config still in the old template directory structure.

I simplified the component macros which resulted in changes to a number of the shared component files, together with moving to the macros the addition of profile and standard attributes to the component configuration for completeness of documentation. 

The core changes were in component.ftl and common.ftl, however where the main processing loops are.

providers/shared/component/component.ftl
```diff
@@ -78,95 +78,294 @@
...
+
+[#function getComponentChildren type]
+    [#return (componentConfiguration[type].Components)![] ]
+[/#function]
+
+[#function getResourceGroupPlacement resourceGroup profile]
+    [#return
+        {
+            "Provider" : "aws",
+            "Region" : "ap-southeast-2",
```
roleyfoley: Should this be defaulted in the masterdata rather than in code?

ml019:
@roleyfoley Think is reflects what we discussed?

PR #828, opened 2019-06-17
<feat> Federated Roles

roleyfoley:
Federated roles provide cloud provider resource credentials to users based on external identity sources. This allows for users to use the same credentials to access an application and cloud provider resources directly 

Example: 
```
                "cloudrole" : {
                    "federatedrole" : { 
                        "Instances" : {
                            "default" : {
                                "DeploymentUnits" : [ "cloudrole" ]
                            }
                        },
                        "Fragment" : "_cloudrole",
                        "Links" : {
                            "userpool" : {
                                "Tier" : "dir",
                                "Component" : "app-userpool"
                            }
                        },
                        "Assignments" : {
                            "authenticated" : {
                                "Type" : "Authenticated",
                                "Links" : {
                                    "datastore" : {
                                        "Tier" : "db",
                                        "Component" : "data-s3",
                                        "Role" : "none"
                                    }
                                }
                            }
                        }
                    }
                }
```

Fragment
```
[#case "_cloudrole" ]

    [#if _context.Assignment == "authenticated" ]
        [#if (_context.Links["datastore"]!{})?has_content ]
            [@Policy  
                [
                    getS3Statement( 
                        [
                            "s3:GetObject",
                            "s3:PutObject"
                        ],
                        _context.Links["datastore"].State.Attributes["NAME"],
                        "submissions/$\{cognito-identity.amazonaws.com:sub}",
                        "*"
                    ) + 
                    getS3BucketStatement(
                        [
                            "s3:ListBucket",
                            "s3:ListBucketVersions"
                        ],
                        _context.Links["datastore"].State.Attributes["NAME"],
                        "submissions/$\{cognito-identity.amazonaws.com:sub}"
                    )
                ]
            /]
        [/#if]
    [/#if]

    [#break]
```
In the example all authenticated users can put objects into an S3 bucket if the prefix of the file matches the pattern `submissions/<the users cognito id>` which essentially provides permission control home drives for users

The authenticated type here will apply to all authenticated users. Instead of this you can create Rule based assignments which check user claims and provided roles based on the values of the claims ( e.g. if a user has claim of `isAdmin` you can apply a role based on this being true or not. Rules have priorities and apply the role from the first match.

In AWS federated roles are based on Cognito Federated Identity pools which map attributes from multiple authentication providers, ( Cognito Userpools, SAML, OIDC, Facebook, Google, Amazon ) into an IAM Role. 

The role in turn then can have policies which allow it to access AWS resources like  you normally would 

A `federatedrole` has a parent component which hosts the identity pool and the role mappings and then has `federatedroleassignments` which mange the roles, the permissions they have on AWS resources and the conditions users must meet to be assigned a given role. 

This PR also
 - removes the existing identity pool and role based resources from the `userpool` component. Since federatedroles are a separate concept from user pools and userpools aren't specifically required for a federated role
- Fixes up some bugs in the authprovider cli based config on the `userpool`

### DIFF
--- a/aws/templates/common.ftl
+++ b/aws/templates/common.ftl
@@ -1359,7 +1359,9 @@ behaviour.
                                 },
                                 "Component" : {
                                     "Id" : componentId,
+                                    "RawId" : component.Id,
                                     "Name" : componentName,
+                                    "RawName" : component.Name,
                                     "Type" : type
                                 },
                                 "Instance" : {
--- a/aws/templates/resource/resource_cognito.ftl
+++ b/aws/templates/resource/resource_cognito.ftl
@@ -138,16 +138,47 @@
     ]
 [/#function]
 
-[#function getIdentityPoolCognitoProvider UserPoolId userPoolClientId serverSideToken=true ]
-
-    [#assign userPoolRef = getReference(UserPoolId, NAME_ATTRIBUTE_TYPE)]
-    [#assign userpoolClientRef = getReference(userPoolClientId )]
+[#function getIdentityPoolCognitoProvider userPool userPoolClient ]
 
     [#return
+        [
+            {
+                "ProviderName" : userPool,
+                "ClientId" : userPoolClient,
+                "ServerSideTokenCheck" : true
+            }
+        ]
+    ]
+[/#function]
+
+
+[#function getIdentityPoolMappingRule priority claim matchType value roleId ]
+    [#return 
+        [
+            {
+                "Priority" : priority,
+                "Rule" :  {
+                    "Claim" : claim,
+                    "MatchType" : matchType,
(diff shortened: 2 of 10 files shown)


roleyfoley:
*note* 
Roles need to be crated as part of the template subset due to the condition which restricts the role to only be used by the identity pool.

roleyfoley:
> **note**
> Roles need to be created as part of the template subset due to the condition which restricts the role to only be used by the identity pool.
> 
> Not sure what the approach should be. Do we remove the condition and move it back to the iam subset or keep the condition which secures the role more?

Fixed made it work as a double pass. The identity pool id condition will be added on a second iam pass

roleyfoley:
@ml019 Should be good to go now. Just had to do some cleanup

PR #830, opened 2019-06-18
<feature> Generic logging

ml019:
Implement a logging approach aligned to that used in the script
utilities.

This makes the framework available to all providers.

The level of logging included is controlled by GENERATION_LOG_LEVEL.

This means macros can be left enabled but won't generate output unless
their corresponding level is enabled.

Refactor the existing exception detection to work off the new format.

Support serialising messages as comments for scripts so the messages
don't break the script functionality.


roleyfoley:
@ml019 could you fix this up please?

ml019:
@roleyfoley ready for review - thanks

PR #833, opened 2019-06-23
<refactor> Component service dependencies

ml019:
A provider implementation of a component needs to draw on the use of one
or more services that the provider offers. Thus a provider can be
configured with one or more services, which in turn offer utility
defininitions/ and macros for the service. Both the component and the
service need to support the same deployment framework for the service to
be usable.

All AWS component implementations now define their service dependencies.
These are in the "id" file in the provider component tree. All of the
legacy id, name, policy and resource files have been moved under the aws
serivces tree. In general a service can have an id, name, policy and
resource file. There is also an id, name, policy and resource file at
the top level of the services tree for common definitions across all
services.

The component id files for a provider are included first in order that the
service dependencies are known, then the services they depend on are loaded,
then finally the component setup and state includes are loaded. In this
way, static code analysis during inclusion shouldn't throw any errors.

Each component can have an id, name, setup and state file, again with an
equivalent at the root of the components tree for the provider to
capture anything that needs to be shared. These shared files are loaded
as part of the initial inclusion of a provider for the first time.

There is a service.ftl file under services which is loaded as part of
the basic provider inclusion, which is where constants for all the
services are defined. Service dependencies can thus be defined in terms
of constants specific to the provider.

Adding a provider should thus start by defining the services it
provides, then the component can be built consuming the definitions of
the services. Hence ids can by common across deployment frameworks, while
the policy and resource definitions are deployment framework specific.

All the shared component definitions have been renamed to be id files
for consistency, and the files named after the components removed. This
aligns the shared tree with the provider specific component trees.

In order to ensure configuration files for a service or component are
only processed once, a cache has been added to the component inclusion
process to detect this. While there is a similar feature in the
inclusion of individual files, this cache provides coarser granularity
in order to speed up the inclusion checks that need to be done every
time a component occurrence is generated for processing.

Remove processing of component dependences are the introduced of
services should resolve the missing definitions when static analysis of
code during inclusion happens.

Introduce the idea of a bootstrap file which knows what else needs to be
loaded, and can be varied depending on the include mechanism.


ml019:
Lots of change but most of the key logic changes are in the /engine files. provider.ftl has all the inclusion logic, component.ftl has all the component configuration and core processing logic. PLan to add an occurrence.ftl with the core occurrence logic too and refactor it out of common.ftl.

Next step is to add dynamic overall template selection based on formats as discussed.

ml019:
@roleyfoley rebased and good to go when you are.

PR #834, opened 2019-06-24
<refactor> baseline component lookups

roleyfoley:
This refactor introduces the idea of a baseline profile which is used to determine the appropriate baseline resources for a component to use. 

The profile defines a collection of Component names which are used to look up specific subcomponents on the base line component. 

The baseline components can be filtered to ensure that a given component only looks up the resource which it needs 

Only applied the configuration to SPA at the moment to fix a bug with OAI lookups. But if it works as expected will rollout across all components


providers/aws/components/baseline/state.ftl
```diff
@@ -2,6 +2,39 @@
 [#assign LOCAL_SSH_PRIVATE_KEY_RESOURCE_TYPE = "sshPrivKey" ]
 [#assign AWS_SSH_KEY_PAIR_RESOURCE_TYPE = "sshKeyPair" ]
 
+[#function formatSegmentCMKId ]
```
ml019: with the work I've done on services, I think these functions should be in the id file not the state file. hat was the logic for moving them?

providers/shared/components/baseline/baseline.ftl
```diff
@@ -6,6 +6,71 @@
     [#return formatSegmentResourceId(SEED_RESOURCE_TYPE)]
 [/#function]
 
+
+[#function getBaselineLinks baselineProfileName baselineComponentNames idsOnly=true  ]
```
ml019: Think this is AWS specific so really should be in aws tree?

roleyfoley: It is intended to be a global lookup. Everyone would have baseline and the baselineComponentNames filter can be used for provider specific lookups

providers/aws/components/baseline/state.ftl
roleyfoley: To get rid of the KMS id file mainly. These functions where for baseline only Id formatting and are for legacy support

PR #836, opened 2019-06-24
<refactor> Template processing

ml019:
This is the first of several PRs to rework the processing of templates
to add multi-provider support.

The first step is to factor out some generic processing for standard
output formats (text, json), which is what this change does.

Next step will be to make the top level macro for a template
dynamically.

This PR should be reviewed/merged after https://github.com/codeontap/gen3/pull/833.

### DIFF
--- a/aws/templates/createAccountTemplate.ftl
+++ b/aws/templates/createAccountTemplate.ftl
@@ -15,4 +15,4 @@
         [#break]
 [/#switch]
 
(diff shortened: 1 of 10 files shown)


roleyfoley:
@ml019 This one looks like it needs to be rebased based on the merge of #833

ml019:
@roleyfoley rebased so should be good to merge

PR #837, opened 2019-06-25
<refactor> use baselineprofile for all components

roleyfoley:
This PR moves to using the baseline profile introduced in #834 for all components. 

I've commented out the setContext for the operations and data buckets to make sure that they aren't in use 

Also has some fixes for ES based on the split of the AWS identity pool from our userpool component


roleyfoley:
@ml019 Should be good to go now.

PR #838, opened 2019-06-29
<feature> Dynamic output formatting

ml019:
Providers now have deployment framework configuration to control the
format of outputs produced.

Core processing of text and json formats has been added, with XML an
easy extension in the future.

Each provider framework then defines one or more output types, which then
leverage the formats offered by the core routines.

The shared provider offers a default framework which provides script and
JSON types. The script type supports both bash and powershell as
formats.

The macro used to format the output is dynamically loaded in the same
way as the component processing routines. Critically, this means
processing is now single pass for all templates, rather than the
previous five pass arrangement.

The genplan support has been extended to include provider, deployment
framework, output type and output format in each of the steps it
defines. Next step will be to extend this to include account and region
information from the occurrence information.

The changes are backwards compatible with account template processing as
well.

There will be a follow up commit which will convert the aws provider
across to use the shared output formatters and remove the cfCli,
cfConfig and cfDebug macros.


ml019:
@roleyfoley this one will cut down template processing considerably I think.

I still need to profile the timing because, for instance, a segment eip pass still takes a long time - but a lot less than previously.

I'll put through some cleanup of the AWS provider as a separate PR.

PR #843, opened 2019-07-08
<feature> Topic Component and S3 Notifications

roleyfoley:
This PR adds support for pub/sub topics as dedicated components and also extends S3 change notifications to support topics and lambda as notification endpoints 

A topic creates an SNS topic and has a subcomponent of Subscriptions. The subscription subcomponent is mainly used when the topic needs to send to HTTP(s), or email locations since they can't really be defined as our other components. 

Components can link to an SNS topic for other publish access if required


roleyfoley:
This PR now introduces the ability to subscribe to alert notifications using a topic and its subscriptions  and filter based on the severity of the alert 

A new set of reference/master data has been added as Alert Rules. Rules specify a severity along with a collection of destinations which will receive the alert. Destinations are links off to supported endpoints for Cloudwatch alarms, at the moment only SNS is supported 

An Alert Profile is a collection of rules which are applied at the component level. When an alert is added it now uses its profile to lookup the alert actions it needs to add to an alarm 

Example 
```
    "AlertRules" : {
        "All" : { 
          "Severity" : "error",
          "Destinations" : {
            "Links" : {
                "email" : {
                    "Tier" : "mgmt",
                    "Component" : "alerts",
                    "Instance" : "",
                    "Version" : ""
                }
            }
          } 
        }
    },
    "AlertProfiles" : {
        "default" : {
            "Rules" : [ "All" ]
        }
    },
``` 
This rule will send all logs of error and above ( based on log severity filters ) to the alerts component 
```
        "mgmt" : {
            "Components" : {
                "alerts" : {
                    "topic" : {
                        "Instances" : {
                            "default" : {
                                "DeploymentUnits" : [ "alerts"]
                            }
                        },
                        "Subscriptions" : {
                            "users" : {
                                "Links" : {
                                    "roleyfoley" : {
                                        "Tier" : "external",
                                        "Component" : "roleyfoley"
                                    }
                                }
                            }
                        }
                    }

                }
            }
        }
``` 

The alerts Topic has a subscription to an external endpoint `roleyfoley` which is configured as an email endpoint  via link setting attributes 

```
{
    "ROLEYFOLEY_SUBSCRIPTION_ENDPOINT" : "roleyfoley@gmail.com",
    "ROLEYFOLEY_SUBSCRIPTION_PROTOCOL" : "email-json"
}
``` 

Then when you configure an alert on a component it will add this topic to the alert actions when required. In this exmaple 
```
                        "Alerts" : {
                            "5XXLBResponse" : {
                                "Description" : "A high number of 5xx response codes have been sent from load balanced targets",
                                "Name" : "5XXLBResponse",
                                "Metric" : "HTTPCode_Target_5XX_Count",
                                "Threshold" : 1,
                                "Severity" : "error",
                                "Statistic" : "Sum",
                                "Periods" : 2
                            },
                            "4XXLBResponse" : {
                                "Description" : "A high number of 5xx response codes have been sent from load balanced targets",
                                "Name" : "4XXLBResponse",
                                "Metric" : "HTTPCode_Target_4XX_Count",
                                "Threshold" : 1,
                                "Severity" : "info",
                                "Statistic" : "Sum",
                                "Periods" : 2
                            }
                        }
``` 

The topic will receive alerts when 5XXLBResponse goes into an alarm state but not if 4XXLBResponse does

roleyfoley:
**NOTE** This PR also removes the existing segment baseline SNS topics in favour of configuring your own topic so that you can control how it used and manage subscriptions. This saved duplicating all of topic configuration in the baseline component.

ml019:
@roleyfoley so for projects like appointments where I am using the current segment support, I'll need to explicitly add a topic, along with rules to connect currently alerting components with the topic?

roleyfoley:
@ml019 It would yeah
Adding it to baseline would have required SubComponents of SubComponents and duplicating most of what's in the topic already. With the move to getting rid of levels it seemed like the way to go

PR #844, opened 2019-07-09
<refactor> Move master data under respective provider

ml019:
Provider inclusion process now loads the master data. For now its
wrapped in an ftl file until we can directly pull in the json.

Note that Id and Name atributes need to be explciitly provided where
expected by code, as the files are not not subject to the Id/Name
injection that occurs for the blueprint.


providers/aws/masterData.json
```diff
@@ -0,0 +1,2462 @@
+{
```
roleyfoley: Should this file still be here? would it be replaced by masterData.ftl?

ml019: The current process to update the aws master data assumes a json file. So I left the masterdata.json there but the update script now creates an ftl file from the json file.

ml019:
@roleyfoley good for review when you are

PR #847, opened 2019-07-09
<fix> Align blueprint generation with output framework

ml019:
Use the output framework to generate the blueprint.

Because of the format required, redefine the key processing macro to
generate the output required.

Switch to using the config subset as it is a little simpler but that;s
hidden inside the genplan generation.

### DIFF
--- a/aws/createTemplate.sh
+++ b/aws/createTemplate.sh
@@ -238,9 +238,9 @@ function process_template_pass() {
         pass_region_prefix["${p}"]=""
       done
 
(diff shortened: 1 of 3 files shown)


ml019:
@roleyfoley I didn't confirm this worked with the automation framework variant. Would you mind confirming its ok?

roleyfoley:
Looks good, I think createBuildblueprint.ftl would need the same treatment

ml019:
@roleyfoley updated createbuildlbueprint as well. I removed the error stuff as discussed.

PR #852, opened 2019-07-11
<refactor> Remove commonSegment.ftl

ml019:
Move macros to associated services.

Remove the file from /engine and all includes of it.


providers/aws/services/cw/resource.ftl
```diff
@@ -44,6 +44,15 @@
     /]
 [/#macro]
 
+[#macro createVPCLogGroup id name retention ]
```
roleyfoley: Do we need this one? looks like it just creates a Log Group

ml019: Removed.

PR #853, opened 2019-07-11
<fix> Update Expo config setup to support baseline

roleyfoley:
With the changes to using the baseline profile the APPDATA_BUCKET and OPSDATA_BUCKET variables aren't available until the setup stage instead of the state stage like the used to be 

This PR updates the expo build configuration to rely on a dedicated config file rather than using the build blue print. 

The build blueprint is still required to determine where the config file lives but the actual values required by the build are now baked into the config file provided for the build

### DIFF
--- a/providers/aws/components/mobileapp/setup.ftl
+++ b/providers/aws/components/mobileapp/setup.ftl
@@ -41,9 +41,7 @@
         attributes +
         defaultEnvironment(occurrence, {})
     ]
-
-    [#local buildConfig += attributes ]
-
+    
     [#local fragment = getOccurrenceFragmentBase(occurrence) ]
 
     [#local contextLinks = getLinkTargets(occurrence) ]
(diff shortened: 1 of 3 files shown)


providers/aws/components/mobileapp/setup.ftl
```diff
@@ -10,16 +10,40 @@
...
+            "APP_BUILD_FORMATS" : solution.BuildFormats?join(","),
+            "BUILD_REFERENCE"   : getOccurrenceBuildReference(occurrence),
+            "OPSDATA_BUCKET"    : operationsBucket,
+            "APPDATA_BUCKET"    : dataBucket
+        } + 
+        attributes +
+        defaultEnvironment(occurrence, {})
+    ]
+
+    [#local buildConfig += attributes ]
```
ml019: Looks redundant based on line above?

PR #882, opened 2019-07-20
<fix> SPA baseline processing

ml019:
Ensure baseline links defined before use in generating _context and handle case of missing baseline components.

Improve error handling in baseline link routines so they don't result in freemarker exceptions. This way, the actual source of any errors can be reported.


ml019:
@roleyfoley @RossMurr4y going to merge this one myself as its breaking my ICS deploys

PR #886, opened 2019-07-20
<fix> userpool role id lookup

roleyfoley:
fix to restore role for userpool after it was removed during the refactoring of federated roles


ml019:
@roleyfoley thanks for the prompt response

PR #887, opened 2019-07-20
<fix> Userpool role creation resource dependency

ml019:
Check for the correct deployment unit should use the role id not the pool id.

@roleyfoley please confirm this won't break any other deployments for products using user pools.


roleyfoley:
If it does it’s a bug fix and should probably be fixed up any way

PR #888, opened 2019-07-20
<feature> sentry releases

kshychko:
Introducing a script that allows to upload source maps files for a specified release. Sentry CLI is installed using yarn, configuration is done via environment variables on Jenkins level.

### DIFF
--- a/aws/runExpoAppPublish.sh
+++ b/aws/runExpoAppPublish.sh
@@ -259,12 +259,16 @@ function main() {
 
     if [[ -f "${AUTOMATION_DATA_DIR}/current-app-manifest.json" ]]; then 
         EXPO_CURRENT_APP_VERSION="$(jq -r '.version' < "${AUTOMATION_DATA_DIR}/current-app-manifest.json" )"
+        EXPO_CURRENT_APP_REVISION_ID="$(jq -r '.revisionId' < "${AUTOMATION_DATA_DIR}/current-app-manifest.json" )"
     fi 
   fi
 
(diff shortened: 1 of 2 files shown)


aws/runExpoAppPublish.sh
```diff
@@ -289,9 +293,13 @@ function main() {
 
   # Create master export 
   info "Creating master OTA artefact with Extra Dirs: ${EXPO_EXPORT_MERGE_ARGUMENTS}"
-  expo export --public-url "${PUBLIC_URL}" --output-dir "${SRC_PATH}/app/dist/master/" ${EXPO_EXPORT_MERGE_ARGUMENTS}  || return $?
+  expo export --dump-sourcemap --public-url "${PUBLIC_URL}" --output-dir "${SRC_PATH}/app/dist/master/" ${EXPO_EXPORT_MERGE_ARGUMENTS}  || return $?
```
roleyfoley: Probably don't need the master export for the source maps. The master export is only run to generate an expo client manifest with multiple versions of expo installed. Its not used for a real mobile app

aws/runSentryRelease.sh
```diff
@@ -0,0 +1,89 @@
...
+    done
+
+}
+
+
+function main() {
+
+  options "$@" || return $?
+
+  yarn global add @sentry/cli || return $?
```
roleyfoley: Might be safer to make this optional so that we choose when to update the sentry-cli version or lock it to a major version

PR #905, opened 2019-07-24
<refactor> APIGW: move deployment details in to swagger update

roleyfoley:
Now that the api gateway updates are run as part of the setup script we can access the attributes for the deployment

This PR moves the updates added to a swagger spec as part of the publish process into the standard template extensions. As the deployment updates only affect information and try it now functionality this shouldn't impact the API deployment


roleyfoley:
This also adds publish support for openapi3 api specs

ml019:
@roleyfoley Cool. Much nicer this way

PR #910, opened 2019-07-25
<fix> Default port id and name

ml019:
Use the port key as the default Id and Name


roleyfoley:
This one has been replaced by #911

PR #938, opened 2019-07-31
<feature> CDN Component

roleyfoley:
This PR adds support for a dedicated CDN solution level component and replaces the existing solution level SPA CDN deployment

A CDN component can be used by multiple HTTP based services to provide content distribution services and can also be used for routing between different origins while maintaining a consistent client facing endpoint 

The component itself manages the default configuration for the disribution, certificates, default error responses, default document etc 

A route subcomponent provides the link between the distribution and the origin. Multiple routes can be configured with each route having its own PathPattern applied. 

Example of the Solution Configuration 

```
        "web" : {
            "Components" : {
                "catalog-cdn" : {
                    "cdn" : {
                        "Instances" : {
                            "default" : {
                                "DeploymentUnits" : [ "catalog-cdn"]
                            }
                        },
                        "Certificate" : {
                            "IncludeInHost" : {
                                "Component": false,
                                "Host" : true
                            },
                            "Host" : "data"
                        },
                        "Routes" : {
                            "default" : {
                                "PathPattern" : "_default",
                                "Origin" : {
                                    "BasePath" : "public/",
                                    "Link" : {
                                        "Tier" : "web",
                                        "Component" : "catalog"
                                    }
                                }
                            }
                        }
                    }
                },
                "catalog" : {
                    "s3" : {
                        "Instances" : {
                            "default" : {
                                "DeploymentUnits" : [ "catalog"]
                            } 
                        },
                        "Links" : {
                            "cdn" : {
                                "Tier" : "web",
                                "Component" : "catalog-cdn",
                                "Route" : "default",
                                "Direction" : "inbound"
                            }
                        }
                    }
                }
            }
        },
``` 

This is an S3 Backed CDN which can be used for serving content which isn't specifically generated by a code build. 

The CDN has a single route called `default` which uses a path pattern of `_default` this means that for all requests that don't match another path pattern use this one 

The Origin.BasePath property allows you to specify where requests will be sent to in the S3 bucket 

The catalog s3 bucket also needs an inbound link to the CDN in order to enable OAI based access to the bukcet. This link also uses the Origin.BasePath paramter to specify the bucket permissions 

SPA application components must have at least one inbound link to a CDNRoute and the linking CDN should have a link to the SPA.

For an SPA this allows it to invalidate the cache for the data that it provides to the CDN

### DIFF
--- a/providers/aws/components/cdn/id.ftl
+++ b/providers/aws/components/cdn/id.ftl
@@ -1,0 +1,15 @@
+[#ftl]
+[@addResourceGroupInformation
+    type=CDN_COMPONENT_TYPE
+    attributes=[]
+    provider=AWS_PROVIDER
+    resourceGroup=DEFAULT_RESOURCE_GROUP
+    services=
+        [
+            AWS_CLOUDFRONT_SERVICE,
+            AWS_WEB_APPLICATION_FIREWALL_SERVICE,
+            AWS_BASELINE_PSEUDO_SERVICE,
+            AWS_CERTIFICATE_MANAGER_SERVICE,
+            AWS_ROUTE53_SERVICE
+        ]
+/]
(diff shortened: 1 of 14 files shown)


roleyfoley:
@ml019 Still need to test with an SPA but this should be ready to review

roleyfoley:
@ml019 tested and confirmed working for S3 and SPA

PR #943, opened 2019-08-07
Adding composite output filtering for azure.

RossMurr4y:
If after filtering the composite outputs the resulting file is empty, filter it for Azure ARM template outputs instead.


ml019:
@RossMurr4y Thanks for your first PR. May it be the first of many more...

PR #952, opened 2019-08-14
<feature> Autoscaling ECS & Ec2 

roleyfoley:
Uses AWS Application AutoScaling ( for some reason AWS has 3 autoscaling services ) to configure autoscaling for ECS based Services

Autoscaling basically uses cloudwatch Metrics/Alarms to update the desired count of services. So we use our existing metric and alarm setup combined with links to enable a service to track the metrics of another component. Services can also track their own metrics if a link is not specified 

Autoscaling is configured using `ScalingPolicies` on a service. Multiple policies can be applied to a service and the one which results in the highest desired count wins. 

Scaling Policies have 3 different types 
- Stepped - An alarm is created which sets a water mark level - Adjustments are then made relative to the water mark - Changes the desired Count of the service 
- Tracked - You specific a metric and a target value and AWS will automatically scale the services to maintain the metric at the target value - Changes the desired Count of the Service 
- Scheduled - Allows you to change the Min and Max counts of the service to limit the scalability of the service based on time ( e.g. business hours vs after hours ) 

Example 

```
                            "squid" : {
                                "Engine" : "fargate",
                                "DeploymentUnits" : [ "squid" ],
                                "Profiles" : {
                                    "Processor" : "default"
                                },
                                "Containers" : {
                                    "_squid" : {
                                        "Version" : "latest",
                                        "Cpu" : 256,
                                        "Memory" : 512,
                                        "MaximumMemory" : 512,
                                        "Ports" : {
                                            "squid" : {
                                                "IPAddressGroups" : [ "_global" ]
                                            }
                                        }
                                    }
                                },
                                "NetworkMode" : "awsvpc",
                                "ScalingPolicies" : {
                                    "TrackMemoryUsage" : {
                                        "Type" : "Tracked",
                                        "TrackingResource" : {
                                            "Link" : {
                                                "Tier" : "app",
                                                "Component" : "apphost",
                                                "Instance" : "",
                                                "Version" : ""
                                            },
                                            "MetricTrigger" : {
                                                "Name" : "MemoryUsage",
                                                "Metric" : "MemoryUtilization",
                                                "Threshold" : 2,
                                                "Severity" : "info",
                                                "Statistic" : "Average",
                                                "Unit" : "Percent"
                                            }
                                        },
                                        "Tracked" : {
                                            "TargetValue" : 2
                                        }
                                    },
                                    "BusinessHours" : {
                                        "Type" : "Scheduled",
                                        "Scheduled" : {
                                            "ProcessorProfile" : "businesshours",
                                            "Schedule" : "cron(0 9 ? * MON-FRI *)"
                                        }
                                    },
                                    "AfterHours" : {
                                        "Type" : "Scheduled",
                                        "Scheduled" : {
                                            "ProcessorProfile" : "afterhours",
                                            "Schedule" : "cron(0 18 ? * MON-FRI *)"
                                        }
                                    }
                                }
                            }
```

This service has 3 policies defined 

- TrackMemoryUsage - Which tracks the memory usage of the ECS cluster `apphost` 
- BusinessHours - which updates the Min/Max Counts at 9 am to match those in the `businesshours`  processor profile 
- AfterHours - which updates the Min/Max Counts at 6pm to matchin the afterhours processor profile 

This PR also introduces processor Profiles for ECS services, the current solution level configuration is backwards compatible as well


roleyfoley:
A couple of extra changes while I was going along 

- renamed the CloudWatch alarm setup to createAlarm - Unit is actually just the measurement of the metric not something special which I thought it was 
- getExistingReference fix for passing REFERENCE_ATTRIBUTE_TYPE as a attributeType instead of leaving it null ( used with the metric resource lookups )

engine/common.ftl
```diff
@@ -1214,6 +1285,7 @@ behaviour.
                         }]]
                         [#break]
                     [#case "Output" ]
+                        [@debug message="output lookup" context={ "resourceId" : resource.Id, "value" : value } enabled=true /]
```
ml019: Should this be left enabled?

roleyfoley: Just removed it. I hadn't pushed the change where I cleaned that out

roleyfoley:
Taking back to WIP so that I can add ec2 autoscaling. Almost got it going and the config is very similar

roleyfoley:
PR has now been updated to include ec2 based autoscaling for ecs cluster hosts and for compute clusters 

@ml019 please review

engine/common.ftl
```diff
@@ -631,6 +634,74 @@ behaviour.
...
+        [/#if]
+    [#else]
+        [@fatal
+            message="Processor profile does not have a MinCount"
+            context=processorProfile
+        /]
+    [/#if]
+
+    [#if fixedDesiredCount?has_content ]
+        [#local desiredCount = fixedDesirecCount ]
```
RossMurr4y: > fixedDesire**c**Count  :)

roleyfoley: Fixed, who needs spell checkers when you've got reviewers :)

PR #953, opened 2019-08-14
<feature> Round out context based model support

ml019:
Support full template generation via the "context" model.

It is now backwards compatible with existing subcomponent use, but
depending on how the setup routines are defined, can also allow
subcomponents to be created independent of their parents. This should
enable a smooth transition of services and tasks to be independent
components.

Text output generation has been refactored to support the ability to
have multiple sections in the output, define the order in which the
sections are serialised and allow for a "converter" per section to
convert the data structure used in the section to a text format. This
feature is used to ensure that generation passes are only run once, even
though requested by multiple components via the selected deployment
unit.

The context model supports qualification of any content and is recursive.

The context model is not yet the default - to enable it,

```
export GENERATION_MODEL=context
```

Fix a bug where an "Enabled" setting was being added to the environment.

Also add explicit id/names for some internal links.


ml019:
As part of testing, I found that the lambda deployment type was defined on the lambda component rather than on the individual functions, so I shifted it as part of this change.

roleyfoley:
Looks pretty good and can see the idea. Still need to work on getting my head around the execution. But happy to merge in

PR #954, opened 2019-08-15
<feature> DB: Major version update support

roleyfoley:
Adds support for major version upgrades in RDS 

By default when the engine version parameter is updated on an RDS instance the instance will be removed and replaced with a new instance instead of upgrading the instance 

This PR makes a couple of changes to support RDS upgrading an instance when the major version is changed 

- Parameter groups  - have some weird behaviour where the update of the family parameter can not be changed once it has been set, but the parameter does not trigger the replacement of the group. So instead we now create a new parameter group for each family ( engine + engine version ) combination 
- Option groups - Support having their version updated, however this requires replacement which means the instance is removed instead of the normal update process. So we also create a new option group per family. This also ensures compatibility with snapshots that have been manually created 
- EngineVersion - This parameter is now mandatory instead of defaulting to a value for each engine type. This ensures that we know the version of RDS that is running and removes the requirement for code changes when an engine version is deprecated by the cloud provider 

Tested with a new instance and upgraded from 9.4 -> 9.5 -> 9.6 and it went through smoothly

### DIFF
--- a/providers/aws/components/db/setup.ftl
+++ b/providers/aws/components/db/setup.ftl
@@ -18,9 +18,6 @@
     [#-- Baseline component lookup --]
     [#local baselineLinks = getBaselineLinks(occurrence, [ "Encryption" ] )]
     [#local baselineComponentIds = getBaselineComponentIds(baselineLinks)]
-
-    [@debug message="Baseline Ids" context=baselineComponentIds enabled=true /]
-
     [#local cmkKeyId = baselineComponentIds["Encryption"]!"" ]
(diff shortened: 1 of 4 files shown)


ml019:
@roleyfoley this will require solution updates?

roleyfoley:
No solution updates required. But the initial replacement of the parameter group will cause the RDS instance to reboot

roleyfoley:
Actually I lie, it will if the EngineVersion parameter has not been set

PR #956, opened 2019-08-19
<refactor> Remove support for type from component id

roleyfoley:
Removes using the component Id to determine the type of the component. 
We have a number of ways to determine the type of a component and using the Id is a bit of magic that we don't really need when we have explicit Type paramters


ml019:
Think you also will need to address https://github.com/codeontap/gen3/blob/93557222e003404f6199d744ea38bd3ff0c8983d/engine/component.ftl#L269 if you want to remove types from ids. The component id and type routines only return the first part of a hyphen separated value - so you still won't be able to use a hyphen separated value as the id.  

Maybe need to check if the last token is the same as the type and remove it if present, but otherwise return the entire value.

PR #967, opened 2019-08-25
<feature> Core setting namespaces

roleyfoley:
Adds support for settings namespaces. Namespaces allow for custom settings structures for components. At the moment settings are lookedup based on fixed values based on the name of occurrence details. 

SettingNamespaces work in a similar way to domain naming where different occurrence name sections are combined to generate a new namespace. 

Combining Namespaces with deployment profiles offers a way to provide settings to groups of components 

Example 

```
    "DeploymentProfiles" : {
        "backend" : {
            "Modes" : {
                "*" : {
                    "service" : {
                        "SettingNamespaces" : {
                            "base" : {
                                "Name" : "backend",
                                "Match" : "partial",
                                "IncludeInNamespace" : {
                                    "Tier" : true,
                                    "Component" : false,
                                    "Type" : false,
                                    "SubComponent" : false, 
                                    "Instance" : true,
                                    "Version" : false,
                                    "Name" : true
                                }
                            }
                        }
                    },
                    "task" : {
                        "SettingNamespaces" : {
                            "base" : {
                                "Name" : "backend",
                                "Match" : "partial",
                                "IncludeInNamespace" : {
                                    "Tier" : true,
                                    "Component" : false,
                                    "Type" : false,
                                    "SubComponent" : false, 
                                    "Instance" : true,
                                    "Version" : false,
                                    "Name" : true
                                }
                            }
                        }
                    }
                }
            }
        }
    }
```

assuming the following occurrence properties and the backend profile assigned to a component 
- Tier: `application`
- Instance: `foo`

An extra lookup would be added during settings processing for a settings object with the key/folder name of
`application-foo-backend`

The order of the key can also be changed to accommodate partial Matching

### DIFF
--- a/engine/component.ftl
+++ b/engine/component.ftl
@@ -83,6 +83,7 @@
         (resourceGroup == DEFAULT_RESOURCE_GROUP) ]
         [#local extendedAttributes = [] ]
         [#local profileAttribute = coreProfileChildConfiguration[0] ]
+        [#local settingsNamespacesAttribute = coreSettingsNamespacesConfiguration[0]]
         [#list attributes as attribute ]
             [#if asArray(attribute.Names!attribute.Name)?seq_contains("Profiles")]
                 [#local profileAttribute +=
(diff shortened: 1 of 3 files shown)


ml019:
Nice idea. I've been refactoring the settings as part of the switch to contexts so will need to factor this in but the principle seems sounds. I was thinking about setting groups which is kind of the same thing.

PR #971, opened 2019-08-31
<fix> combineEntities array merging

roleyfoley:
Not sure if this is quite the right approach. But when adding content to a cli/JSON based output the mergeObjects uses the append behaviour which makes sense for objects, however when appending an object that contains an array this results in the array duplicating its values each time mergeObjects is called 

This PR makes the array behaviour separate from the object behaviour in combining entities to allow for objects to be merged and their child arrays kept with unique values


ml019:
@roleyfoley think we close this one for now?

PR #974, opened 2019-09-04
moved unused attributes from Shared to aws provider

RossMurr4y:
Drafting because I think we should talk about this - I suspect this is going to happen a lot, where the vast majority of attributes wont be shared. 

I was thinking that shared\id.ftl could contain a skeleton of high-level attribute headings, which we could then fill in within each provider. Some of these we already have - Notifications, Lifecycle etc. as well as some like Links. This would provide a logical place to then add into in provider\component\id.ftl, whilst not really requiring a debate about what to call every single attribute or shoe-horning an existing shared attribute into a place it doesn't fit. 

Thoughts?

### DIFF
--- a/providers/aws/components/s3/id.ftl
+++ b/providers/aws/components/s3/id.ftl
@@ -1,7 +1,81 @@
 [#ftl]
 [@addResourceGroupInformation
     type=S3_COMPONENT_TYPE
-    attributes=[]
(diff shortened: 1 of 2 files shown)


RossMurr4y:
Cancelling - think this needs to be looked at after we have the first additional provider tested and working.

Issue #998, opened 2019-09-26
Move reference data into dedicated feature


roleyfoley:
Currently we use the master data file to add extra data which can be shared across components, this includes IP Address Groups, CORS Profile, log file profiles etc. 

This data should be available through a standard process which can be used across providers/frameworks and also allows us to document which components use the reference data. 

How the data is used is up to the component designer but querying it should be a generic service.

roleyfoley:
Reference data is not validated and supports different input providers

PR #999, opened 2019-09-26
<feature> Reference Data Validation

roleyfoley:
Part of #998 

Adds support for dynamic loading and validation of Reference Data/Configuration which is not Component. 

Not sure if reference data is the best name for it, but the majority of this data is referenced by a component to determine shared configuration. 

This initial commit implements this through the setContext.ftl however I think the next stage in this would be to merge this into occurrence building and actually specify required ReferenceTypes on each component which would then be loaded dynamically and also provide a bit more context on what uses each set of Reference Data


ml019:
I think these are what I[ve been thinking about as configuration objects. Key aim is to be able to validate these using the same mechanism we do currently for component configuration?

roleyfoley:
> I think these are what I[ve been thinking about as configuration objects. Key aim is to be able to validate these using the same mechanism we do currently for component configuration?

Exunclty, the main goal was for documentation and validation purposes. The next step would be to move the reference data retrieval process into components so that we document the reference data that is use and also make it easier to manage the dependencies

roleyfoley:
@ml019 Do you want any changes to this PR or is it right in the naming and setup?

PR #1002, opened 2019-09-27
Allow for loading of a plugin utility script.

RossMurr4y:
As a part of the Azure plugin baseline component, I need to utilise the pki/ssh utilities within the generation\aws\utilities.sh - however in their current state they are specific to AWS.

Rather than update them to allow for a new provider (which would then have to happen for every possible plugin) this will allow dynamic loading of a second utility.sh that is specific to the provider. The "shared" provider utility.sh will load first and provide the bulk of the functions whilst anything provider specific can come from the plugin directly.

One thing to note is that plugin function names should include their provider, both for debugging (knowing where the function originated) and also to avoid naming conflicts between both utility scripts.

### DIFF
--- a/aws/common.sh
+++ b/aws/common.sh
@@ -7,6 +7,11 @@
 . ${GENERATION_DIR}/utility.sh
 . ${GENERATION_DIR}/contextTree.sh
 
+# Load plugin utilities if they exist
+if [[ -f ${GENERATION_PLUGIN_DIR}/utility.sh]]; then
+  . ${GENERATION_PLUGIN_DIR}/utility.sh
+fi
+
 function getLogLevel() {
   checkLogLevel "${GENERATION_LOG_LEVEL}"
 }


ml019:
Hi @RossMurr4y. think we need to think about this more generally as a refactoring of the way we handle script stuff in general - an area we really haven;'t tackled yet - thanks for volunteering!

Anything generic should be under the shared provider, while things provider specific should be in the same subdir under the provider.

Thinking maybe a /script subdirectory under each provider for this?

RossMurr4y:
@ml019 I agree wholeheartedly - I didn't make a script directory because I was mimicing the same structure of the AWS provider, but do like to have the files all tucked away somewhere. Ideally the root of the plugin is kept pretty organised. 

I think we should also have a directory in there to make it clear which part of CodeOnTap the script belongs to. In this case I need to add functionality into generation, but what if my plugin also needs to extend Automation for something like Azure Pipelines?

How would you feel about the following plugin structure?

```
${PLUGINS_DIR}/azure/
   automation/
     scripts/
        example.sh
   generation/
     components/
     deploymentframeworks/
     scripts/
        utility.sh
     services/ 
```

or if you just wanted a single scripts directory:

```
${PLUGINS_DIR}/azure/
components/
deploymentframeworks/
services/ 
scripts/
    automation/
        example.sh
    generation/
        utility.sh
```

The former may create some unnecessary folder structure for only one or two scripts, however I do like it over the second because it mimics the structure of the whole of CodeOnTap.

roleyfoley:
Closing based on feedback from https://github.com/codeontap/gen3/pull/1267

PR #1009, opened 2019-10-08
Minor fixes for scaling policy and task schedule

kshychko:
The PR contains two minor fixes:
1) The correct default value for scaling policy type
2) Event Rule as a part of CF template for Fargate ESC task.

### DIFF
--- a/providers/aws/components/ecs/setup.ftl
+++ b/providers/aws/components/ecs/setup.ftl
@@ -1091,11 +1091,15 @@
 
                 [#local ruleCleanupScript = []]
                 [#local ruleCleanupOutput = {}]
+                [#local cliCleanUpRequired = false]
 
                 [#list solution.Schedules?values as schedule ]
 
                     [#local scheduleRuleId = resources["schedules"][schedule.Id]["schedule"].Id ]
-                    [#local cliCleanUpScheduleRuleId = getExistingReference(scheduleRuleId, "cleanup") ]
+                    [#local cliCleanUpRequired = cliCleanUpRequired?then(
+                                cliCleanUpRequired,
+                                getExistingReference(scheduleRuleId, "cleanup")?has_content
+                    )]
 
                     [#local scheduleEnabled = hibernate?then(
                                 false,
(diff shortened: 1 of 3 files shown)


providers/aws/components/ecs/setup.ftl
```diff
@@ -1124,129 +1124,43 @@
...
-                                                    "Subnets" : subnets,
-                                                    "AssignPublicIp" : publicRouteTable?then("ENABLE", "DISABLED")
-                                                }
-                                            }
-                                        ) +
-                                        attributeIfTrue(
-                                            "LaunchType",
-                                            engine == "fargate",
-                                            "FARGATE"
-                                        )
```
roleyfoley: I think this needs to be moved over to
 https://github.com/codeontap/gen3/pull/1009/files#diff-48e25ae413e06c838128cd60d543fcdfR1147 

The network configuration still needs to be applied for fargate and awsvpc

providers/aws/components/ecs/setup.ftl
```diff
@@ -1124,129 +1124,43 @@
...
+                                    " case $\{STACK_OPERATION} in",
+                                    "   delete)",
+                                    "       # Manage Scheduled Event",
+                                    "       info \"Deleting Scheduled Task...\"",
+                                    "       delete_cloudwatch_event" +
+                                    "       \"" + region + "\" " +
+                                    "       \"" + scheduleRuleId + "\" " +
+                                    "       \"true\" || return $?",
+                                    "       ;;",
+                                    " esac"
```
roleyfoley: I think this needs to be a little bit different. To remove the existing cli created events we need to clean them up on the first deployment of the cloudformation based events, otherwise we run the risk of the scheduled task being executed twice or at random times. 

So I think we need to actually make this STACK_OPERATION `update|create` and add a pseduo stack output so that we can check and see if the cleanup has been completed. This makes sure that we don't accidentally delete the cloudwatch schedules created by cloudformation 

An example of this cleanup process is available in 
https://github.com/codeontap/gen3/blob/fece5bf524683e50f007aaf9982383d11b4c55d1/providers/aws/components/lb/setup.ftl#L465

kshychko: @roleyfoley I added schedules to resources and added epilogue script for clean up, hope it's correct now.

providers/aws/components/ecs/setup.ftl
kshychko: @roleyfoley , fixed

kshychko:
@roleyfoley , I force-pushed my branch, please review the updates

providers/aws/components/ecs/setup.ftl
```diff
@@ -1089,166 +1089,88 @@
...
-
-                            [@createScheduleEventRule
-                                id=scheduleRuleId
-                                enabled=scheduleEnabled
-                                scheduleExpression=schedule.Expression
-                                targetParameters=targetParameters
-                                dependencies=fnId
-                            /]
-                        [/#if]
+                    [#if cliCleanUpScheduleRuleId?has_content ]
```
roleyfoley: Should that be if it doesn't have content? I.e. we haven't cleaned it up before? 
Also it will need an additional condition to see if the task already exists. Otherwise on a new task deployment the rule could end up removing the new scheduled event.

kshychko: @roleyfoley `cliCleanUpScheduleRuleId` has content when we set the pseudo stack output variable with the "cleanup" postfix - https://github.com/codeontap/gen3/pull/1009/files/04b8881d936289bb2ef79bb30cb4ec1eedd7c512#diff-48e25ae413e06c838128cd60d543fcdfR1098
then we re-generate the pseudo stack output file and set this variable to the empty string. If delete fails, the stack output file will not be re-generated.

kshychko:
@roleyfoley the cleanup process logic is fixed now

PR #1010, opened 2019-10-09
<feature> Dyanmic Load - Settings, Masterdata, Blueprint, Definitions

roleyfoley:
Following on from #999  and #1007  this applies the same style of dataloading to settingsObject ,blueprintObject, definitionsObject and masterData inputdata. 

Rather than being required like stackoutputs are these imports are an optional data set, so instead of having to load data, input sources and providers can choose to seed data into the these objects. The exception is masterData which can only be seeded at the provider level which ensures that the master data is consistent across input sources, given that masterData cannot be user controlled. 

The standard getCompositeObject validation process doesn't quite fit with these ones, so it has been left out for now


roleyfoley:
Following on from the Dyanmic loading of data this PR also includes the initial implementation of scenarios. Scenarios provide a way to inject intput data using the engine itself. 

Scenarios can add 3 different inputs 

- blueprint - Actual CMDB content 
- settingsSets - which are an array of settings which can specify the scope, type and namspace of the settings to add 
- stackoutputs - which can inject stack outputs 

Combined with the mock data provider added as part of #1007 this allows you to generate templates without having any data specified in the CMDB. 

Due to the way that createTemplate.sh works at the moment you still need to be in a CMDB folder structure but the composite data provided by this will be ignored.

roleyfoley:
An example of a scenario has been added in `providers/aws/scenarios/lbhttps.ftl` 

In this example running the following commands would generate a template for the `https-lb` component without it being defined in the CMDB 

```
export GENERATION_INPUT_SOURCE=mock
export GENERATION_SCENARIOS=lbhttps
${GENERATION_DIR}/createTemplate.sh -l solution -u https-lb
``` 

The first command changes the input source from the composite input files to the mock input provider 
The second command tells the engine to find a scenario named `lbhttps` in the provider directories 

Then during the template process this scenario will be loaded and inserted into the blueprint

PR #1020, opened 2019-10-15
Fix multiprovider loading

RossMurr4y:
This PR implements the following: 
- Incorporate provider/framework options into getOpts as per features implimented in #1017
- As an extension to #1017, several function calls in setContext.ftl rely on functions defined in the AWS provider, which are then not loaded if you're testing another provider. Implemented temporary if-statements to only run those variable assignments if the provider is AWS.*

* - these if-statements will be removed as such functionality is made provider independent and/or moved into the AWS provider.


aws/createTemplate.sh
```diff
@@ -57,7 +57,7 @@ EOF
 function options() {
 
   # Parse options
-  while getopts ":c:d:g:hl:q:r:u:z:" option; do
+  while getopts ":c:d:g:hl:q:r:u:z:p:f:" option; do
```
roleyfoley: might need to rebase your PR. Looks like some other changes have leaked in

PR #1028, opened 2019-10-16
update RDS CA epilogue script

kshychko:
This change adds an extra script to RDS epilogue setup to update CA identifier for existing and new (created before the 1st of November) RDS instances.

### DIFF
--- a/providers/shared/references/Port/id.ftl
+++ b/providers/shared/references/Port/id.ftl
@@ -1,6 +1,6 @@
 [#ftl]
 
-[@addReference 
+[@addReference
     type=PORT_REFERENCE_TYPE
     pluralType="Ports"
     properties=[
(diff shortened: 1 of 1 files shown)


providers/aws/components/db/setup.ftl
```diff
@@ -505,6 +507,27 @@
                         "reset_master_password || return $?"
                     ],
                 []) +
+                (!rdsCA?has_content)?then(
+                    [
+                        "# Update RDS CA",
+                        "function update_rds_ca() {",
+                        "info \"Updating RDS CA ... \"",
+                        "rdsCAIdentifier=\"rds-ca-2019\"",
```
roleyfoley: Can we make this a component configuration value and add the script if the saved psedudo stack is different from the one specified in the configuration? 

CA's always have an expiry so it might be nice to make this configurable if we are going to add it

roleyfoley:
Added a couple of updates to move the CA over to the Security Profile. This allows for the configuration to updated in staged rollouts or big bang across a whole tenant

PR #1029, opened 2019-10-16
corrected an unusual scneario.

RossMurr4y:


roleyfoley:
Fingers need spellcheckers

Issue #1027, opened 2019-10-16
Testing - Add testconfig output


roleyfoley:
Epic #994 

Following on from #1026  we need a new JSON based output which provides the test config output. This would need to be generated as part of a template generation process 

The source of the data is from the test case itself and won't be the result of a component output

roleyfoley:
This has been implemented as testcase outputs

PR #1030, opened 2019-10-17
<feature> testplan generation

roleyfoley:
#1027 
Adds support for the generation of testplans based on testcases 
This allows for external testing platforms to receive a set of assertions and test cases and to use those assertions against the template generated using one or more scenarios defined in the test case. 

- The deployment unit is defined in the testcase to allow for scenarios with multiple deployment units defined due to dependencies but still defining which template output you want to test 

- Command line options are now their own input type which allows for any command line option to be overridden for both mocking and test cases. 

- The testplan is a new output type which is generated if a testcase has been specified. It is automatically added to the generation plan since we can't add it at the component level ( testcases are blueprint level not component ) 

- The composite validation functionality has been moved up into common from component so that it can be used across other input sources 

- Removed the product level templates completely as they are no longer supported 

- Updated setContext.sh so that it only assembles the composites if the `GENERATION_INPUT_SOURCE` is `composite` - This is part way to making createTemplate.sh usable without actually having a CMDB 

- masterData is now treated like the other inputsources and can be loaded for each inputsource. A shared inputsource is also available so that the masterData for a provider is available when all other input sources are used. As a result it has now moved in to the inputsources directory


engine/provider.ftl
```diff
@@ -110,6 +110,39 @@
     [/#list]
 [/#macro]
 
+[#macro includeTestCaseConfiguration provider testCase ]
+    [#if isConfigurationIncluded([provider, testCase]) ]
+        [#return]
+    [/#if]
+
+    [#local templates = []]
+    [#list ["testcase"] as level ]
```
RossMurr4y: ```"testcase"```
I can't find conclusive advice anywhere as to whether this will work without camel-case or not (as it was defined). Probably works fine, but bringing this to your attention.

RossMurr4y: Same in the list below also.

roleyfoley: the convention for template and macro dynamic loading seems to be lowercase based on the component dynamic loading. I think there are limits to the characters you can use in function/macro names

roleyfoley:
@ml019 @RossMurr4y  I think this PR has enough in it now :) 

Please review when you can

engine/bootstrap.ftl
```diff
@@ -24,40 +41,79 @@
...
+                                settings?eval,
+                                {}
+            ),
+            "Definitions" : (definitions!"")?has_content?then(
+                                definitions?eval,
+                                {}
+            ),
+            "StackOutputs" : (stackOutputs!"")?has_content?then(
+                                stackOutputs?eval,
+                                {}
```
ml019: Should be an array if empty?

engine/bootstrap.ftl
```diff
@@ -83,35 +140,89 @@
...
 
-[#-- Build the base state of the blueprint using the seeded master data --]
-[@addBlueprint blueprint=getMasterData(SHARED_PROVIDER) /]
+[#-- start the blueprint with the masterData --]
+[@addBlueprint blueprint=getMasterData() /]
 
-[#-- Include any command line provider/input source --]
-[#if commandLineOptions.Deployment.Provider.Name?has_content ]
-    [@addBlueprint blueprint=getMasterData(commandLineOptions.Deployment.Provider.Name) /]
+[#-- Set the scnearios provided via the CLI --]
```
ml019: Typo

engine/bootstrap.ftl
roleyfoley: Fixed

engine/bootstrap.ftl
roleyfoley: fixed

PR #1032, opened 2019-10-18
<feature> distinctInstance placement constraint for ECS Tasks

kshychko:
Set a placement constraint with type `distinctInstance` to a ECS services when MultiAZ is enabled so each task is running on a different container instance.

### DIFF
--- a/providers/aws/services/ecs/resource.ftl
+++ b/providers/aws/services/ecs/resource.ftl
@@ -284,6 +284,15 @@
             dependencies=""
     ]
 
+    [#-- define an array of constraints --]
+    [#-- for potential support of "memberOf" type placement constraint --]
+    [#local placementConstraints = [] ]
+    [#if placement.DistinctInstance]
+        [#local placementConstraints += [{
+            "Type" : "distinctInstance"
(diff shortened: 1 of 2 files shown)


kshychko:
> I think we need to make this a solution level configuration option. In order for it to work the service has to have a desired count lower or equal to the underlying host count, otherwise ECS won't be able to place the rest of the tasks.
> 
> This isn't an issue in most of our current deployments but as we start to scale up applications it could become an issue

@roleyfoley , I added a boolean property named  `DistinctInstance` to the `Placement` object of ECS Service. If needed we can support the `memberOf` type of placement constraints as well.

roleyfoley:
Thanks @kshychko Looks good. 
Is it still based on MultiAZ being enabled as well? or just the configuration option? I'm happy with that since the default desired count is 1 so it will still work, just want to make sure that was your intention

kshychko:
@roleyfoley, it's just a configuration option with a default value set to `true`, which works for both MultiAZ enabled and disabled environments, and only if we need to override it we set it to false in the solution file on the ECS Service task definition level.

PR #1041, opened 2019-10-22
migrated shared resources functions to the shared provider.

RossMurr4y:
This PR addresses the need for any provider to be able to check to see if a given resource is apart of the current deployment. Previously these functions were within the AWS provider, however with a small change to use a provider lookup instead of a hardcoded value, this can be used for any provider.

### DIFF
--- a/providers/shared/services/resource.ftl
+++ b/providers/shared/services/resource.ftl
@@ -1,0 +1,30 @@
+[#ftl]
+
+[#-- Is a resource part of a deployment unit --]
+[#function isPartOfDeploymentUnit resourceId deploymentUnit deploymentUnitSubset]
+  [#local resourceObject = getStackOutputObject(commandLineOptions.Deployment.Provider, resourceId)]
+  [#local 
+    currentDeploymentUnit =
+      deploymentUnit +
+      deploymentUnitSubset?has_content?then(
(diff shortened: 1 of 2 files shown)


RossMurr4y:
I have opted to place these functions within a new shared/services/resources file because I felt the purpose of the functions justified it. Please let me know if you feel theres a more suitable location for the functions.

providers/shared/services/resource.ftl
```diff
@@ -0,0 +1,30 @@
+[#ftl]
+
+[#-- Is a resource part of a deployment unit --]
+[#function isPartOfDeploymentUnit resourceId deploymentUnit deploymentUnitSubset]
+  [#local resourceObject = getStackOutputObject(commandLineOptions.Deployment.Provider, resourceId)]
```
roleyfoley: I think this needs to be `commandLineOptions.Deployment.Provider.Name` 

https://github.com/codeontap/gen3/blob/8580ab5da9e7313d0e78762cad9ca2b7414eb356/engine/bootstrap.ftl#L8

RossMurr4y: Oddly, this was working for me without the specificity, however I recognise that was undesirable.

I have corrected this in c9d516b

PR #1065, opened 2019-11-05
<feature> Lambda Create new version on deploy

roleyfoley:
When using Lambda@Edge you can only use versions of lambda functions. You can't use the $LATEST version which is normally implied 

This PR adds a configuration option on Lambda functions which creates a new version of a lambda function on each deployment. This is useful when you have fragment based lambda functions which have been updated


roleyfoley:
@kshychko Would you mind having a look at this if you have time? Just want to get someone to review it and see if it makes sense

PR #1069, opened 2019-11-08
<refactor> Remove redundant or unused scripts

roleyfoley:
Cleanup chore

- Removes add* scripts which have now been replaced with the cookie cutter patterns approach
- Removes the create*Level* templates which provided a minimal shortcut to adding a parameter
- Removes some gen2-gen3 conversion scripts


roleyfoley:
@ml019 can you please review this one. Just want to make sure I haven't removed something i shouldn't hav

ml019:
All look lik candidates for removal. Could be some weird edge case thing where one is needed but I think we just deal with it, particularly with the work on the CLI happening.

PR #1075, opened 2019-11-12
<refactor> Move output file naming into engine

roleyfoley:
Moves the output file suffix into the engine rather than createTemplate 

This is part of moving output definitions into the engine completely rather than half in the engine half in script


roleyfoley:
@ml019 @RossMurr4y This is now ready for review

RossMurr4y:
Looks good to me and working as expected with the Azure provider.

providers/aws/inputsources/mock/commandlineoption.ftl
```diff
@@ -2,14 +2,17 @@
...
 [#macro aws_input_mock_commandlineoption_seed ]
+
     [@addCommandLineOption
         option={
             "Regions" : {
                 "Segment" : "ap-southeast-2",
-                "Product" : "ap-southeast-2",
                 "Account" : "ap-southeast-2"
+            },
+            "References" : {
```
ml019: Is this test data?

roleyfoley: Yep that's the mock data input

PR #1081, opened 2019-11-20
<feature> SNS topic Subscriptions for AWS services

roleyfoley:
Adds support for Lambda and SQS to be used for topic subscriptions

### DIFF
--- a/providers/aws/components/topic/setup.ftl
+++ b/providers/aws/components/topic/setup.ftl
@@ -72,6 +72,8 @@
         [#local solution = subOccurrence.Configuration.Solution ]
         [#local resources = subOccurrence.State.Resources ]
 
+        [#local subscriptionDependencies = []]
+
         [#switch core.Type]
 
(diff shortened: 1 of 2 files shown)


providers/aws/components/topic/setup.ftl
```diff
@@ -112,6 +114,37 @@
...
+                                    formatDependentPolicyId(
+                                        subscriptionId,
+                                        resourceId) ]
+
+                            [#local subscriptionDependencies += [policyId] ]
+
+                            [@createSQSPolicy
+                                id=policyId
+                                queues=resourceId
+                                statements=sqsWritePermission(
```
ml019: I think the changes to the parameters on the sqsWritePermission don't match this call

roleyfoley: What did you think was missing? Seems to have the three parameters now required in the update

ml019: Thought sqswrite permission had four params not three.

PR #1082, opened 2019-11-24
<refactor> Subnet Calculation with Java Library

roleyfoley:
In https://github.com/codeontap/gen3-freemarker-wrapper/pull/3 a new function was added to the freemarker wrapper which provides a way to calculate the potential subnets available in a network based on a given CIDR mask 

We never used it, but now that we have different network requirements for providers ( azure vs AWS ) we need a more flexible way to manage the subnet allocations. 

This PR moves to using this function along with some extra functions used to determine the appropriate subnet mask based on the zone and tier count ( for AWS, any number will actually work). 

In this setup, for AWS, any number of zones and tiers are acceptable, odd number counts will be rounded to the closest, highest even number to match subnet masking setups and so we can get an even square root. 

Also handles variable Network CIDR masks without a problem, and could potentially support IPv6 as well.

**Note** This PR depends on an update to the freemarker wrapper which handles the java call more efficiently ( @kshychko  is working on that ) 

@RossMurr4y This should be what you need to get Azure networking going


kshychko:
@roleyfoley I added update to 1.8.2 freemaerker-wrapper version to your PR

ml019:
@kshychko presume the version number won;'t interfere with th work we did on extending the wrapper to do directory scans etc?

PR #1084, opened 2019-11-24
<fix> APIGateway Occurrence Naming

roleyfoley:
With the ability to not include the stage in custom Domain names on API Gateways the stage might not been seen on an APIGateway

This gets around an issue at the moment where you have to set a version on an APIGateway to get the stage name to be populated.


ml019:
Intent was that we would always verison the API internally via the version even if not exposed because of the domain mappings.

Have we got a use case where this doesn't apply?

roleyfoley:
I was going more for a consistency approach, on every other component the version and instance is defined as they see fit by a codeontap user 

Here it is defined by the component designer and it has tripped me up a few times when doing a deployment 

The alternative to this is raise an exception if version and instance  aren’t provided

ml019:
@roleyfoley ok. Convention is to include the version but having a fallback seems reasonable.

PR #1085, opened 2019-11-25
<fix> check if lifecycle is enabled when adding an expiration rule

kshychko:
to enable versioning without enabling expiration rules


roleyfoley:
Great Catch!

PR #1086, opened 2019-11-26
<refactor> Split client bash scripts from execution scripts

roleyfoley:
With our move to the gen3 cli we need to cleanup the layout of the scripts to make them a bit more understandable to the development team and so we can see what needs to be worked on. 

This PR introduces a distinction between two types of scripts we have 

- client/frontend - Scripts which would either be called by a user of codeontap or the automation scripts 
- backend/library - Scripts which are called by the clients and should not be called directly by users 

Also removes some files which don't appear to be in use anymore


ml019:
@roleyfoley is it worth defining a new GENERATION_EXECUTION_DIR or equivalent? Same goes for the other type of script?

roleyfoley:
> @roleyfoley is it worth defining a new GENERATION_EXECUTION_DIR or equivalent? Same goes for the other type of script?

Adding one in would make it a breaking change. I wasn't sure if it was quite enough to justify a breaking change

ml019:
@roleyfoley OK - can merge it in. Whole env thing may change as we process the cli so potentially becomes a non-issue.

Would you mind rebasing to fix conflicts?

roleyfoley:
@ml019 done

ml019:
@roleyfoley Missed the update :-( If you rinse and repeat, I promise I'll merge it asap.

roleyfoley:
@ml019 Rebase done

PR #1087, opened 2019-11-28
<feature> Create test case templates without a cmdb

roleyfoley:
This allows for a testing create template pass to generate outputs without have a CMDB
All configuration is managed through the test case, scnearion and mocking input configuration

Also allows you to override the output directory so that you know where the outputs end up when running in this mode 

Example  ( from any directory as long as you have the gen3 repo cloned and setup with the required environment variables) 
```
${GENERATION_DIR}/createTemplate.sh -o /root/testout -t testlbhttps -i mock -l solution
``` 

Will create the load balancer templates for the testlbhttps testcase ( https://github.com/codeontap/gen3/blob/7d94833e14227795abcca10d6e2606908de1a37a/providers/aws/testcases/testlbhttps.ftl ) and output them to /root/testout 

Input source and level are still required though which we can tweak to meet requirements

### DIFF
--- a/aws/createTemplate.sh
+++ b/aws/createTemplate.sh
@@ -11,6 +11,7 @@ DEPLOYMENT_MODE_DEFAULT="update"
 GENERATION_PROVIDER_DEFAULT="aws"
 GENERATION_FRAMEWORK_DEFAULT="cf"
 GENERATION_INPUT_SOURCE_DEFAULT="composite"
+GENERATION_TESTING_DEFAULT="false"
 
 function usage() {
   cat <<EOF
@@ -26,6 +27,7 @@ where
 (o) -i GENERATION_INPUT_SOURCE is the source of input data to use when generating the template
     -h                         shows this text
 (m) -l LEVEL                   is the template level - "blueprint", "account", "segment", "solution" or "application"
+(o) -o OUTPUT_DIR              is the directory where the outputs will be saved - defaults to the PRODUCT_INFRASTRUCTURE_DIR
 (o) -q REQUEST_REFERENCE       is an opaque value to link this template to a triggering request management system
 (o) -r REGION                  is the AWS region identifier
 (m) -u DEPLOYMENT_UNIT         is the deployment unit to be included in the template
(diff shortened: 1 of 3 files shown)


roleyfoley:
@kshychko @RossMurr4y Would you mind reviewing this one?

aws/createTemplate.sh
```diff
@@ -89,27 +94,36 @@ function options() {
   GENERATION_PROVIDER="${GENERATION_PROVIDER:-${GENERATION_PROVIDER_DEFAULT}}"
   GENERATION_FRAMEWORK="${GENERATION_FRAMEWORK:-${GENERATION_FRAMEWORK_DEFAULT}}"
   GENERATION_INPUT_SOURCE="${GENERATION_INPUT_SOURCE:-${GENERATION_INPUT_SOURCE_DEFAULT}}"
+  GENERATION_TESTING="${GENERATION_TESTING:-${GENERATION_TESTING_DEFAULT}}"
 
-  # Check level and deployment unit
-  ! isValidUnit "${LEVEL}" "${DEPLOYMENT_UNIT}" && fatal "Deployment unit/level not valid" && return 1
+  # Skip context generation when testing
+  if [[ "${GENERATION_TESTING}" == "false" ]]; then
```
kshychko: Will it work if we just check if `GENERATION_TESTCASE` has been passed here? `GENERATION_TESTING` is introduced but isn't allowed to be set it explicitly. Not a big deal, just an extra variable.

aws/createTemplate.sh
```diff
@@ -50,7 +52,6 @@ GENERATION_INPUT_SOURCE = "${GENRATION_INPUT_SOURCE_DEFAULT}"
 NOTES:
 
 1. You must be in the directory specific to the level
-2. REGION is only relevant for the "product" level
 3. DEPLOYMENT_UNIT must be one of "s3", "cert", "roles", "apigateway" or "waf" for the "account" level
```
kshychko: @roleyfoley , could you please update the numbering below?

roleyfoley: Fixed up

aws/createTemplate.sh
roleyfoley: Removed testcase and swapped it to enable on GENERATION_INPUT_SOURCE - that's more what its controlling anyway

PR #1088, opened 2019-12-02
<feature> Add Aurora cluster support

roleyfoley:
Adds support for AWS Aurora as a database engine 

Aurora runs as a cluster of db instances with storage managed by the cluster resource and the instances used to manage the compute resources. You can deploy up to 16 database instances across all AZs in a region 

Support has been setup in this PR for postgres based installations, mysql is available as well but has a long list of extra options that can be configured 

Most of the configuration is managed behind the scenes within the framework with the following configuration options added 

- new engine - aurora-postgresql
- An extra configuration section for cluster level database parameters 
- Desired, Max and Min counts on the processor component - Desired is in the initial release, Min and Max can be added for autoscaling the db instances

### DIFF
--- a/aws/utility.sh
+++ b/aws/utility.sh
@@ -1543,111 +1543,226 @@ function delete_oai_credentials() {
 }
 
 # -- RDS --
+
+function add_tag_rds_resource() {
+  local region="$1"; shift
+  local rds_identifier="$1"; shift
+  local key="${1}"; shift
+  local value="${1}"; shift
+
+  aws --region "${region}" rds add-tags-to-resource --resource-name "${rds_identifier}" --tags "Key=${key},Value=${value}" || return $?
+
+}
+
 function create_snapshot() {
   local region="$1"; shift
+  local db_type="$1"; shift
   local db_identifier="$1"; shift
   local db_snapshot_identifier="$1"; shift
 
   # Check that the database exists
-  db_info=$(aws --region "${region}" rds describe-db-instances --db-instance-identifier ${db_identifier} )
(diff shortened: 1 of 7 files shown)


aws/utility.sh
```diff
@@ -1543,111 +1543,226 @@ function delete_oai_credentials() {
...
+    if [[ -n "${db_info}" ]]; then
+      aws --region "${region}" rds create-db-cluster-snapshot --db-cluster-snapshot-identifier "${db_snapshot_identifier}"  --db-cluster-identifier "${db_identifier}" 1> /dev/null || return $?
+    else
+      info "Could not find db ${db_identifier} - Skipping pre-deploy snapshot"
+      return 0
+    fi
 
-  if [[ -n "${db_info}" ]]; then
-    aws --region "${region}" rds create-db-snapshot --db-snapshot-identifier "${db_snapshot_identifier}" --db-instance-identifier "${db_identifier}" 1> /dev/null || return $?
+    sleep 5s
```
RossMurr4y: could we use here another api call rather than an explicit sleep?
```
aws rds wait db-snapshot-completed ...
```
or will this not register cluster snapshots?

roleyfoley: Its a lovely annoying thing with AWS. Submitting the create-db-snapshot command is asynchronous so there can be a delay ( around 2 seconds ) between running the command and the status changing to match what the wait command expects. 

We actually use the wait command after this sleep but in order for the wait command to work you need to sleep for a bit 🤦‍♂

PR #1101, opened 2019-12-06
<feature> ExternalService Fragment and Env control

roleyfoley:
Adds support for external services to use fragments
Which also adds support for _context environment control

Minor fix for apigateway region lookup


ml019:
@rolefoley Thanks for this. I'll test with ICS

PR #1102, opened 2019-12-07
<Feature> Add external tier for external services

ml019:
Need to tweak the external link processing to check there isn't actually an external tier. If there is, then assume all external links will be managed via components in the external tier of the solution.

The use of the external tier is now our preferred approach rather than external links. It makes the dependencies much more obvious and the layout of the settings more logical. (Sharing of settings could be done via namespaces but links are a lot more explicit). It also means any generated documentation automatically documents the external dependencies and the values expected for eac external dependency.

We will need to change any components that used to rely on setting the type of the link. They will now need to explicitly look for links to "externalservice" components and check that any required attributes are defined.

For now, if an external link is required even if an external tier is present, set a type on the link.

Also a couple of fixes
- typos for setting namespace management
- a correction to the API gateway logic to manage AWS regions in URLs.

### DIFF
--- a/providers/aws/inputsources/shared/masterdata.ftl
+++ b/providers/aws/inputsources/shared/masterdata.ftl
@@ -1381,7 +1381,8 @@
         "ana",
         "mgmt",
         "docs",
-        "gbl"
+        "gbl",
+        "external"
       ]
     }
   },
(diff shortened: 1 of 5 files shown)


providers/aws/inputsources/shared/masterdata.ftl
```diff
@@ -1381,7 +1381,8 @@
         "ana",
         "mgmt",
         "docs",
-        "gbl"
+        "gbl",
+        "external"
```
roleyfoley: Should the Id be ext rather the external?

PR #1103, opened 2019-12-07
<Feature> Permit escaping of characters in setting names

ml019:
Use the "^" character to escape any non-alphanumeric characters in an setting name that should not be converted to "_".

The most common use of this would be in settings added to create environment variables, since the setting name becomes to environment name.

### DIFF
--- a/engine/inputdata/setting.ftl
+++ b/engine/inputdata/setting.ftl
@@ -16,7 +16,15 @@
 [/#macro]
 
 [#function formatSettingName parts...]
-    [#return concatenate(parts, "_")?upper_case?replace("-", "_")]
+    [#-- First join the parts and force to upper case --]
+    [#local name = concatenate(parts, "_")?upper_case ]
+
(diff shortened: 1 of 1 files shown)


engine/inputdata/setting.ftl
```diff
@@ -16,7 +16,15 @@
 [/#macro]
 
 [#function formatSettingName parts...]
-    [#return concatenate(parts, "_")?upper_case?replace("-", "_")]
+    [#-- First join the parts and force to upper case --]
+    [#local name = concatenate(parts, "_")?upper_case ]
+
+    [#-- Use a lookbehind regex to permit any non-alphanumeric to be escaped by "^" --]
+    [#-- Double backslash is one for freemarker and one for regex --]
+    [#local name = name?replace("(?<!\\^)[^0-9A-Z\\^]", "_", "r")]
```
roleyfoley: Could use replaceAlphaNumericOnly function for that

PR #1104, opened 2019-12-08
<Feature> semver range checking

ml019:
Add a routine to determine whether a given semver value lies within a range of values.

Also add extra convenience methods for logging.

### DIFF
--- a/aws/utility.sh
+++ b/aws/utility.sh
@@ -93,12 +93,20 @@ function trace() {
   message "${LOG_LEVEL_TRACE}" "$@"
 }
 
-function info() {
+function information() {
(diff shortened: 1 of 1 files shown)


aws/utility.sh
```diff
@@ -93,14 +93,22 @@ function trace() {
...
 
+function info() {
+  information "$@"
+}
+
 function warning() {
   message "${LOG_LEVEL_WARNING}" "$@"
 }
 
+function warn() {
```
roleyfoley: Do we really need this? feels like we should just find and replace where we are using the wrong ones

PR #1107, opened 2019-12-08
<refactor> v2.0.0 restructure of CMDB

ml019:
Reorganise cmdb to make it easier to manage via branches and dynamic cmdbs

- State is now in its own directory at the same level as config and infrastructure
- Solutions is now under infrastructure
- Builds are separated from settings and are now under infrastructure
- Operations are now in their own directory at same level as config and infrastructure. For consistency with config, a settings subdirectory has been added.

With this layout,
- infrastructure should be the same across environments assuming no builds are being promoted
- product and operations settings are managed consistently
- all the state info is cleanly separated (so potentially in its own repo)

/config/settings
/operations/settings
/infrastructure/solutions
/infrastructure/builds
/state/cf
/state/cot

If config and infrastructure are not in the one repo, then the upgrade must be performed manually and the cmdb version manually updated.

Functions for context tree access now support the pre and post v2.0.0 cmdb directory formats.

Previously defined global variables have been redefined to permit support for pre and post v2.0.0 cmdbs.

NOTE: 
v2.0.0 has not been included in the update list, so switching to v2.0.0 will be done via a separate PR. For now, the aim is to test the changes do not cause issues for pre v2.0.0 cmdbs. Manual testing of v2.0.0 upgrades can also be performed.

### DIFF
--- a/aws/contextTree.sh
+++ b/aws/contextTree.sh
@@ -204,8 +204,8 @@ function assemble_settings() {
     [[ -z "${name}" ]] && name="${id}"
     [[ (-n "${ACCOUNT}") && ("${ACCOUNT,,}" != "${name,,}") ]] && continue
 
-    local account_dir="$(filePath "${account_file}")/settings"
-    debug "Processing ${account_dir} ..."
+    local account_dir="$(findGen3AccountSettingsDir "${root_dir}" "${name}")"
+    debug "Processing account dir ${account_dir} ..."
     pushd "${account_dir}" > /dev/null 2>&1 || continue
 
     tmp_file="$( getTempFile "account_settings_XXXXXX.json" "${tmp_dir}")"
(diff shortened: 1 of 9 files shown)


ml019:
@roleyfoley this is ready for review. Tested with pre and post cmdb 2.0.0 formats and builds work ok. Haven't tried deploying anything but will update HA with latest of this PR and the equivalent automation one once you have had a look.

roleyfoley:
Looks good, couldn't see any immediate danger

Should this now be our default output layout? and we should make our cookie cutter templates align? 

What happens with account/ tenant repositories and products in account repositories if account repos haven't been included

ml019:
Making v2 the default via cookie cutter seems reasonable. We can actually do the v2 upgrade like others in the past (the breaking change was going to be using branches but I think with planning that should be incrementally possible too - its just trickier to copy things between environments).

The logic handles lookups against account and product separately so they should be ok if not everything is upgraded. 

Once the support is in place, I can test a few of these edge cases manually based on your suggestions.

PR #1115, opened 2019-12-11
<refactor> Prepare for transition to v2.0.0 cmdb format

ml019:
Reorder the searches to favour the more common cmdb directory layouts.

Preference v2.0.0 as it will become the most common layout.

### DIFF
--- a/aws/contextTree.sh
+++ b/aws/contextTree.sh
@@ -445,8 +445,8 @@ function findGen3AccountInfrastructureDir() {
   local account="$1"; shift
 
   findDir "${root_dir}" \
-    "infrastructure/**/${account}" \
-    "${account}/infrastructure"
+    "${account}/infrastructure" \
+    "infrastructure/**/${account}"
 }
 
 # It would be very rare that accounts would have standard settings and operations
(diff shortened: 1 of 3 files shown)


roleyfoley:
Should this be part of the PR to enable the v2.0.0 upgrade rather than doing it before

ml019:
@roleyfoley - ok I can reorder to reflect current structure for now and add reorder to PR to switch to v2.0.0

ml019:
@roleyfoley tweaked things to permit parallel use of pre and post v2.0.0, with a favouring of pre v2.0.0 where possible. Ready for review

roleyfoley:
Can this just be part of the PR to enable the v2 cutover? Feels like we are adding something we don't need yet

ml019:
It is already added - this is just making it more optimised for current cmdb layout. Happy to repurpose this as the v2.0.0 upgrade PR.

ml019:
New scope 

 Add a mechanism to separate the list of available CMDB upgrade and cleanup versions from the version targetted by default.

This supports a few scenarios;

 - selective upgrade or cleanup of cmdbs ahead of a change to the default targetted value
 - pinning of the upgrade or cleanup CMDB version.

Two new environment variables can be set to control this behaviour;

GENERATION_MAX_CMDB_UPGRADE_VERSION
GENERATION_MAX_CMDB_CLEANUP_VERSION

Add v2.0.0 to the list of available upgrades, and optimise context directory searches assuming v2.0.0 is the default.

PR #1116, opened 2019-12-12
created VM Image reference type

RossMurr4y:
New reference type for creating profiles for virtual machine images.

New profiles can now be created as necessary in each providers masterData.ftl.

```
"VMImageProfiles" : {
  "bastion" : {
    "Publisher" : "Canonical",
    "Offering" : "UbuntuServer",
    "SKU" : "18.04-LTS"
  }
}
```

These profiles can then be referenced in component setup easily through the global var vmImageProfiles:

```
[#local vmssVMImageProfile = vmImageProfiles[BASTION_COMPONENT_TYPE]]
```

### DIFF
--- a/providers/shared/inputsources/shared/masterdata.ftl
+++ b/providers/shared/inputsources/shared/masterdata.ftl
@@ -984,7 +984,7 @@
             "Rules" : [ "All" ]
           }
         },
-        "OSProfiles": {},
+        "VMImageProfiles": {},
         "PlacementProfiles": {
           "external": {
             "default": {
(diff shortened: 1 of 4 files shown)


providers/shared/inputsources/shared/masterdata.ftl
```diff
@@ -984,6 +984,7 @@
             "Rules" : [ "All" ]
           }
         },
+        "OSProfiles": {},
```
roleyfoley: Does this need to be aligned with our name change?

RossMurr4y: Thanks for the catch - thought I'd already adjusted that. 4f4a8d6 corrects this.

PR #1119, opened 2019-12-13
<fix> move source maps uploads configuration to CMDB

kshychko:
This PR separates the configuration and process of source maps uploads.
Source maps upload process:

1. create a release using a build reference as a name (`SENTRY_RELEASE`)
2. download sources and map files from the specified path (`SENTRY_SOURCE_MAP_S3_URL`)
3. upload sources and maps files specifying a prefix (`SENTRY_URL_PREFIX`)
4. finalize the release 

Initially source maps uploads were added for expo builds, and all configuration was calculated at expo publish stage. For SPA there is only deploy, so the configuration needs to be read from the config file.  `CONFIG_BUCKET`/`CONFIG_FILE` already are added to a build blueprint for `mobileapp` and now for `spa` application type as well. `DEPLOYMENT_UNIT` is added as a mandatory argument to generate a build blueprint and read the configuration location from it. 

Order of lookup:

1. local environment
2. configuration file

This gives the flexibility either pass the settings from another build (mobileapp), define as Jenkins job environment variable or read it from the config.

For expo builds source map location depends on expo sdk version and OTA configuration, and as these settings are already accessed at the publish phase, `SENTRY_SOURCE_MAP_S3_URL` and `SENTRY_URL_PREFIX` are defined there as well and passed as chain properties to the sentry release job. 
`SENTRY_PROJECT`, `SENTRY_URL` and `SENTRY_ORG` are the default Sentry cli configuration variables.

### DIFF
--- a/aws/runExpoAppPublish.sh
+++ b/aws/runExpoAppPublish.sh
@@ -269,10 +269,8 @@ function main() {
   fi
 
   # variable for sentry source map upload
-  SENTRY_RELEASE_NAME="${BUILD_REFERENCE}"
   SENTRY_SOURCE_MAP_S3_URL="s3://${PUBLIC_BUCKET}/${PUBLIC_PREFIX}/packages/${EXPO_SDK_VERSION}"
   echo "SENTRY_SOURCE_MAP_S3_URL=${SENTRY_SOURCE_MAP_S3_URL}" >> ${AUTOMATION_DATA_DIR}/chain.properties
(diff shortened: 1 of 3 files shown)


kshychko:
@roleyfoley , I tested the changes on both mobileapp and SPA application types, worked well for me. Though they requires additional configuration in CMDB fragment files, this change makes it easy to control and adjust.

PR #1123, opened 2019-12-17
<feature> null cleaning json

roleyfoley:
Freemarker doesn't support null json values, this PR adds a manageJSON function which strips any null values and their keys from the JSON input


aws/nullClean.jq
```diff
@@ -0,0 +1,11 @@
...
+def walk(f):
+  . as $in
+  | if type == "object" then
+      reduce keys[] as $key
+        ( {}; . + { ($key):  ($in[$key] | walk(f)) } ) | f
+  elif type == "array" then map( walk(f) ) | f
+  else f
+  end;
+
+walk( if type == "object" then with_entries(select(.value != null)) else . end )
```
RossMurr4y: In testing I found this was stripping all the nulls except where the objects value was a null array - ```"test" : [null]```. Unfortunately this is how arrays that aren't specified are output from an ARM deployment.

```
{
  "locations": [
    null
  ],
  "resourceType": "storageAccounts/blobServices"
},
```

Amending to the following worked:
```
walk( if type == "object" then with_entries(select(.value != null and .value != [null])) else . end )
 ```

roleyfoley: Thanks moved this PR into #1124

PR #1136, opened 2019-12-20
<refactor> move genplan to a dedicated invoke macro

roleyfoley:
Moves the genplan to a dedicated component macro which is invoked dynamically 

This is inline with the idea of introducing other plans which don't need to load the setup routine to generate what they need

### DIFF
--- a/engine/component.ftl
+++ b/engine/component.ftl
@@ -203,7 +203,7 @@
             [@debug
                 message="Unable to invoke any of the macro options"
                 context=macroOptions
-                enabled=true
+                enabled=false
             /]
         [/#if]
     [/#if]
--- a/engine/component.ftl
+++ b/engine/component.ftl
@@ -167,19 +167,26 @@
     [#return profile[key]!{} ]
 [/#function]
 
-[#function invokeComponentMacro occurrence resourceGroup levels=[] parent={} baseState={} ]
+[#function invokeComponentMacro occurrence resourceGroup type qualifiers=[] parent={} baseState={} ]
     [#local placement = (occurrence.State.ResourceGroups[resourceGroup].Placement)!{} ]
     [#if placement?has_content]
         [#local macroOptions = [] ]
-            [#list asArray(levels) as level]
-                [#-- An empty level is assumed to be ignored --]
-                [#local macroOptions +=
-                    [
-                        [placement.Provider, occurrence.Core.Type, resourceGroup, placement.DeploymentFramework, level],
-                        [placement.Provider, occurrence.Core.Type, placement.DeploymentFramework, level],
-                        [placement.Provider, resourceGroup, placement.DeploymentFramework, level]
-                    ]]
-            [/#list]
+        [#list asArray(qualifiers) as qualifier]
+            [#-- An empty level is assumed to be ignored --]
+            [#local macroOptions +=
+                [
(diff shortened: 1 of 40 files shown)


engine/component.ftl
```diff
@@ -167,19 +167,26 @@
...
-                [#-- An empty level is assumed to be ignored --]
-                [#local macroOptions +=
-                    [
-                        [placement.Provider, occurrence.Core.Type, resourceGroup, placement.DeploymentFramework, level],
-                        [placement.Provider, occurrence.Core.Type, placement.DeploymentFramework, level],
-                        [placement.Provider, resourceGroup, placement.DeploymentFramework, level]
-                    ]]
-            [/#list]
+        [#list asArray(qualifiers) as qualifier]
+            [#-- An empty level is assumed to be ignored --]
```
ml019: Fix/remove comment

providers/aws/components/apigateway/setup.ftl
```diff
@@ -1,11 +1,11 @@
 [#ftl]
-[#macro aws_apigateway_cf_application occurrence ]
-    [@debug message="Entering" context=occurrence enabled=false /]
 
-    [#if deploymentSubsetRequired("genplan", false)]
-        [@addDefaultGenerationPlan subsets=["pregeneration", "prologue", "template", "epilogue", "config"] /]
-        [#return]
-    [/#if]
+[#macro aws_apigateway_cf_genplan_application ]
```
ml019: should accept occurrence in case needed in future

providers/shared/deploymentframeworks/default/legacy.ftl
```diff
@@ -39,15 +39,25 @@
...
                     [#list occurrence.State.ResourceGroups as key,value]
-                        [#if invokeSetupMacro(occurrence, key, ["setup", level]) ]
-                            [@debug
-                                message="Processing " + key + " ..."
-                                enabled=false
-                            /]
-                        [/#if]
+                        [#switch commandLineOptions.Deployment.Unit.Subset ]
+                            [#case "genplan" ]/]
+                                [#if invokeGenPlanMacro(occurrence, key, ["genplan", level] ) ]
```
ml019: should just be level as the qualifier?

providers/shared/deploymentframeworks/default/legacy.ftl
```diff
@@ -39,15 +39,25 @@
...
+                                    [@debug
+                                        message="GenPlan Processing key:" + key + "level:" + level + "..."
+                                        enabled=false
+                                    /]
+                                [/#if]
+                                [#break]
+
+                            [#default]
+                                [#if invokeSetupMacro(occurrence, key, ["setup", level]) ]
```
ml019: should just be level as the qualifier?

engine/component.ftl
roleyfoley: Fixed

providers/shared/deploymentframeworks/default/legacy.ftl
roleyfoley: Doh!

providers/shared/deploymentframeworks/default/legacy.ftl
roleyfoley: Doh!

PR #1140, opened 2019-12-20
<fix>pass engine to get_rds_url function

kshychko:
A minor fix for correct db URL generation

### DIFF
--- a/providers/aws/components/db/setup.ftl
+++ b/providers/aws/components/db/setup.ftl
@@ -591,7 +591,7 @@
                     [
                         "info \"Generating URL... \"",
                         "rds_url=\"$(get_rds_url" +
-                        " \"" + hostType + "\" " +
+                        " \"" + engine + "\" " +
                         " \"" + rdsUsername + "\" " +
                         " \"$\{master_password}\" " +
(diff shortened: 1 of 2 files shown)


providers/aws/components/db/setup.ftl
```diff
@@ -591,7 +591,7 @@
                     [
                         "info \"Generating URL... \"",
                         "rds_url=\"$(get_rds_url" +
-                        " \"" + hostType + "\" " +
+                        " \"" + engine + "\" " +
```
roleyfoley: I think that actually needs to be `scheme` instead of engine
For Aurora postgres we override the scheme to be `postgres://` when the engine is `aurora-postgres` 
I added both hostType and scheme and must have mised them up

roleyfoley: https://github.com/codeontap/gen3/blob/101794e493b3dffbad72d34962a556e86a0bb3a9/providers/aws/components/db/state.ftl#L12 for reference

kshychko: fixed and aligned get_rds_url

PR #1148, opened 2019-12-31
<fix> lb updates from lint testing

roleyfoley:
Some minor fixes to lb rules based on the testing results in #1145 

- Change the type of value for target groups - cfn-lint 
- Make the egress security rule an explicit rather than implicit - cfn_nag F1000


ml019:
Looks like we don't currently process affected settings via getCompositeObject(). If/when we do, can provide both string and number as acceptable types.

PR #1149, opened 2019-12-31
<feature> Deployment Unit list

roleyfoley:
Adds a new level deployment which will generate a list of all deployment units found a blueprint and return them as a list 

This is intended to be used as a way to locate deployment units and loop through them for template generation

### DIFF
--- a/aws/createTemplate.sh
+++ b/aws/createTemplate.sh
@@ -25,7 +25,7 @@ where
 (o) -g RESOURCE_GROUP          is the deployment unit resource group
 (o) -i GENERATION_INPUT_SOURCE is the source of input data to use when generating the template - "composite", "mock"
     -h                         shows this text
-(m) -l LEVEL                   is the template level - "blueprint", "account", "segment", "solution" or "application"
+(m) -l LEVEL                   is the template level - "unitlist", "blueprint", "account", "segment", "solution" or "application"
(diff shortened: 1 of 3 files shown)


ml019:
Do we need something to be able to group/order these?

roleyfoley:
For what I want it for ( test generatation ) no, but it probably would be useful

Are you thinking group by level?

RossMurr4y:
Is the intention here to use it where all deployment units are for the same provider and framework? I can see that being very useful. For multi-provider stacks, would they continue to use the current individual deployment creation process or would it make sense to be able to provide an override that excludes certain deployment units. 

```Ie: ${GENERATION_DIR}/createTemplate -p aws -f cf -l unitlist --exclude-unit "s3"```

roleyfoley:
The intention with deployment unit is that they are a collective deployment which can span multiple providers/frameworks  so in the future running this would list all the units across all providers/frameworks 

But at the moment with the current setup the unit list will return based on the provider framework that has been set and this is based on the components it can find in the blueprint 


The intention for this PR at the moment is to be able to generate all of the templates defined in the testing scenarios without knowing the deployment unit. This list will be looped through to generate the template and test cases for each of the tests defined

roleyfoley:
@ml019 Did you want me to change anything for this PR or happy with it as it is?

PR #1156, opened 2020-01-08
<feature> ES Improvements

roleyfoley:
A collection of updates for ES 
- Adds support for ES on VPC 
- Moves cognito from CLI to Cloudformation 
- Adds logging support to cloudwatch 
- Adds Processor support for controlling node counts with fixed counts 
- Version upgrade support for ES cluster - by default cloudformation replaces on version changes 

Updated processor profile reference to include specific coniguration for each component type which use processor profiles

### DIFF
--- a/providers/aws/components/db/setup.ftl
+++ b/providers/aws/components/db/setup.ftl
@@ -21,17 +21,13 @@
     [#local cmkKeyArn = getReference(cmkKeyId, ARN_ATTRIBUTE_TYPE)]
 
     [#local networkLink = getOccurrenceNetwork(occurrence).Link!{} ]
-
     [#local networkLinkTarget = getLinkTarget(occurrence, networkLink ) ]
-
     [#if ! networkLinkTarget?has_content ]
         [@fatal message="Network could not be found" context=networkLink /]
         [#return]
     [/#if]
-
     [#local networkConfiguration = networkLinkTarget.Configuration.Solution]
     [#local networkResources = networkLinkTarget.State.Resources ]
-
     [#local vpcId = networkResources["vpc"].Id ]
 
     [#local auroraCluster = false]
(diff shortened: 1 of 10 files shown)


providers/aws/components/es/setup.ftl
```diff
@@ -1,6 +1,6 @@
 [#ftl]
 [#macro aws_es_cf_genplan_solution occurrence ]
-    [@addDefaultGenerationPlan subsets=["template", "epilogue", "cli"] /]å
+    [@addDefaultGenerationPlan subsets=["template"] /]å
```
ml019: Looks like some weird character has snuck in ?

providers/aws/components/db/setup.ftl
```diff
@@ -191,10 +187,7 @@
         [/#if]
     [/#list]
 
-    [#local processorProfile = (getProcessor(occurrence, "RDS")?has_content)?then(
-                                getProcessor(occurrence, "RDS"),
-                                getProcessor(occurrence, DB_COMPONENT_TYPE )
-                            )]
+    [#local processorProfile = getProcessor(occurrence, "db" )]
```
ml019: So distinguishing between types and profile names? No actual dependency - yes?

providers/aws/components/es/setup.ftl
```diff
@@ -66,20 +103,6 @@
 
     [#local AccessPolicyStatements = [] ]
 
-    [#local esAccounts = getAWSAccountIds( solution.Accounts )]
```
ml019: This code was to support access to shared ES instances. Are we removing it on the basis of using links for this type of situation in the future?

providers/aws/components/db/setup.ftl
roleyfoley: Since each type is now defined in the reference configuration we can use the alternate name setup to handle this instead

providers/aws/components/es/setup.ftl
roleyfoley: Yeah I think we can look at it again once links are added. It also had some issues with the deny rules not working properly

PR #1166, opened 2020-01-09
<refactor> Move test case scenarios into provider

roleyfoley:
Another move around for getting the testing idea working 

Moves the loading processing of scenarios into the ability to load additional providers as SubProviders. SubProviders load like normal providers except when it comes to output management which is set by default to the existing GENERATION_PROVIDER variable 

Along with working for scenarios, this allows for codeontap users to load their own providers on top of standard providers, overriding default values or implementing their own components while still using the resources of another provider 

From a testing side this introduces a new provider `awstest` which contains scenarios designed to test the templates that we generate in a unit testing style, while keeping the testing isolated from the real world provider code

### DIFF
--- a/aws/createBlueprint.sh
+++ b/aws/createBlueprint.sh
@@ -7,9 +7,9 @@ trap 'exit ${RESULT:-1}' EXIT SIGHUP SIGINT SIGTERM
 function options() {
 
   # Parse options
-  while getopts ":f:hi:o:p:t:" option; do
+  while getopts ":f:hi:o:p:" option; do
       case "${option}" in
-          f|i|o|p|s|t) TEMPLATE_ARGS="${TEMPLATE_ARGS} -${option} ${OPTARG}" ;;
+          f|i|o|p) TEMPLATE_ARGS="${TEMPLATE_ARGS} -${option} ${OPTARG}" ;;
           h) usage; return 1 ;;
           \?) fatalOption; return 1 ;;
       esac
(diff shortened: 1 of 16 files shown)


ml019:
@roleyfoley Went to merge this but looks like it needs tweaking based on the discussion on providers yesterday?

roleyfoley:
removed subProviders ( GENERATION_SUBPROVIDERS)  and moved from `GENERATION_PROVIDER` to `GENERATION_PROVIDERS` which takes a comma separated list of providers 

The first provider in the list is used when determining outputs etc

Had to update the genplan output to go off the command line list of providers instead of determining them based on the placement

aws/createBlueprint.sh
```diff
@@ -30,17 +30,15 @@ DESCRIPTION:
...
-  $(basename $0) -i GENERATION_INPUT_SOURCE -t GENERATION_TESTCASE -s GENERATION_SCENARIOS
+  $(basename $0) -i GENERATION_INPUT_SOURCE -s GENERATION_SUBPROVIDERS
 
 PARAMETERS:
 
 (o) -i GENERATION_INPUT_SOURCE is the source of input data to use when generating the template
     -h                         shows this text
 (o) -o OUTPUT_DIR              is the directory where the outputs will be saved - defaults to the PRODUCT_STATE_DIR
-(o) -p GENERATION_PROVIDER     is the provider to for template generation
+(o) -p GENERATION_PROVIDERS     is the provider to for template generation
```
ml019: Normally we just allow the switch multiple times and gather the values e.g. -v switches for freemarker.sh

roleyfoley: Fixed up now accepts a list of -p values or an env var `GENERATION_PROVIDERS`

roleyfoley:
@ml019 ready when you are on this one?

PR #1169, opened 2020-01-12
<feature> CMDB v2.0.1 - Add du/placement subdirectories to the state tree

ml019:
Move existing state files into subdirectories based on the deployment unit.

Also add a placement subdirectory based on the the region.

Definition files are deleted as their names are occurrence rather than deployment unit based, but they are recopied as part of each build so will be refreshed on next build and placed in the correct deployment directory s a result. 

NOTE: The v2.0.1 upgrade will not be triggered by default as the maximum cmdb version for upgrade hasn't been altered. This will permit selective testing before the upgrade being automated more generally.

To test this upgrade set

```export GENERATION_MAX_CMDB_UPGRADE_VERSION=v2.0.1```

### DIFF
--- a/aws/createTemplate.sh
+++ b/aws/createTemplate.sh
@@ -679,33 +679,23 @@ function process_template() {
   # Defaults
   local passes=("template")
   local template_alternatives=("primary")
-  local cf_dir_default="${PRODUCT_STATE_DIR}/cf/${ENVIRONMENT}/${SEGMENT}"
 
   case "${level}" in
 
-    unitlist)
-      cf_dir_default="${PRODUCT_STATE_DIR}/cot/${ENVIRONMENT}/${SEGMENT}"
-      ;;
-
-    blueprint)
-      cf_dir_default="${PRODUCT_STATE_DIR}/cot/${ENVIRONMENT}/${SEGMENT}"
-      ;;
-
-    buildblueprint)
-      cf_dir_default="${PRODUCT_STATE_DIR}/cot/${ENVIRONMENT}/${SEGMENT}"
+    unitlist|blueprint|buildblueprint)
+      local cf_dir_default="${PRODUCT_STATE_DIR}/cot/${ENVIRONMENT}/${SEGMENT}"
       ;;
 
     account)
-      cf_dir_default="${ACCOUNT_STATE_DIR}/cf/shared"
+      local cf_dir_default="${ACCOUNT_STATE_DIR}/cf/shared"
       ;;
 
-    solution)
+    product)
+      local cf_dir_default="${PRODUCT_STATE_DIR}/cf/shared"
       ;;
 
-    segment)
-      ;;
-
-    application)
+    solution|segment|application)
+      local cf_dir_default="${PRODUCT_STATE_DIR}/cf/${ENVIRONMENT}/${SEGMENT}"
       ;;
 
     *)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: execution/contextTree.sh [hunk 1]
```diff
@@ -1500,6 +1506,88 @@ function upgrade_cmdb_repo_to_v2_0_0() {
...
+                    local placement="default"
+                    ;;
+                esac
+                local du_dir="${state_filepath}/${stack_deployment_unit}/${placement}"
+                info "${dry_run}Moving ${state_file} to ${du_dir} ..."
+                [[ -n "${dry_run}" ]] && continue
+                if [[ ! -d "${du_dir}" ]]; then
+                  mkdir -p "${du_dir}"
+                fi
+                git_mv "${state_file}" "${du_dir}" || { return_status=1; }
```
RossMurr4y: On my existing files this has not determined the ```${stack_deployment_unit}``` so it has copied them to the dir state/cf/integration/default rather than individual du dirs.

Example file name: ```seg-vpc-mswdev-eastus-template.json```

It does however create correct directories for them if I recreate the template.

This move has also moved the entire contents of the directory. I had a number of unrelated files in this dir which I had meant to tidy up and I can see that they've made the move to the new directory too.

ml019: @RossMurr4y thanks for testing on Azure

As discussed, the move is part of the CMDB v2.0.0 change and is expected. 

The failure to add the du directory was because the format of regions for Azure doesn't match that of AWS. I've added logic to change the expected file format based on deployment framework. While the AWS format is fairly unique and so anchors the regex correctly, te Azure ones are a bit more problematic. Would you mind providing the list of Azure regions you are currently using so I can look for those?

ml019:
@RossMurr4y I've coded it for ```eastus``` for now so you can confirm the logic works for Azure.

I suggest you run the v2.0.0 upgrade first and commit so you have an easy commit point to revert to if needing to test again.

Once advised of the other regions you require, I'll update the PR.

RossMurr4y:
> Once advised of the other regions you require, I'll update the PR.

@ml019 I pulled these directly from az cli with ```az account list-locations | jq '.[] | .name'```. There's more, but this should give you good coverage for your regex.

```
"eastasia"
"southeastasia"
"centralus"
"eastus"
"eastus2"
"westus"
"northcentralus"
"southcentralus"
"northeurope"
"westeurope"
"japanwest"
"japaneast"
"brazilsouth"
"australiaeast"
"australiasoutheast"
"southindia"
"centralindia"
"westindia"
"canadacentral"
"canadaeast"
"uksouth"
"ukwest"
"westcentralus"
"westus2"
"koreacentral"
"koreasouth"
"francecentral"
"francesouth"
"australiacentral"
"australiacentral2"
"uaecentral"
"uaenorth"
"southafricanorth"
"southafricawest"
"switzerlandnorth"
"switzerlandwest"
"germanynorth"
"germanywestcentral"
"norwaywest"
"norwayeast"
```

ml019:
@RossMurr4y I'd rather just cover the small set we are using right now. It is only important in the context of the upgrade. So probably just the two australia ones?

ml019:
@RossMurr4y  I've added the Australia ones for now along with eastus. Would you be able to test if it now correctly moves the Azure state files to du directories?

ml019:
@RossMurr4y thanks for testing on Azure
@roleyfoley This now works as expected for AWS and Azure. Ready to be merged when you are. Suggest squashing it as I had a few attempts to get the Azure stuff working.

aws/createTemplate.sh
```diff
@@ -718,6 +713,37 @@ function process_template() {
...
+    unitlist|blueprint|buildblueprint)
+      # No subdirectories for deployment units
+      ;;
+    *)
+      readarray -t legacy_files < <(find "${cf_dir}" -mindepth 1 -maxdepth 1 -name "*${deployment_unit}*" )
+
+      if [[ (-d "${cf_dir_default}/${deployment_unit}") || "${#legacy_files[@]}" -eq 0 ]]; then
+        case "${region}" in
+          us-east-1)
+            local placement="global"
```
roleyfoley: I think this would need to be further filtered for cert only. We have some actual deployments to us-east-1 which are part of the default placement

aws/createTemplate.sh
```diff
@@ -718,6 +713,37 @@ function process_template() {
       ;;
   esac
 
+  # Handle >=v2.0.1 cmdb where du/placement subdirectories were introduced for state
```
roleyfoley: Do we need the same in the manageStack/Deployment scripts?

aws/createTemplate.sh
ml019: @roleyfoley it is also for lambda at edge. What are the examples where the placement should be "default" when us-east-1 is targetted?

ml019:
@roleyfoley @RossMurr4y please review again. I've added support for the new layout in setStackContext which should cover managestack and manageDeployment.

Keen to try this on HA. I've not tested the manageStack side yet - just various combinations of repo cmdb versions and template creation.

PR #1170, opened 2020-01-13
Update deployment parameters

RossMurr4y:
This PR implements the following:

* Created a new PARAMETERS variable available through setStackContext. If the file referenced by the variable exists then the template is to deploy using it.
* Removed a legacy function from manageDeployment that was a left over artifact of the first design of the script designed to allow for parameter files. We're now handling them differently and so its no longer requried.
* Simplified the Template/Parameter file handling by removing unnecessary calls to jq. Can add them back in if file size starts to become a problem.
* Dynamically creating the list of arguments that get passed to the validate/create azure commands to allow for re-use and using only those applicable.


aws/manageDeployment.sh
```diff
@@ -241,19 +189,11 @@ function wait_for_deployment_execution() {
 
 function process_deployment() {
 
-  stripped_template_file="${tmp_dir}/stripped_template"
```
ml019: This will definitely be a problem with AWS - it is why we added the stripping in the first place :-)

Need to reinstate if this script will eventually be used for AWS as well...

aws/manageDeployment.sh
```diff
@@ -111,58 +111,6 @@ function register_resource_providers() {
   done
```
ml019: This file should be moved under /execution?

roleyfoley: this is a script used by a human, so shouldn't got under `/execution` 
scripts in /execution essentially shouldn't be a command in the cli. where commands in `/aws` are commands that should have a cli command

PR #1173, opened 2020-01-14
<feature> Dynamic loading of stack outputs

ml019:
Use the dynamic CMDB lookup capability to assemble the stack outputs at run-time.

Some additional parameters are needed to feed this process which previously was all done in the external composite assembly process.

@RossMurr4y There are a couple of spots in manageDeployments where assemble_stack_outputs is still being called. Need to discuss these with you.

Basics are in place - I just need to test it a bit more.


RossMurr4y:
After the rebase this is much clearer. Just need to remove all 3 calls to ```  assemble_composite_stack_outputs``` in place. I have tested and confirm this works as expected. I will be submitting a PR for one minor change to azure provider ```stackoutput.ftl``` as well.

ml019:
Superceded by https://github.com/codeontap/gen3/pull/1195

PR #1175, opened 2020-01-16
Update Baseline Storage Reference

RossMurr4y:
Azure provider doesn't use "buckets" storage type. This change allows the providers to set their own, whilst still defaulting to use the AWS values in alignment with previous changes to "cmk", "ec2KeyPair" etc as well.


RossMurr4y:
@roleyfoley any idea what I've done differently all of a sudden that would cause my rebasing of my master branch to keep showing up as commits like this? I've only made 1 change here.

providers/shared/components/baseline/id.ftl
```diff
@@ -6,7 +6,7 @@
     [#return formatSegmentResourceId(SEED_RESOURCE_TYPE)]
 [/#function]
 
-[#function getBaselineComponentIds links cmk="cmk" ssh="ec2KeyPair" oai="originAccessId"]
+[#function getBaselineComponentIds links storage="bucket" cmk="cmk" ssh="ec2KeyPair" oai="originAccessId"]
```
roleyfoley: When adding new optional parameters its probably better to add it to the end instead of the front. Functions use positional arguments so to prevent updates across the codebase its probably safer

RossMurr4y: @roleyfoley I considered that but after doing a quick search in the gen3 repo found there were no current calls to it with non-defaults, and with only the one plugin provider at the moment I thought we'd be safe. Can update it if you feel we should.

RossMurr4y: Moved dd96aec

PR #1188, opened 2020-01-22
<feature> APIGateway - error on missing cert

roleyfoley:
Yell if the certificate can't be found so that cloudformation doesn't have to do it for us

### DIFF
--- a/providers/aws/components/apigateway/setup.ftl
+++ b/providers/aws/components/apigateway/setup.ftl
@@ -458,7 +458,10 @@
                     valueIfTrue(
                         {
                             "RegionalCertificateArn":
-                                getArn(value["domain"].CertificateId, true, regionId),
+                                ( getArn(value["domain"].CertificateId, true, regionId) )?has_content?then(
+                                    getArn(value["domain"].CertificateId, true, regionId),
+                                    "COTFatal: Could not find certificate " + value["domain"].CertificateId
+                                ),
                             "EndpointConfiguration" : {
                                 "Types" : [endpointType]
                             }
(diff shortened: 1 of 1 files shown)


providers/aws/components/apigateway/setup.ftl
```diff
@@ -458,15 +458,21 @@
                     valueIfTrue(
                         {
                             "RegionalCertificateArn":
-                                getArn(value["domain"].CertificateId, true, regionId),
+                                ( getArn(value["domain"].CertificateId, true, regionId) )?has_content?then(
```
ml019: contentIfContent(
  getArn(value["domain"].CertificateId, true, regionId) ), 
  "COTFatal: Could not find certificate " + value["domain"].CertificateId
)

means function only called once mostly

providers/aws/components/apigateway/setup.ftl
```diff
@@ -458,15 +458,21 @@
...
+                                ),
                             "EndpointConfiguration" : {
                                 "Types" : [endpointType]
                             }
                         },
                         isRegionalEndpointType,
                         {
                             "CertificateArn":
-                                getArn(value["domain"].CertificateId, true, "us-east-1")
+                                ( getArn(value["domain"].CertificateId, true, "us-east-1") )?has_content?then(
```
ml019: same as previous content...

providers/aws/components/apigateway/setup.ftl
roleyfoley: Fixed

providers/aws/components/apigateway/setup.ftl
roleyfoley: Fixed

PR #1190, opened 2020-01-23
SPA Port Attribute Option

RossMurr4y:
Rather than hard-code HTTPS into the SPA Component, this PR introduces the ability to set a port for an SPA provider whilst defaulting to HTTPS.

This PR also migrates some DNS related functions and variables and moves them into Common.ftl alongside other domain-related functions.

### DIFF
--- a/providers/shared/components/spa/id.ftl
+++ b/providers/shared/components/spa/id.ftl
@@ -50,8 +50,18 @@
             {
                 "Names" : "Port",
                 "Description" : "The port used to access the SPA content.",
-                "Type" : STRING_TYPE,
-                "Default" : "https"
+                "Children" : [
+                    {
+                        "Names" : "HTTP",
+                        "Type" : STRING_TYPE,
+                        "Default" : "http"
+                    },
+                    {
+                        "Names" : "HTTPS",
+                        "Type" : STRING_TYPE,
+                        "Default" : "https"
+                    }
+                ]
             }
         ]
 /]
--- a/providers/shared/components/spa/id.ftl
+++ b/providers/shared/components/spa/id.ftl
@@ -46,6 +46,12 @@
                 "Description" : "The path appended to files that will be used for config",
                 "Type" : STRING_TYPE,
                 "Default" : "config"
+            },
+            {
+                "Names" : "Port",
+                "Description" : "The port used to access the SPA content.",
+                "Type" : STRING_TYPE,
+                "Default" : "https"
             }
         ]
 /]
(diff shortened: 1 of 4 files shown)


ml019:
@RossMurr4y doesn't this need code to read the new setting as well? Maybe I'm missing something - are you planning a second PR to use this new setting?

RossMurr4y:
@ml019 It's not a mandatory option, so was going to leave the AWS provider alone. This was for the Azure provider, so my PR will accompany that components PR.

providers/shared/components/spa/id.ftl
```diff
@@ -46,6 +46,12 @@
                 "Description" : "The path appended to files that will be used for config",
                 "Type" : STRING_TYPE,
                 "Default" : "config"
+            },
+            {
+                "Names" : "Port",
+                "Description" : "The port used to access the SPA content.",
+                "Type" : STRING_TYPE,
+                "Default" : "https"
```
roleyfoley: I thin this should actually be 

```
           {
                "Names" : "Port",
                "Description" : "The port used to access the SPA content.",
                "Children" : [
                    {
                        "Names" : "HTTP",
                        "Type" : STRING_TYPE,
                        "Default" : "http"
                    },
                    {
                        "Names" : "HTTPS",
                        "Type" : STRING_TYPE,
                        "Default" : "https"
                    }
                ]
           }
```

Then the SPA should priovide 2 attribtues which match the port number from these lookups

providers/shared/references/Port/id.ftl
```diff
@@ -23,7 +23,18 @@
         },
         {
             "Names" : "Port",
-            "Type" : NUMBER_TYPE
+            "Children" : [
```
roleyfoley: This is replaced by my suggestion above

roleyfoley: Reference data is shared data between all component types. And not all component types care about Http or Https

RossMurr4y: Have completed this in dbf1030

providers/shared/references/Port/id.ftl
```diff
@@ -23,7 +23,18 @@
...
+                    "Names" : "Http",
+                    "Type" : NUMBER_TYPE,
+                    "Default" : 80
+                },
+                {
+                    "Names" : "Https",
+                    "Type" : NUMBER_TYPE,
+                    "Default" : 443
+                }
+            ]
```
roleyfoley: If you've added the protocol specific ports in the solution configuration it shouldn't be required in the reference data

RossMurr4y: Yep, thanks. I had a misunderstanding on how these definitions and reference data worked that the meetings on Friday cleared up. I've fixed that in 389057e

PR #1191, opened 2020-01-23
WAF MasterData Inputs into Shared Provider

RossMurr4y:
Moved WAF values/conditions/rules/groups/profiles & securityprofiles into shared provider masterdata.

These options are unlikely to be applicable for every provider exactly as is, but basing WAF decisions on these settings within a shared provider ensures that the outlines for the OWASP configurations only need to be defined in one place and the providers can implement them in their own way.

* Also moves the getWAFProfileRules freemarker function into the shared provider, alongside the reference definition.


RossMurr4y:
@roleyfoley I've moved the utility function ```getWAFProfileRules``` into a new file alongside the reference definition. This seems like quite a logical place to me to store such utility functions. Does that work?

ml019:
@RossMurr4y Looks as good a spot as any to me. Happy to merge if you are. I'll be deploying lots of APS API Gateways over the weekend so will kick the tyres from the AWS side.

PR #1194, opened 2020-01-25
<fix> WAF Support

ml019:
The change is primarily to correct a few issues with the changes made to WAF support to enable reuse of some of the AWS implementation with Azure;

- add loading of WAF reference data
- corrections to the reference data definitions
- add the ability to specify shared routines associated with a particular reference type via a reference.ftl file along side the id.ftl file.
- adjust the AWS WAF implementation to use the reference data rather than direct blueprint access

In the process, a couple of more general improvements were made;

- abstract access to reference data into a getReferenceData() function base using the reference data type as input
- simplify the initialisation of reference data from a blueprint given that we already know the plural for the reference type when loading the data
- add the ability for getCompositeObject() to validate arrays of objects as needed for the WAF reference data.

The last point above means that where in the past the parsing needed tostop with declaring the value of ARRAY_OF_OBJECT_TYPE, the config can now be prescriptive as to the content of the child objects as well. This can also be nested e.g. array of objects, each of which has an attribute which is an array of objects etc.


ml019:
@roleyfoley @RossMurr4y need to merge this to do more testing with dynamic loading. Please still review it and I will make any necessary changes.

PR #1195, opened 2020-01-27
<feature> Dynamic loading of inputs

ml019:
Use the dynamic CMDB lookup capability to assemble the blueprint, settings, stack outputs and definitions at run-time.

The composite blueprint is still assembled in bash to keep the various checks happy - a subsequent PR will clean this up.

Also add the idea of a stack of input contexts to permit changing of the inputs for instance when following cross segment links. Again a subsequent PR will move a lot of the logic currently inside the bootstrap into the input context switching logic.

Dynamic loading requires that the engine now receive the inputs that formerly fed the composite assembly process. They are used to populate the initial inputs context.

Some effort has gone into caching both the dynamically loaded CMDB information and input contexts. More work will be needed.

Generic routines were also added for handling caches, dictionaries and stacks, with the cmdb and input context support using these new routines. A subsequent PR will refactor the caching of provider templates to use the same mechanisms.

A new version of the freemarker wrapper is also introduced which supports searching of the plugins tree to permit dynamic generation of lists of templates to load.


ml019:
This PR was for exposure comments only. Closing in preparation for the formal PR.

PR #1230, opened 2020-02-12
<feature> Plugin Introspection

ml019:
Using the getPluginTree() function introduced in the new version of the wrapper, determine at runtime what providers are present in the plugins provided on the command line.

The core files for each provider are automatically included, but services and components are left to be lazy loaded.

As the process knows which templates are present, it directly includes them rather than using the includeTemplates macro.

Rename the provider specific top level ftl file to become a provider "marker" file, along the lines of the various marker files used in cmdbs.

Tidy up the bootstrap a bit my moving the logic of numeric/text log levels into logging.ftl and the seeding of input data into a seed.ftl file under engine/inputdata.

@RossMurr4y you will need to rename azure.ftl to provider.ftl for the azure provider to be automatically found and loaded.


ml019:
@roleyfoley @RossMurr4y this is ready for review. I will further test with ICS at HA but would appreciate some Azure testing @RossMurr4y to make sure I haven't broken anything there.

ml019:
@roleyfoley At present I am automatically including the basics of all found providers (via presence of provider.ftl file). It keeps things simple code wise - we don't have to continually check if the basics are loaded. Depending on the speed, I could even load all the services and components as well because it is trivial to iterate over them.. Freemarker seems pretty efficient at processing includes but I figured components and services was where the bulk of the templates are so lazy loading is probably still warranted.

However if one of these includes comes with a side effect, then it needs to be explicitly loaded - hence the distinction in the marker file. The actual processing is identical - its just whether it is automatically or explicitly loaded. 

I think if we refactored the way the way scenarios are triggered, then I could go back to every plugin being the same. I agree having them all the same is easier to explain.

At present the awstest plugin has the side effect of adding to the scenariolist which in turn triggers changes to the input data via the bootstrap.

Open to suggestions - would like to be able to treat all providers as equals (albeit with shared loaded first and the remainder in alphabetical order)

roleyfoley:
I thought the idea of introducing the provider flag -p was to explicitly load the plugins rather than load through discovery. Since this implies that you might not want to load providers all the time. 

The order of the -p flags also controls which providers output framework is used. This is a workaround for the resource group setup but will this still work? or just work because aws is longer than awstest? 

If you are running an include on the provider file does it matter if it has a different name?

ml019:
@roleyfoley There is the point you want to start (for which yes a particular provider is needed, but links could take you to products using other providers, so you potentially end up loading them anyway.

I can change back to the current behaviour for now, but it means more logic in the link following to load extra providers if required.

Exchanges might fix some of this in that it isolates the linker from the specifics of what's behind the exchange but we're not quite there yet.

ml019:
@roleyfoley I've modified the startup to follow the existing pattern where the providers are explicitly provided, along with their load order.

Issue #1232, opened 2020-02-13
Contract Schema


roleyfoley:
Part of #1231 - This is to determine the format of the contrac which will be used to pass information between the generation and execution process 

Rough structure to start with 

- A contract has 1 or many stages 
- A Stage has 1 or many steps 
- A step outlines a task that needs to be completed 

**Contract Structure** 
A contract has the following structure 
- Name: A name of the contract ( meta data ) 
- Timestamp: when the contract was generated 
- Stages: An array of Stages 

**Stage Structure** 
A stage has the following structure 
- Id: A unique Id of the stage 
- Execution Mode: Parallel, Serial 
    - How the steps in the stage will be executed 
- Steps - An array of Steps that will be run 

**Step Structure**
A step has the following structure 
- Id: A unique id of the step 
- Task: The name of the task that needs to be executed
- Parameters: A collection of parameters which will be passed to the task being run 

Example 
```
{
    "Metadata" : {
        "Name" : "GenerationContract",
        "TimeGenerated" : "2019-02-13:14:14
    },
    "Stages" : [
        {
            "Id" : "CreateCFN",
            "ExecutionMode" : "Serial",
            "Steps" : [
                {
                    "Id" : "CreateCFNTemplate",
                    "Task" : "aws-cf-create-template",
                    "Parameters" : {
                        "DeploymentUnit" : "thing-v1",
                        "Subset" : "thing-v1"
                    }
                },
                {
                    "Id" : "CreateCLITemplate",
                    "Task" : "aws-cf-create-template",
                    "Parameters" : {
                        "DeploymentUnit" : "thing-v1",
                        "Subset" : "cli"
                    }
                }
            ]
        }
    ]
}
```

So this contract would run the aws-cf-create-template task twice, with the given parameters

roleyfoley:
Implemented as part of #1240 I think we can call the inital schema implementation done

Issue #1231, opened 2020-02-13
Architecture - Engine and Executor 


roleyfoley:
This issue is to introduce an architectural idea to clearly define the different functions performed in how codeontap works 

At the moment we have a tight coupling between our bash based scripts and the freemarker template generation engine. With the introduction of the python based cli tool ( https://github.com/codeontap/gen3-cli ) this causes issues in how we can integrate the CLI into codeontap. At the moment we have wrapped the existing bash scripts in python based code. While python is OS independent, bash isn't easily OS independent and also doesn't have a packagement process ( outside of the OS). This isn't ideal.

This process has highlighted two distinct parts of the codeontap gen3 function ( in this repo) :

- **Generation process** This side of codeontap is responsible for generating the files ( CFN templates, bash scripts, config files ) that are required to perform a deployment. This process is provided through our freemarker engine and the CMDB 
- **Execution process**. This side is responsible for implementing the deployment defined in the template generation process. This process is provided through our bash based scripts ( see manageStack.sh manageDeployment.sh etc. 

At the moment there is quite tight coupling between these two process with the generation process creating bash based scripts that are run directly by the execution process. This ties directly in with the issue we have with the python CLI 

Instead of directly generating the scripts that are executed by the execution engine we should establish an agreed contract between the generation and execution processes

roleyfoley:
The proposal here is to make the contract between the generation and the execution process Plans. 

Plans are generated by the generation process and outline the tasks required to complete a given task. 

At the moment we have 2 tasks:
- Generation - This outlines the documents which need to be created by a call to createTemplate. Currently this actually has a plan, however this is implemented as a bash output file which locks it into a bash based execution process
- Deployment - Implementing the deployment of a component based on the documents created by the Generation task. The plan for this is currently implemented in the code of manageStack or manageDeployment.

With the introduction of plans we would have a standard format for both of these tasks. The plan would outline each piece of work required to perform the task along with the parameters required to do this. The plan would be in a standard data model ( JSON ) and outline standard work item names which the execution process is responsible for implementing. 

The work items in the plan should have no knowledge of the execution process which will run the plan,  this separation allows for mobility beween the generation and the execution process with the potential to have multiple execution runners ( e.g. a python cli execution process, and a powershell module ). Both of the execution runners would receive the same plan but they would have individual implementations of the work items. As long as they all achieve the same result, we are happy.

ml019:
[...]

PR #1240, opened 2020-02-19
<feature> Contract output generation support

roleyfoley:
This PR introduces the idea of contracts outlined in #1231 based on the schema defined in #1232 

The idea is that the engine generates contracts which outline a set of tasks which need to be completed in order to perform a piece of work. 

Tasks can be scheduled and ordered through two levels, stages and steps. Stages group steps and specify how its steps can be executed, either parallel or serial at this stage. 

The first implementation of this which is part of this PR is the Generation Contract. This replaces the existing genplan which essentially outlines the same thing but without the execution ordering. 

The Generation plan creates a stage per template alternative along with an extra stage for the pregeneration subset. Each of these stages as a task `process_template_pass` which runs the freemarker engine.

The standard generation stages are configured for parallel execution with pregeneration run in serial. This is to allow for other tasks to be run during pregeneration which depend on the initial script.

**Tasks** 

Tasks are documented in the engine, with the expectation that the driver making the call can support each tash. Each task has a set of parameters that it accepts and they are validated in the same way that we validate our other input sources

**Bash driver** 

This PR adds initial support for parsing contracts, the initial implementation reads through the stages and steps in the contract and generates a single line function call with the type of the task and the parameter values as a delimited list next to it 

E.g. for the step 
```        {
          "Id": "generationXprimaryXconfig",
          "Type": "process_template_pass",
          "Parameters": {
            "provider": "aws",
            "framework": "cf",
            "outputType": "json",
            "outputFormat": "",
            "outputSuffix": "config.json",
            "subset": "config",
            "alternative": "primary"
          }
        }
``` 
a line in the contract command file will be created as 

```
core_process_template_pass aws;cf;json;;config.json;config;primary
```

The delimiter can be configured through the utilty task. The intention with this setup is that if the delimiter " " is used you could potentially execute the created script directly to run the tasks provided. We don't do this as of yet.


ml019:
should output macro should be under shared/deploymentframeworks/default/output.ftl?

ml019:
@roleyfoley are you removing the genplan files as part of the new process? I think they will be left behind because they will no longer be generated. 

If easy to do, worth including cleanup in this PR bit otherwise, can deal with it as a cmdb cleanup linked to the train version which includes these changes (assuming we start including gen3 version in the gen3 repo).

PR #1241, opened 2020-02-19
<fix> Build blueprint production

ml019:
Build blueprint generation was failing because it was forcing the extension of its genplan to be .json.

Tweak the output file expected by all the consumers so the standard output suffix as included in the generated genplan will work correctly.

Also fix the generation of a blueprint to only force the output suffix in the pass that generates the blueprint.

The investigation was triggered by an attempt to encrypt credentials using manageCrypto.sh - one of the clients for createBuildblueprint.

This change thus includes some extra debug messages and ciphertext prefix handling, along with a fix to the handling on in file attribute updates.

### DIFF
--- a/aws/createTemplate.sh
+++ b/aws/createTemplate.sh
@@ -353,7 +353,7 @@ function process_template_pass() {
 
       pass_level_prefix["config"]="blueprint"
       pass_description["config"]="blueprint"
-      output_suffix=".json"
+      [[ ${pass} == "config" ]] && output_suffix=".json"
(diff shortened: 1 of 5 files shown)


ml019:
@roleyfoley I also tested that the createBlueprint.sh works as expected. The only side effect you get is a genplan.sh file under state/cot/{product}/{segment} but that doesn't feel wrong.

roleyfoley:
looks good to me, thanks for catching that

PR #1247, opened 2020-02-21
<feature> ECS OnDemand Compute Provider

roleyfoley:
Adds support for on demand Ec2 Instance autoscaling for ECS ( https://docs.amazonaws.cn/en_us/AmazonECS/latest/developerguide/cluster-auto-scaling.html ) 

Essentially it creates a tracked autoscaling policy on the Ec2 Autoscale cluster along with a metric of the required resources for the ECS cluster. When a task starts and there are now instances to service the task, it will sit in a provisioning state until the autoscaling action has completed. 
The same goes when there aren't any tasks to run, the cluster will remove any ec2 instances running without tasks on them 

This required a basic implementation of the ECS capacity providers which provide a way to define where containers should be deployed ( ec2, fargate, fargate spot etc) using weighted rules 

As this is a new feature and people might have their own EC2 autoscaling rules setup this is not enabled by default 

An example - This is an ECS cluster using a autoscale group with a min and desired processor count of 0. When I schedule a task or a service on this cluster, ECS will update a metric on the demand required on the cluster and autoscaling will bump it up to 1 EC2 instance to meet the need. 

```
{
    "Tiers" : {
        "app" : {
            "Components" : {
                "apphost" : {
                    "ECS" : {
                        "Instances" : {
                            "default" : {
                                "DeploymentUnits" : ["apphost"]
                            }
                        },
                        "Profiles" : {
                            "Processor" : "default"
                        },
                        "ComputeProvider" : "ec2OnDemand"
                        }
                }
            }
        }
    },
    "Processors" : {
        "default" : {
            "ECS" : {
                "Processor" : "t3.small",
                "DesiredCount" : 0,
                "MaxCount" : 2,
                "MinCount" : 0
            }
        }
    }
}
```

### DIFF
--- a/execution/utility.sh
+++ b/execution/utility.sh
@@ -593,9 +593,9 @@ function split_cli_file() {
   local cli_file="$1"; shift
   local outdir="$1"; shift
 
-  for resource in $( jq -r 'keys[]' <"${cli_file}" ) ; do
-    for command in $( jq -r ".$resource | keys[]"<"${cli_file}" ); do
-        jq ".${resource}.${command}" >"${outdir}/cli-${resource}-${command}.json" <"${cli_file}"
+  for resource in $( jq -r 'keys[]' < "${cli_file}" ) ; do
+    for command in $( jq -r ".$resource | keys[]" < "${cli_file}" ); do
+        jq ".${resource}.${command}" > "${outdir}/cli-${resource}-${command}.json" <"${cli_file}"
     done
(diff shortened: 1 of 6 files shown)


ml019:
@roleyfoley looks cool and exactly what code publican needs. I like the way you still decide what the base is, so you can decide whether you have some capacity ready to go all the time or none.

PR #1251, opened 2020-02-23
<fix> read ecsCapacityProviderOnDemand id

kshychko:
Read ecsCapacityProviderOnDemand id only when compute Provider is ec2OnDemand
Use ecsCapacityProviderOnDemand id in cli as a capacity provider name.

### DIFF
--- a/providers/aws/components/ecs/setup.ftl
+++ b/providers/aws/components/ecs/setup.ftl
@@ -21,7 +21,6 @@
     [#local ecsLogGroupName = resources["lg"].Name ]
     [#local ecsInstanceLogGroupId = resources["lgInstanceLog"].Id]
     [#local ecsInstanceLogGroupName = resources["lgInstanceLog"].Name]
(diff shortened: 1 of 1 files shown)


providers/aws/components/ecs/setup.ftl
```diff
@@ -579,7 +579,7 @@
                     "       || return $?",
                     "       create_ecs_capacity_provider" +
                     "       \"" + region + "\" " +
-                    "       \"$\{asgName}\" " +
+                    "       \"" + ecsOnDemandCapacityProviderId + "\" " +
```
roleyfoley: I ended up making it the asgname as the capacity providers can’t be updated or deleted at the moment
So if the asg changes then using the same Id doesn’t work

I left the Id field there for when aws add capacity providers to CloudFormation

Probably better to remove it to make it more obvious why I did it this way

kshychko: @roleyfoley , I rolled back the changes and added a comment

PR #1256, opened 2020-03-04
<fix> Only delete legacy OAI if not in use

ml019:
When migrating legacy OAIs to the baseline equivalent, don't try and delete the legacy OAI if it is still in use.

Because the legacy stack will still be deleted (as is replaced with baseline's own OAI), cleanup of the legacy OAI will need to be performed manually.

### DIFF
--- a/execution/utility.sh
+++ b/execution/utility.sh
@@ -1631,6 +1631,31 @@ function delete_oai_credentials() {
   return 0
 }
 
+function is_oai_credential_used() {
+  local region="$1"; shift
+  local name="$1"; shift
+
+  local oai_status_file="$( getTempFile oai_used_XXXXXX.json)"
+  local oai_id=
+  local oai_etag=
+
+  # Check for existing identity
+  aws --region "${region}" cloudfront list-cloud-front-origin-access-identities > "${oai_status_file}" || return $?
+  oai_id=$(jq -r ".CloudFrontOriginAccessIdentityList.Items[] | select(.Comment==\"${name}\") | .Id" < "${oai_status_file}") || return $?
+
+  # check if used if present
+  if [[ -n "${oai_id}" ]]; then
+    oai_ids=$(aws --region "${region}" cloudfront list-distributions --query "DistributionList.Items[].Origins.Items[].S3OriginConfig.OriginAccessIdentity" --output text)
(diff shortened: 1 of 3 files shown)


providers/aws/components/baseline/setup.ftl
```diff
@@ -472,9 +472,15 @@
                                         "    ;;",
                                         "  create|update)",
                                         "    info \"Removing legacy oai credential\"",
-                                        "    delete_oai_credentials" + " " +
+                                        "    if is_oai_credential_used" + " " +
                                                "\"" + regionId + "\" " +
-                                               "\"" + legacyOAIName + "\" || return $?",
+                                               "\"" + legacyOAIName + "\"; then",
+                                        "      warn \"Legacy OAI in use - it will need to be manually deleted ...\"",
```
roleyfoley: "it will be deleted once it is no longer in use"

execution/utility.sh
```diff
@@ -1631,6 +1631,31 @@ function delete_oai_credentials() {
   return 0
 }
 
+function is_oai_credential_used() {
+  local region="$1"; shift
+  local name="$1"; shift
+
+  local oai_status_file="$( getTempFile oai_used_XXXXXX.json)"
+  local oai_id=
+  local oai_etag=
```
roleyfoley: Is this required?

ml019: Extra vars are a copy/paste - I'll clean up. At the moment, the logic needs a temp file to extract the oai_id but I could trim it to use --query instead.

providers/aws/components/baseline/setup.ftl
ml019: Is there some cleanup logic somewhere? The AOI cleanup code from baseline is removed once the OAI pseudo stack is deleted.

ml019:
@roleyfoley tested this with HA and is working ok - I restored the cmk oai pseudo stack to simulate what should have happened. I'll keep going and see if the cleanup works when the oai use is removed.

Also tested that the flowlogs configuration tweak works correctly.

ml019:
@roleyfoley cleanup happened correctly once legacy OAI not used. First pass removed it, next pass deleted the code to check for it.

This PR is ready for squashing and merging

PR #1260, opened 2020-03-05
<feature> RDS Minor Version Control

ml019:
Add the ability for the Instance version to include a minor version indication, initially for MySQL.

### DIFF
--- a/providers/aws/components/db/setup.ftl
+++ b/providers/aws/components/db/setup.ftl
@@ -33,8 +33,21 @@
     [#local auroraCluster = false]
 
     [#local engine = solution.Engine]
+    [#local majorVersion = solution.EngineVersion]
+
+    [#-- If a minor version is provided, assume engine version is the major version --]
+    [#-- Also disable minor version upgrades --]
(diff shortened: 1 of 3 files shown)


ml019:
@roleyfoley tested ok on smallcraft.

ml019:
There is quite a bit of variety based on the engine  as to what format versions take. I was just trying to accommodate the scheme mysql uses but happy to expand.

So thinking we deprecate "EngineVersion" in favour of explicit major and minor indications. Presence of minor then disable auto-minor updates. We will then combine these to give the full engine version based on the engine type.

Is that what you are thinking? Other qustion is whether we add major and minor as top level attributes or add a "Version" attribute with major and minor as children. Thinking this is better as it permits additional attributes around version if we need them.

Sounds ok?

ml019:
@roleyfoley Ended up that just adding an optional minor version was pretty straightforward. Ready for review.

roleyfoley:
Closing as part of https://github.com/codeontap/gen3/issues/1022

PR #1267, opened 2020-03-12
source all plugin-provided utility.sh files

RossMurr4y:
## Description

createTemplate will now dot-source all specified provider plugin utility scripts as well as the shared provider one.
Though a common need, this could not be added into the common.sh script as it will not consistently have access to the GENERATION_PROVIDERS variable. Moving the dot-sourcing of common.sh until later in existing shell scripts (say, to after getopts and defaults have been applied) only causes other problems due to other variables now being missing.

## Motivation and Context
Currently, the createTemplate.sh script only loads the shared provider utility.sh when it dot-sources the file common.sh at the top of the script. This change implements a check for a similar utility.sh script within all specified provider plugin directories and dot-sources those too, if found. This allows for each provider to load provider-specific related utilities which are particularly important for non-"template" generation-contract passes such as "pregeneration".

## How Has This Been Tested?

As this is a minimal change effecting only the azure plugin at this stage. I have tested it by executing the following on my existing test CMDB (successful): 
`cot create template -p azure -f arm -l application -u apigw`

## Types of changes

## Breaking Actions

Non-breaking change.

## Checklist:

- [x] none above apply.

### DIFF
--- a/aws/createTemplate.sh
+++ b/aws/createTemplate.sh
@@ -93,6 +93,13 @@ function options() {
   fi
   GENERATION_PROVIDERS=$(listFromArray "GENERATION_PROVIDERS" ",")
 
+  # load utility functions for all loaded providers
+  for provider in "${GENERATION_PROVIDERS}"; do
+    if [[ -f "${GENERATION_PLUGIN_DIRS}/${provider}/utility.sh" ]]; then
+      . ${GENERATION_PLUGIN_DIRS}/${provider}/utility.sh
+    fi
+  done
+
   # Check level and deployment unit
   ! isValidUnit "${LEVEL}" "${DEPLOYMENT_UNIT}" && fatal "Deployment unit/level not valid" &&  return 1
 


RossMurr4y:
Try as I might I couldn't add this to the common.sh (where I think is the best place for it) and move the dot-sourcing of the common.sh in createTemplate.sh to a location that actually worked. Happy to revisit this if another pair of eyes have a better idea.

roleyfoley:
What about a find in `GENERATION_PLUGIN_DIRS` instead of the GENERATION_PROVIDERS lookup as part of the common.sh load process? 

GENERATION_PROVIDERS is only required when you are generating documents, if you have a utility used during manageDeployment tasks?

That would make it a more generic approach

roleyfoley:
Actually this shouldn't be in the engine, it should be in the executor. We either need to support plugins for the executor or its just a part of the executor

RossMurr4y:
I have created a new ticket for this - hamlet-io/executor-bash#11 - closing.

PR #1295, opened 2020-03-26
<feature> Namespaced provider specific configuration attributes

ml019:
## Description
Configuration attributes that are provider specific must now be prefixed with the provider value, e.g.

```
"aws:Parameter" : "xyz"
```

It also fixes a bug in the combining of shared and provider specific attributes, where the precedence of the missing value operator prevented the provider specific attributes from being considered.

## Motivation and Context
It is desirable to explicitly identify parameters which are targetted at a specific provider. It will also permit documentation to indicate what are the shared and provider specific attributes for a component.

To date, no provider specific attributes have been used so it is a good time to make this change. 

## How Has This Been Tested?
Local deployment

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/engine/component.ftl
+++ b/engine/component.ftl
@@ -76,29 +76,53 @@
     }
 ] ]
 
-[#macro addResourceGroupInformation type attributes provider resourceGroup services=[] ]
-    [#-- Special processing for profiles --]
-    [#if
-        (provider == SHARED_ATTRIBUTES) &&
-        (resourceGroup == DEFAULT_RESOURCE_GROUP) ]
-        [#local extendedAttributes = [] ]
-        [#local profileAttribute = coreProfileChildConfiguration[0] ]
-        [#local settingsNamespacesAttribute = coreSettingsNamespacesConfiguration[0]]
-        [#list attributes as attribute ]
-            [#if asArray(attribute.Names!attribute.Name)?seq_contains("Profiles")]
-                [#local profileAttribute +=
+[#macro addResourceGroupInformation type attributes provider resourceGroup services=[] prefixed=true ]
+    [#if provider == SHARED_ATTRIBUTES ]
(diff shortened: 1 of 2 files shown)


ml019:
@RossMurr4y you weren't going crazy. The binding behaviour of the missing value was preventing any provider specific attributes from being processed.

RossMurr4y:
@ml019 looks like only parent attributes require the prefix. That is quite functional but thought it worth documenting here. 
```
"aws:Parameter" : {
  "instanceOfParameter" : {
    "childParameter" : "has-no-prefix"
  }
}
```

RossMurr4y:
[Freemarker suggests](https://freemarker.apache.org/docs/dgui_template_exp.html#dgui_template_exp_var_hash) that you should be able to reference these object names via either `solution.aws\:Parameter` or `solution["aws:Parameter"]` but neither are working for me. It's also a but harder to use.

If this method is going to be harder to use, could we accomplish the same thing either by underscore, a provider "tier" ( `solution.aws.Parameter` ) or simply add an extra parameter below the attribute name that identifies it as a non-shared parameter: 

```
"Parameter" : {
  "Provider" : "aws",
  "Enabled" : true
}
```

ml019:
@RossMurr4y @roleyfoley I can go through and add namespace to children but that does seem like a lot of overhead - effectively children are scoped to parent which explicitly says azure only.

Thoughts?

RossMurr4y:
Leave it as it is, I just thought it would be worth documenting somewhere.

RossMurr4y:
@ml019 we good to merge this? Starting to need it more and more.

roleyfoley:
I think just at the top is fine, it makes it obvious that everything under this namespaced paramter is provider specific

PR #1298, opened 2020-03-30
Refactor - Adaptor to be Provider Specific

### LINKED ISSUE #1296: Issue - Adaptor Component - Provider dependencies
## Expected Behavior

All providers should be capable of using the shared component definitions.

## Current Behavior

The Adaptor component is intended to work out of the box for all providers and has its setup instructions defined there, however it involves a call to the function `getBuildScript()` which  will author a bash script that itself will refer to the AWS plugin utility function `copyFilesFromBucket()`. 

## Possible Solution

Allow for the name of the function to call and the arguments for that function to be attributes on the component. Each provider can then either set defaults within their plugin or require users to add to their own cmdb.

If this were to be implemented as a global variable in providers/shared/components/component.ftl, then providers could provide their own component.ftl and overwrite this global variable.

## Context

Working with the Azure plugin provider.

RossMurr4y:
## Description

Addresses & Closes #1296 
Removes the adaptor component setup file from the shared provider.

## Motivation and Context

After discussions with @roleyfoley it was determined that this file was not intentional. It is not possible/feasible at this stage for the entire component to be provider independent and not require provider setup/state files. This file is the AWS version, although requires minimal changes to be available to other providers also. 

## How Has This Been Tested?

Original testing for the Adaptor component was performed when implemented - this file is just a remnant of that testing.

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.


roleyfoley:
Thanks for catching that. I thought I'd cleaned them up

PR #1305, opened 2020-04-23
Feature - Sku Profiles

RossMurr4y:
## Description

This PR implements `skuProfiles` the loading of the skuProfiles reference type into the shared provider. Though the skuProfiles reference type exists within the Azure plugin, there is a current issue with the load-order of reference data from plugins that will be looked at in another ticket. This allows correct loading in the meantime.

## Motivation and Context

Azure has a common concept of a SKU or "Stock Keeping Unit" (think of it like a make/model information) across many resources. The SKU requirements change across resources - sometimes its just a name to indicate the pricing tier of a resource you want, other times its make, model, family, generation etc. This reference type allows for the base provider to implement some default profiles within masterdata, whilst allowing customisation through the CMDB.

## How Has This Been Tested?

Local template generation alongside hamlet-io/engine-plugin-azure#112

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/engine/common.ftl
+++ b/engine/common.ftl
@@ -385,6 +385,20 @@ behaviour.
     [/#if]
 [/#function]
 
+[#function getSkuProfile occurrence type extensions... ]
+    [#local tc = formatComponentShortName(
+                    occurrence.Core.Tier,
+                    occurrence.Core.Component,
+                    extensions)]
+    [#local defaultProfile = "default"]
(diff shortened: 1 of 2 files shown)


RossMurr4y:
As touched on in the description, there is currently a load-order issue with reference data that originates in plugin providers. I will lodge a separate ticket for this and upon its resolution this configuration can be moved entirely into the Azure plugin.

PR #1311, opened 2020-04-28
feature: public login provider support

roleyfoley:
## Description
Enables the ability to configure userpool federation with the different public identity providers ( google, apple, amazon and facebook ) 

## Motivation and Context
Support public identity logins instead of just enterprise level providers ( SAML or OIDC )

## How Has This Been Tested?
Tested with a template generation on a userpool, haven't tested with an app as I didn't have one configured with the providers 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.


ml019:
Thanks for the timely support - I'll test with the customer I'm currently working with.

How are the secrets handled? Are they plain text in the templates? In the config?

PR #1327, opened 2020-05-26
Router component

roleyfoley:
## Description
Adds support for a new router component and support for the gateway to use the router as a destination 

Also includes a fix to the Subnet calculation process, the current process only supports tier counts with perfect square values, this fix uses the closet perfect square for tier counts which don't map to a perfect square. 

## Motivation and Context
The router component is used to provision network traffic between different network components. This allows for private networking and propagated routes to be used when you want to build a mesh based network 

- A router is deployed by itself and by default is not connected to anything 
- A network is deployed by itself and by default is not connected to anything 
- A gateway is deployed to a network based on the tier the gateway belongs to, and its destinations determine which network paths are available to that network. If a gateway is configured with the router engine it needs to link to the router component. This establishes the full path between the components 
- Other gateways can then link to the same route to enable routing between networks 

## How Has This Been Tested?
Tested on local deployment

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.


engine/base.ftl
```diff
@@ -71,11 +71,20 @@
 [/#function]
 
 [#function intSquareRoot value ]
+    [#local lastPerfect = 1 ]
     [#list 1..100 as square]
```
RossMurr4y, line 75: It looks like break directives in lists are [deprecated](https://freemarker.apache.org/docs/ref_directive_list.html#ref.directive.list.break). Should we try and avoid that? It recommends using `?take_while()` instead. Something like:

```
[#list (1..100)?take_while(s -> ((s * s) < value)) as square]
```

engine/base.ftl
```diff
@@ -71,11 +71,20 @@
...
-        [#if square * square == value ]
+        [#local multiple = square * square ]
+        [#if multiple == value ]
             [#return square ]
         [/#if]
+        [#if multiple?floor == multiple ]
```
ml019, line 80: value is an integer so when will multiple?floor not equal multiple?

Am I correct in that the function will now return the square root of the smallest perfect square greater than or equal to the provided value?

engine/base.ftl
roleyfoley: Couldn't get this to work, still seems to be missing the value when it goes to the next one up.

roleyfoley: Couldn't get this to work, still seems to be missing the value when it goes to the next one up.

engine/base.ftl
roleyfoley: Good point, removed the floor check 

Yes the idea is to find the closest perfect square which can then be used for the smallest subnet mask that we need. I've renamed the function and added a comment

engine/base.ftl
RossMurr4y: Not really an issue then, just something I picked out as I hadn't seen a `break` used here before.

PR #1334, opened 2020-06-06
External endpoints

roleyfoley:
## Description
- Adds support for external endpoints, endpoints represent the IP/port of an external service and the ability to set static endpoints for load balancing
- Also adds a utility to provide the usable hosts in a given network 

## Motivation and Context
The intention here is to use external service endpoints in network access control or routing. For example you might want to proxy traffic to an external endpoint through a cloud hosted load balancer. In this case you can use a link from the Load balancer using the StaticEndpoints Links also introduced in this PR to route traffic to this external endpoint. 

 In future work this can also be used to control Egress Security group access so that you can link to an external service and these endpoints will be used for the Access control rules  

## How Has This Been Tested?
Tested on local deployment 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/engine/base.ftl
+++ b/engine/base.ftl
@@ -487,7 +487,14 @@ are added.
 
 [#function getHostsFromNetwork networkCIDR ]
     [#local networkIPs = getSubnetsFromNetwork(networkCIDR, 32)?map(cidr -> (cidr?split("/"))[0] )]
-    [#return networkIPs[1..(networkIPs?size - 2)]]
+
+    [#-- networks with greater than 2 hosts ( i.e. above /31 ) --]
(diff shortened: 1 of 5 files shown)


ml019:
I have a few of these in other projects for things like STS. However the endpoint uses a dns entry rather then IP addresses. They also end up with extra parameters like timeouts and issuer URLs etc.

This subcomponent seems only applicable to networking involving IP addresses. Do we broaden its usage or narrow the name?

roleyfoley:
It was just to represent IP/Ports so that there were explicit parameters available to other components. Not sure you can use Urls as much for specific usage like security groups or load balancers though

PR #1337, opened 2020-06-07
External Service Endpoint attributes

roleyfoley:
## Description
- Set attributes for external service endpoints based on the IP Address Group and Port configuration 
- Minor fix to mark external service endpoints as deployed

## Motivation and Context
When linking to an endpoint it would be useful to get this information as attributes inline with our other components

## How Has This Been Tested?
tested locally

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/providers/shared/components/externalservice/state.ftl
+++ b/providers/shared/components/externalservice/state.ftl
@@ -64,16 +64,32 @@
     [#local core = occurrence.Core]
     [#local solution = occurrence.Configuration.Solution]
 
+    [#local cidrs = getGroupCIDRs(solution.IPAddressGroups, true, occurrence)]
+    [#local hosts = []]
+    [#list cidrs as cidr ]
+        [#local hosts = combineEntities(hosts, getHostsFromNetwork(cidr), UNIQUE_COMBINE_BEHAVIOUR) ]
+    [/#list]
+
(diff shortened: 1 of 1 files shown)


providers/shared/components/externalservice/state.ftl
```diff
@@ -64,16 +64,32 @@
...
+    [#local parentAttributes = parent.State.Attributes ]
+
     [#assign componentState =
         {
             "Resources" : {
                 "externalEndpoint" : {
                     "Id" : formatId(SHARED_EXTERNAL_RESOURCE_TYPE, core.Id),
                     "Name" : core.FullName,
-                    "Type" : SHARED_EXTERNAL_RESOURCE_TYPE
+                    "Type" : SHARED_EXTERNAL_RESOURCE_TYPE,
```
ml019, line 83: Think the type is wrong - ^C^V?

roleyfoley: I left it as an external resource type intentionally, it didn’t feel like it needed a distinct resource when it’s not actually used

ml019: hmm - not sure I'm comfortable with this but I'll commit for mow

PR #1343, opened 2020-06-15
feat(console): Update console support for encryption

roleyfoley:
## Description
Updates the existing console account level unit with the following:
- Enables KMS encryption for SSM Sessions
- Adds support for both S3 and cloudwatch for SSM session log storage 
- Moves to Cloudformation for creating the document now that the Session Document Type is supported 
    - Includes a cleanup script to remove the old documents and to also override manual updates which were made before this was applied 

## Motivation and Context
Session manager is very useful in secure environments where internet access isn't permitted, these environments usually have strong encryption requirements so we need to support these and also support logging to different data stores 

## How Has This Been Tested?
Tested on local deployment

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/legacy/account/account_console.ftl
+++ b/legacy/account/account_console.ftl
@@ -1,80 +1,127 @@
 [#if getDeploymentUnit()?contains("console") || (allDeploymentUnits!false) ]
 
     [#if deploymentSubsetRequired("generationcontract", false)]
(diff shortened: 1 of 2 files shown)


ml019:
So we sometimes still need alternatives...

PR #1349, opened 2020-06-26
Fix: Support legacy types in profiles

ml019:
## Description
Capture the mapping between legacy types and their canonical equivalent.

Rework the component utility functions to drive off the legacy type mapping.

Rework the lookup of profiles during processing to support the use of legacy types. In the process, also make this logic case insensitive.

## Motivation and Context
Legacy types should work consistently across all aspects of hamlet operation. 

## How Has This Been Tested?
Local builds.

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/providers/shared/components/component.ftl
+++ b/providers/shared/components/component.ftl
@@ -1,5 +1,8 @@
 [#ftl]
 
+[#-- Legacy type mappings --]
+[#assign legacyTypeMapping = {} ]
+
 [#-- Known component types --]
 [#assign ADAPTOR_COMPONENT_TYPE = "adaptor"]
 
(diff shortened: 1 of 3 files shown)


providers/shared/components/component.ftl
```diff
@@ -36,6 +39,7 @@
 
 [#assign DB_COMPONENT_TYPE = "db" ]
 [#assign DB_LEGACY_COMPONENT_TYPE = "rds" ]
+[#assign legacyTypeMapping += { DB_LEGACY_COMPONENT_TYPE : DB_COMPONENT_TYPE } ]
```
roleyfoley, line 42: I thought you had this on the Id for the DB Component? Any reason this moved?

roleyfoley: I think I like that its part of where you set the component structure. Makes it managed from a central point then

ml019: Totally agree but unfortunately we need to know its a legacy type to work out its canonical type to then know which component to lazy load. So at the point of analysis, nothing from the component macro has been loaded.

As this file is where all the types live, it felt like the next best spot.

PR #1362, opened 2020-07-25
Deployment groups and Resource Labels

roleyfoley:
## Description
Adds support for Deployment groups and resource labels

**Deployment Groups**
The intention of deployment groups is to provide a blueprint driven approach to the current implementation of levels. Levels imply the ordering of deployment units and outline the general role that the component provides in the solution. 

Deployment groups allows a hamlet user to change the grouping of these components based on their own requirements. The initial implementation just includes the current levels, segment, solution, application, account with future changes allowing for creating custom groups and to automate the ordering of component deployments 

**Resource Labels and ResourceSets** 
Currently each level supports a set of special deployment units which generate templates for specific resources ( iam, lg, cmk, eip ). ResourceSets allows for this to be controlled as part of the DeploymentGroup, the initial implementation just covers the current resource sets with future PR's implementing resource labels on each resource. This would allow for custom resourceSets based on the resource labels defined on resources 

## Motivation and Context
The primary motivation for these changes is to be able generate a list of deployment units and their levels so that we can create templates and deployments for all or multiple components

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.


engine/resourceLabel.ftl
```diff
@@ -0,0 +1,56 @@
+[#ftl]
+
+[#-- tasks are executed in contracts --]
```
ml019, line 3: This looks like  cut/paste leftover?

engine/resourceLabel.ftl
```diff
@@ -0,0 +1,56 @@
...
+[#-- Each task should perform a specifc action to manage a deployment --]
+
+[#assign resourceLabelConfiguration = {}]
+
+[#-- Macros to assemble the component configuration --]
```
ml019, line 8: Same for this comment

providers/shared/resourcelabels/resourcelabel.ftl
```diff
@@ -0,0 +1,11 @@
+[#ftl]
+
+[#-- Core Tasks --]
```
ml019, line 3: Wrong comment?

client/createDeployment.ftl
```diff
@@ -1,37 +1,24 @@
...
+[#-- ResourceSets  --]
+[#-- Seperates resources from their component templates in to their own deployment --]
+[#list ((deploymentGroup.Deployment.ResourceSets)!{})?values?filter(s -> s.Enabled ) as resourceSet ]
+    [#if getDeploymentUnit() == resourceSet.DeploymentUnit ]
 
-[#assign categoryId = "account"]
+        [#assign contractSubsets = []]
+        [#list resourceSet.ResourceLabels as label ]
+            [#assign resourceLabel = getResourceLabel(label, level) ]
+            [#assign contractSubsets = combineEntities( contractSubsets, (resourceLabel.Subsets)![], UNIQUE_COMBINE_BEHAVIOUR ) ]
```
ml019, line 16: Will this return anything given there aren't label defintions for iam, lg etc?

ml019: Need to merge in changes in the aws plugin to match. Reslving conversation

PR #1380, opened 2020-08-26
feat(userpool): control oauth on clients

roleyfoley:
## Description
Allow for OAuth to be disable on userpool clients

## Motivation and Context
In some cases, such as aws identity pools you might want to create a cilent but not include the oAuth configuration. This allows for this to be controlled on a client by client basis.

## How Has This Been Tested?
Tested locally

## Types of changes

## Checklist:
- [X] None of the above.

### DIFF
--- a/providers/shared/components/userpool/id.ftl
+++ b/providers/shared/components/userpool/id.ftl
@@ -232,6 +232,11 @@
                         "Type" : ARRAY_OF_STRING_TYPE,
                         "Values" : [ "code", "implicit", "client_credentials" ],
                         "Default" : [ "code" ]
+                    },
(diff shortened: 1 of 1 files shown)


RossMurr4y:
I believe this is just making the Attribute explicit so that it is more obviously available for use. You should be able to set the attribute `"Enabled": false` on any component/subcomponent right?

RossMurr4y (approved):

ml019:
@RossMurr4y at present yes. The Enabled parameter is injected into any children config where it hasn't explicitly been disabled.

Explicit is probably better as it makes things like documentation easier to generate.

Issue #1196, opened 2020-01-27
Deployment Groups


roleyfoley:
Deployment groups provide the ability to orchestrate the provisioning of components in a solution. They are essentially the replacement of levels and allow for the implicit ordering defined via the manageEnvironment Jenkins jobs to be defined in our CMDB configuration explicitly with the ability to have fine grain control over the provisioning ordering of components

Ordering is defined by each group rather than an explicit list, each group specifies another that it follows along with an configuration option on whether this must always happen. 

This order is processed through the unitlist deployment level added in #1149 which will output the full deployment group order list including the deployment units which belong to each group. 

The execution engine uses this along with command line options it has been provided ( i.e. a specific group or unit ) to execute both template generation and deployment in the provided order 

The execution engine should follow rules when using the unit list to manage deployment units 
- A Deployment group command line option of  `_all` should run generation for units in the list
- A command line option specifying a group should run the generation for that group and also generate any templates for groups specified as always preceding this group 
- If only a dpeloyment unit is specified just the unit should be generated 

Along with controlling ordering of deployments, deployment groups also provide the ability to define a component type qualifier, which essentially incorporates the existing level setup and state routines, A group can have a defined level which is used to generate templates used at different levels of deployment 

Groups can also be nominated as subset groups which is used to control where subset resources are created. This is to accommodate large deployments where the existing subset structure based on levels creates templates which are unsupported by providers ( i.e. AWS stack output limit )

ml019:
@roleyfoley not quite sure what you mant by "component type qualifier, which essentially incorporates the existing level setup and state routines" - mainly in relation to state routines. Is the qualifier the equivalent of the current "subset" concept?

roleyfoley:
Done

PR #1381, opened 2020-08-31
Deployment group filtering

roleyfoley:
## Description
Part of #1196 
Enables the use of DeploymentGroups as part of the deployment generation process. Occurrences can now be configured with a deployment group and this group, combined with the deployment unit will be used to determine if an occurrence can be deployed.

Also sets a default deployment group for all components, the default aligns with the current level based groups for backwards compatibility 

DeploymentGroup can be specified on all top level components, Subcomponents are only enabled for DeploymentGroups if they can be deployed independently from their parent component

## Motivation and Context
This allows for control over when an occurrence can be deployed and to allow for different groupings of components then the current level based filtering

## How Has This Been Tested?
Tested on a local deployment

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.


roleyfoley:
This PR has now been extended to support some additional features 

**Deployment Modes ** 
Provide sequencing for deployments based on deployment groups. Deployment modes specify the groups that are included in the Mode. We have these already but as more of an implied process rather than an explicit definition
This can be defined in two ways
   - Priority - which uses pattern matching of the deployment group id combined with the deployment groups priority attribute to determine the order groups should be executed under. This allows for dyanmic deployment modes which fit common purposes. I.e. a normal update mode that would run across all units in all groups or a delete mode which does the same for deletes. 
   - Explicit - uses an ordered array of group Ids to determine the deployment ordering and which groups are included in the mode 

**Deployment Component attributes namespace** 

This change introduces a new set of attributes for components to control their deployment properties and uses the provider prefixing process to add the attributes under the `deployment` name space. 

This namespace currently includes 

- `deployment:Unit` which replaces the existing `DeploymentUnits` attribute - Backwards compatibility has been included to reduce impact on change 
- `deployment:Group` - The deployment group the occurrence belongs to 
- `deployment:Priority` - The priority of the occurrence within the deployment group. This is used to sequence units within a group 


**Management Contract** 

The management contract is the first output based on these new concepts. For a given deployment Mode the management contract will provide  an ordered list of the deployment groups and their units. This contract can be used to then run testing or run whole deployments using manage Units if required.

providers/shared/components/gateway/id.ftl
```diff
@@ -1,24 +1,27 @@
...
-                "Value" : [ "aws" ]
-            },
-            {
-                "Type" : "ComponentLevel",
-                "Value" : "segment"
             }
         ]
     attributes=
         [
+            {
```
ml019, line 20: Redundant? Covered by addComponentGroup?

providers/shared/components/objectsql/id.ftl
```diff
@@ -1,5 +1,10 @@
 [#ftl]
 
+[@addComponentDeployment
```
ml019, line 3: Should everything have a priority for consistency?

ml019 (changes requested):

providers/shared/components/objectsql/id.ftl
roleyfoley: Since its the default priority instead of the actual priority I thought it was better to leave it. The default is sent in the macro and would be available in the component schema

providers/shared/components/gateway/id.ftl
roleyfoley: Fixed

providers/shared/components/objectsql/id.ftl
ml019: ok - that works

PR #1382, opened 2020-09-02
feat(ecs): add healthcheck support

roleyfoley:
## Description
Adds support for [docker based health checks](https://docs.docker.com/engine/reference/builder/#healthcheck) for tasks and services in the ECS component

## Motivation and Context
This allows for an out of band process to evaluate the health of a container. ECS will use this information to determine if the task container is healthy or not. If it is not healthy then ECS will replace it. This is useful for tasks which aren't part of a load balancer or will always respond to TCP based health checks

## How Has This Been Tested?
Tested on an active deployment using a local install of hamlet

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/engine/commonApplication.ftl
+++ b/engine/commonApplication.ftl
@@ -342,7 +342,7 @@
     ]
 [/#macro]
 
-[#macro HealthCheck command useShell=true interval=30 retries=3 startWait=0 timeout=5  ]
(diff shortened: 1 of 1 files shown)


ml019 (approved):
Neat feature

PR #1383, opened 2020-09-02
Feat apigw throttling

RossMurr4y:
## Description

Refactors the API Gateway components throttling configuration into a simpler structure.
* We don't need the "API" level of rate limiting, as as it is, one "stage" or "version" becomes a unique apigateway instance anyway - so its effectively covered by the "Stage" level rate limiting config.
* The existing attribute "Patterns" on the component can be used for creating the method-level rate limits.
* Also implements throttling based on quota as well - A quota has a limit and time period, and if set, throttling will occur if the limit is exceeded within the time period.

## Motivation and Context

Refactor to make better use of existing attributes. Desire to implement rate limiting based on a quota.

## How Has This Been Tested?

Still requires testing.

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Followup Actions

- [ ] Test

## Checklist:

- [x] None of the above.

### DIFF
--- a/engine/openapi.ftl
+++ b/engine/openapi.ftl
@@ -131,57 +131,6 @@
     }
 ]
 
-[#assign throttlingSettingChildren = 
-    [
-        {
-            "Names" : "BurstLimit",
-            "Description" : "The maximum number of concurrent requests.",
-            "Type" : NUMBER_TYPE
-        },
-        {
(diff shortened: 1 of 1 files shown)


ml019 (approved):
Simplifies things a fair bit

PR #1385, opened 2020-09-03
fix(settings): evaluation preferences for setting namespaces

roleyfoley:
## Description
Sets an explicit preference on how setting name spaces should be evaluated based on their matching patterns

With the following rules: 
- When provided with a list of prefixes the settings from the last prefix will win 
- Within a prefix exact matches are preferred over partial matches

## Motivation and Context
This was found when trying to deploy a component that included a partial match to another component in the solution. As a result of this the exact setting defined on the deployment unit was being overridden by the partial match provided by the other component. 

While it was under a very weird context ( component deployment unit was the same as a tier Id ) this did highlight an assumed ordering of settings that wasn't implemented in the code.

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/engine/engine.ftl
+++ b/engine/engine.ftl
@@ -210,15 +210,10 @@ names.
     [#local matches = [] ]
     [#local candidates = [] ]
 
-    [@debug message=
-        {
-            "Method" : "getMatchingNamespaces",
-            "Namespaces" : namespaces,
-            "Prefixes" : prefixes,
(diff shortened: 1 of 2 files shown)


ml019 (approved):
Reordering of alternatives not needed now but it reflects the preferred order so happy for that to go through.

PR #1388, opened 2020-09-04
New Components: containerhost, containerservice, containertask

RossMurr4y:
## Description

Splits out the current subcomponents of "ECS" (`Service` and `Task`) into their own components. The 3 new components are `containerhost`, `containerservice` and `containertask`.

The current "ECS" component remains as it was, allowing for a gradual transition to the new structure.

## Motivation and Context

Services and Tasks have no relationship to each other and so do not need to co-exist on the same component. The "parent" component (currently "ECS") should define its engine type, and then the components that link to the host (Services and Tasks typically) would then review the type of host they are associated with and create the applicable resources. For any given provider, the Engine's available to a Service are not always the same as those for a Task. Splitting the 3 components out allows them to use component links to determine their behaviour, following established patterns.

## How Has This Been Tested?

Local template generation.

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.


ml019:
@RossMurr4y I think the containerservice and task should draw their implementation from the host they are associated with, not the other way around as described in themotivation?

providers/shared/components/component.ftl
```diff
@@ -965,6 +969,480 @@
...
+        "Names" : "ComputeProvider",
+        "Description" : "The default compute provider for the cluster",
+        "Values" : [ "ec2", "ec2OnDemand" ],
+        "Default" : "ec2"
+    }
+]]
+
+[#assign containerServiceAttributes = [
+    {
+        "Names" : "Engine",
```
ml019, line 1207: I think this should be captured via links not on the component. To switch from fargate to ECS implies the creation of a cluster - something that should be managed on the container host not the service.

providers/shared/components/component.ftl
```diff
@@ -965,6 +969,480 @@
...
+    },
+    {
+        "Names" : "RunMode",
+        "Description" : "A per container setting which can be used by the app to determine run mode for a container in a task - defaults to the second half of a dash separated id",
+        "Type" : STRING_TYPE,
+        "Default" : ""
+    }
+]]
+
+[#assign containerHostAttributes = [
```
ml019, line 1083: Placing these here only makes sense if they will be reused across components. IF they will be shared across providers, then they should be shared...

ml019 (changes requested):

providers/shared/components/component.ftl
RossMurr4y: @ml019 I'm not following. They are being used across both providers and components. They need to be available to the `ecs` component where they originated, plus made available to each of the new "container" components. Down the road where there are no `ecs` components implemented anywhere, we could then moved them directly on to each of the new components.

providers/shared/components/component.ftl
RossMurr4y: Updated in fae79f6  and 5cdc74c

RossMurr4y:
> @RossMurr4y I think the containerservice and task should draw their implementation from the host they are associated with, not the other way around as described in the motivation?

Have updated after our discussions this morning

providers/shared/components/containerhost/id.ftl
```diff
@@ -0,0 +1,32 @@
...
+                "Value" : "An autoscaling container host cluster"
+            },
+            {
+                "Type" : "Providers",
+                "Value" : [ "aws" ]
+            },
+            {
+                "Type" : "ComponentLevel",
+                "Value" : "solution"
+            }
```
roleyfoley, line 23: These are no longer required and with the move to deployment groups and placements we probably don't need them anymore

providers/shared/components/containerservice/id.ftl
```diff
@@ -0,0 +1,26 @@
...
+                "Value" : "An orchestrated container with always on scheduling"
+            },
+            {
+                "Type" : "Providers",
+                "Value" : [ "aws" ]
+            },
+            {
+                "Type" : "ComponentLevel",
+                "Value" : "application"
+            }
```
roleyfoley, line 23: Same here

providers/shared/components/containertask/id.ftl
```diff
@@ -0,0 +1,26 @@
...
+                "Value" : "A container defintion which is invoked on demand"
+            },
+            {
+                "Type" : "Providers",
+                "Value" : [ "aws" ]
+            },
+            {
+                "Type" : "ComponentLevel",
+                "Value" : "application"
+            }
```
roleyfoley, line 23: and here

roleyfoley (changes requested):

providers/shared/components/component.ftl
RossMurr4y: @ml019 are you able to elaborate on what you need updated here?

providers/shared/components/containerhost/id.ftl
RossMurr4y: Cheers, have removed in e98df10

providers/shared/components/containerservice/id.ftl
RossMurr4y: Removed

providers/shared/components/containertask/id.ftl
RossMurr4y: Removed

RossMurr4y:
Updated with requested changes and rebased.

providers/shared/components/component.ftl
ml019: IF being shared across components/providers, then here is fine.

ml019: Now I remember, these look to be things that should be configured in the id file of the shared provider, not included here.

The sharing across providers should be via the shared implementation not here?

ml019: OK - so its a transition thing. Once we kill ECS, these definitions should move back to the respective id files - yes?

ml019 (approved):

PR #1391, opened 2020-09-09
Feat - ResourceGroup Attribute Extensions

RossMurr4y:
## Description

Extends the global `componentConfiguration` object to allow for "Extensions". Extensions provide additional valid values to an existing shared provider attribute. 

Extensions are not merged into the current structure but remain separate until after the placement is determined for the current occurrence, and the correct extensions to use for a given occurrence is determined by the placement.

<details>
<summary>Testing Extending the Value Sets for the DB Component</summary>
<br>

```
[@addResourceGroupAttributeValues
    type=DB_COMPONENT_TYPE
    provider=AZURE_PROVIDER
    extensions=[
        {
            "Names" : "Backup",
            "Children" : [
                {
                    "Names" : "DeletionPolicy",
                    "Values" : [ "notadeletionpolicy", "notreal"]
                }
            ]            
        },
        {
            "Names" : "Cluster",
            "Children" : [
                {
                    "Names" : "ScalingPolicies",
                    "Children" : [
                        {
                            "Names" : "Type",
                            "Values" : [ "Slippery-Dip" ]
                        }
                    ]
                }
            ]
        }
    ]
/]
```

</details>

<details>
<summary>Results for Backup.DeletionPolicy (direct children)</summary>
<br>

```
       {
          "Names": "Backup",
          "Children": [
            ...
            {
              "Names": "DeletionPolicy",
              "Type": "string",
              "Values": [
                "Snapshot",
                "Delete",
                "Retain",
                "azure:notadeletionpolicy",
                "azure:notreal"
              ],
              "Default": "Snapshot"
            }
            ...
          ]
        }
```

</details>

<details>
<summary>Results for Cluster.ScalingPolicies.Type (SubObjects = true) with </summary>
<br>

```
        {
          "Names": "Cluster",
          "Description": "Cluster specific configuration when using a clustered database engine",
          "Children": [
            {
              "Names": "ScalingPolicies",
              "Subobjects": true,
              "Children": [
                {
                  "Names": "Type",
                  "Type": "string",
                  "Values": [
                    "Stepped",
                    "Tracked",
                    "Scheduled",
                    "azure:Slippery-Dip"
                  ],
                  "Default": "Stepped"
                }
               ...
```

</details>

## Motivation and Context

A number of shared provider attributes such as "Engine" are common in their usage across providers, but the set of valid values for that attribute can change per provider. 

It's currently necessary for the plugins to use namespaced attributes to accommodate for this, however this causes some unnecessary duplication of attributes - a namespaced attribute would appear as `"provider:Attribute" : value` alongside the shared `"Attribute" : value` in documentation and could cause confusion as to which one should be used.

## How Has This Been Tested?

Local template generation

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/engine/component.ftl
+++ b/engine/component.ftl
@@ -137,7 +137,7 @@
     }
 ] ]
 
-[#macro addResourceGroupInformationExtensions type extensions provider]
+[#macro addResourceGroupAttributeValues type extensions provider]
 
     [@internalMergeComponentConfiguration
         type=type
--- a/engine/component.ftl
+++ b/engine/component.ftl
@@ -152,7 +152,7 @@
                                     profileAttribute.Children +
                                     attribute.Children
                             }
-                        ] ]
+                        ]
                 [#else]
                     [#local providerAttributes += [attribute] ]
                 [/#if]
--- a/engine/component.ftl
+++ b/engine/component.ftl
@@ -137,6 +137,25 @@
     }
 ] ]
 
+[#macro addResourceGroupInformationExtensions type extensions provider]
+
+    [@internalMergeComponentConfiguration
+        type=type
+        configuration=
+            {
+                "ResourceGroups" : {
+                    DEFAULT_RESOURCE_GROUP : {
+                        "Extensions" : {
+                            provider :
+                                extensions
+                        }
+                    }
+                }
+            }
+    /]
+
+[/#macro]
+
 [#macro addResourceGroupInformation type attributes provider resourceGroup services=[] prefixed=true ]
     [#if provider == SHARED_ATTRIBUTES ]
         [#-- Special processing for profiles --]
--- a/engine/component.ftl
+++ b/engine/component.ftl
@@ -137,14 +137,14 @@
     }
 ] ]
 
-[#macro addResourceGroupAttributeValues type extensions provider]
+[#macro addResourceGroupAttributeValues type extensions provider resourceGroup=DEFAULT_RESOURCE_GROUP]
 
     [@internalMergeComponentConfiguration
         type=type
         configuration=
             {
                 "ResourceGroups" : {
-                    DEFAULT_RESOURCE_GROUP : {
+                    resourceGroup : {
                         "Extensions" : {
                             provider :
                                 extensions
--- a/engine/occurrence.ftl
+++ b/engine/occurrence.ftl
@@ -209,15 +209,27 @@
         [#local providerExtensions = extensions[provider]![]]
         [#list attributes as attribute]
 
-            [#local attributeExtension = providerExtensions?filter(e -> asArray(attribute.Names)?seq_contains(e.Names))]
-            [#if attributeExtension?has_content]
-  
-                [#local extendedValues = combineEntities(
-                    attribute.Values,
-                    attributeExtension[0].Values,
-                    ADD_COMBINE_BEHAVIOUR) ]
+            [#local attributeExtension = 
+                providerExtensions
+                ?filter(e -> asArray(attribute.Names)
+                    ?seq_contains(e.Names))]
 
-                [#local result += [mergeObjects(attribute, { "Values" : extendedValues })] ]
+            [#if attributeExtension?has_content]
+
+                [#local extendedValues = attribute.Values]
+  
+                [#list attributeExtension[0].Values as extensionValue]
+                    [#local extendedValues +=
+                        [extensionValue?ensure_starts_with(provider + ":")] ]
+                [/#list]
+
+                [#local result += [
+                    mergeObjects(
+                        attribute,
+                        {
+                            "Values" : getUniqueArrayElements(extendedValues)
+                        }
+                    )]]
             [#else]
                 [#local result += [attribute]]
             [/#if]
--- a/engine/occurrence.ftl
+++ b/engine/occurrence.ftl
@@ -203,33 +203,55 @@
     [#return occurrence]
 [/#function]
 
-[#function extendOccurrenceAttributes provider attributes=[] extensions=[]]
+[#function extendAttributes attributes=[] extensions=[] prefix=""]
     [#local result = []]
     [#if extensions?has_content]
-        [#local providerExtensions = extensions[provider]![]]
         [#list attributes as attribute]
 
             [#local attributeExtension = 
-                providerExtensions
+                extensions
                 ?filter(e -> asArray(attribute.Names)
-                    ?seq_contains(e.Names))]
+                ?seq_contains(e.Names))![]]
 
             [#if attributeExtension?has_content]
+                [#local attributeExtension = attributeExtension?first]
+                [#if (attribute.Children![])?has_content]
+                    [#if (attributeExtension.Children![])?has_content]
+                        [#local result += [
+                            mergeObjects(
+                                attribute,
+                                {
+                                    "Children" : extendAttributes(
+                                                    attribute.Children,
+                                                    attributeExtension.Children,
+                                                    prefix)
+                                }
+                            )
+                        ]]
+                    [#else]
+                        [@fatal
+                            message="Attribute Extension missing children."
+                            context={
+                                "Attribute" : attribute,
+                                "Extension" : attributeExtension
+                            }
+                        /]
+                    [/#if]
+                [#else]
+                    [#local extendedValues = attribute.Values]
+                    [#list attributeExtension.Values as extensionValue]
+                        [#local extendedValues +=
+                            [extensionValue?ensure_starts_with(prefix + ":")] ]
+                    [/#list]
 
-                [#local extendedValues = attribute.Values]
-  
-                [#list attributeExtension[0].Values as extensionValue]
-                    [#local extendedValues +=
-                        [extensionValue?ensure_starts_with(provider + ":")] ]
-                [/#list]
-
-                [#local result += [
-                    mergeObjects(
-                        attribute,
-                        {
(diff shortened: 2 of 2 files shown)


engine/occurrence.ftl
```diff
@@ -203,13 +203,42 @@
...
+[#function extendOccurrenceAttributes provider attributes=[] extensions=[]]
+    [#local result = []]
+    [#if extensions?has_content]
+        [#local providerExtensions = extensions[provider]![]]
+        [#list attributes as attribute]
+
+            [#local attributeExtension = providerExtensions?filter(e -> asArray(attribute.Names)?seq_contains(e.Names))]
+            [#if attributeExtension?has_content]
+  
+                [#local extendedValues = combineEntities(
```
roleyfoley, line 215: I think we had discussed prefixing the attribute when it was being added. 

So for your example, the required value would be `azure:examplevalue` instead of `examplevalue`

I think this provides better understanding as to where the value came from, especially if I was to copy a component definition from one solution to another where the new solution was aws based instead of azure and it started throwing an error that the value provided was incorrect

engine/component.ftl
```diff
@@ -137,6 +137,25 @@
     }
 ] ]
 
+[#macro addResourceGroupInformationExtensions type extensions provider]
```
roleyfoley, line 140: Not sure the name quite portrays what is happening here. Maybe `addResourceGroupAttributeValues` or something like that. At the moment we are introducing a new concept, extensions which we haven't used anywehere else. Where attributes are know and values is what we are changing

roleyfoley (changes requested):

engine/component.ftl
RossMurr4y: Yep agreed, I kept changing this around and figgured it would be better to collaborate on a name. I've implemented your recommendation in cb51fca

engine/occurrence.ftl
RossMurr4y: You're right, we had talked about that! I have added it in cb51fca and testing shows it working as intended

```json
{
  "Timestamp": "2020-09-10T00:44:22.984Z",
  "Severity": "fatal",
  "Message": "Attribute value is not one of the expected values",
  "Context": {
    "Name": "Role",
    "Value": "mockprovider:examplevalue",
    "ExpectedValues": [
      "appdata",
      "operations",
      "staticWebsite",
      "azure:examplevalue"
    ]
 ...
}
```

engine/occurrence.ftl
```diff
@@ -203,13 +203,54 @@
...
+    [#if extensions?has_content]
+        [#local providerExtensions = extensions[provider]![]]
+        [#list attributes as attribute]
+
+            [#local attributeExtension = 
+                providerExtensions
+                ?filter(e -> asArray(attribute.Names)
+                    ?seq_contains(e.Names))]
+
+            [#if attributeExtension?has_content]
```
roleyfoley, line 217: Will this support nested attributes? I.e. Children attributes?

RossMurr4y: You're right, it wont.

I might move the Extensions on to each attribute directly, as it will allow me to re-use the recursion already in place for this purpose on the  function `getCompositeObject()`

Example:

<details>
<summary>The Component definition currently constructed</summary>
<br>

```json
{
  "ResourceGroups" : {
    "DEFAULT_RESOURCE_GROUP" : {
      "Attributes" : [
       ...
      ],
      "Extensions" : [
        "azure" :  {
          ... azureextensions ...
        }
      ]
    }
  }
}
```

</details>

<details>
<summary>Extensions assigned to each attribute</summary>
<br>

```json
{
  "ResourceGroups" : {
    "DEFAULT_RESOURCE_GROUP" : {
      "Attributes" : [
        "engine" {
          "Names" : [ "Engine" ],
          "Type" : "string",
          "Values" : [ "engine1", "engine2" ],
          "Extensions" : [
            {
               "azure" : [ "engine3", "engine4" ]
            }
          ]
        }
      ]
    }
  }
}
```

</details>

And then I could move the current extension value logic into `getCompositeObject()` to perform the merge of values.

<details>
<summary>The final composite object</summary>
<br>

```json
{
  "ResourceGroups" : {
    "DEFAULT_RESOURCE_GROUP" : {
      "Attributes" : [
        "engine" {
          "Names" : [ "Engine" ],
          "Type" : "string",
          "Values" : [ "engine1", "engine2", "azure:engine3", "azure:engine4" ],
          "Extensions" : [
            {
               "azure" : [ "engine3", "engine4" ]
            }
          ]
        }
      ]
    }
  }
}
```

</details>

If I keep the "Extensions" at the same level of "Attributes", this will require updating `getCompositeObject()` too much. As the recursion drills down on the child attributes, it would likewise need to drill down on the child extensions object, which would mean new function parameters.

@ml019 @roleyfoley any concerns with this?

ml019: getCompositeObject isn't the right place to do this. It expects a list of attributes full stop.

Any nested attribute values need to be added before the call to getCompositeObject.

I'd suggest putting the shared extensions under the provider specific section of componentConfiguration - they logically belong with the provider not the shared implementation.

RossMurr4y: I have reworked this to allow for for child attributes & have tested it as follows:

<details>
<summary>Testing Extending the Value Sets for the DB Component</summary>
<br>

```
[@addResourceGroupAttributeValues
    type=DB_COMPONENT_TYPE
    provider=AZURE_PROVIDER
    extensions=[
        {
            "Names" : "Backup",
            "Children" : [
                {
                    "Names" : "DeletionPolicy",
                    "Values" : [ "notadeletionpolicy", "notreal"]
                }
            ]            
        },
        {
            "Names" : "Cluster",
            "Children" : [
                {
                    "Names" : "ScalingPolicies",
                    "Children" : [
                        {
                            "Names" : "Type",
                            "Values" : [ "Slippery-Dip" ]
                        }
                    ]
                }
            ]
        }
    ]
/]
```

</details>


<details>
<summary>Results for Backup.DeletionPolicy (direct children)</summary>
<br>

```
       {
          "Names": "Backup",
          "Children": [
            ...
            {
              "Names": "DeletionPolicy",
              "Type": "string",
              "Values": [
                "Snapshot",
                "Delete",
                "Retain",
                "azure:notadeletionpolicy",
                "azure:notreal"
              ],
              "Default": "Snapshot"
            }
            ...
          ]
        }
```

</details>

<details>
<summary>Results for Cluster.ScalingPolicies.Type (SubObjects = true) with </summary>
<br>

```
        {
          "Names": "Cluster",
          "Description": "Cluster specific configuration when using a clustered database engine",
          "Children": [
            {
              "Names": "ScalingPolicies",
              "Subobjects": true,
              "Children": [
                {
                  "Names": "Type",
                  "Type": "string",
                  "Values": [
                    "Stepped",
                    "Tracked",
                    "Scheduled",
                    "azure:Slippery-Dip"
                  ],
                  "Default": "Stepped"
                }
               ...
```

</details>

RossMurr4y: also, I've updated the extension function to simply `extendAttributes()` as it should also work for Reference Data, although this was out of scope and is untested.

ml019: Good thought re reference data - so we have a likely use case for that?

ml019:
@RossMurr4y mabe correct the example in the PR description to match later comments in case we come back to the 
PR at some stage - easier if description is current.

engine/component.ftl
```diff
@@ -137,6 +137,25 @@
...
+
+    [@internalMergeComponentConfiguration
+        type=type
+        configuration=
+            {
+                "ResourceGroups" : {
+                    DEFAULT_RESOURCE_GROUP : {
+                        "Extensions" : {
+                            provider :
+                                extensions
```
ml019, line 150: There can be shared attributes for any resource group so I think this function needs an extra attribute? If you look at the modified code where the attributes are assembled, it loops through the resource groups.

ml019 (changes requested):

engine/component.ftl
RossMurr4y: Cheers, updated in 1669096

engine/occurrence.ftl
RossMurr4y: @ml019 not quite, but I think we could actually do a fair bit with the idea. 

For example, all of the ARRAY_OF_OBJECT_TYPE attributes are Reference Data. For things like the "SKUProfile" (the combination of hardware make/model/unit used by many Azure resources, but also common hardware term in general) I've created in the Azure Provider - its not a unique-to-azure concept, its just not used by the AWS provider. Instead of having such things in the plugins we could flesh out their definition (much like we would with a new component type) so its an understood, and then expect providers that use it to extend its value sets with arrays of object types. As it stands there is little reason to do this however, but in a world with more plugins I think this could help keep definitions and terms at a "solution" and common-ground conceptual level.

The other use-case I had thought of was removing things like "Regions" from masterdata, and having them implemented as ARRAY_OF_OBJECT_TYPE by each provider. Both approaches have their pros/cons but I know we've discussed getting rid of masterdata down the road.

RossMurr4y:
> @RossMurr4y mabe correct the example in the PR description to match later comments in case we come back to the
> PR at some stage - easier if description is current.

Done :)

ml019 (approved):
I think this is good to go.

roleyfoley:
Nice! This is looking good now

roleyfoley (approved):

PR #1393, opened 2020-09-13
feat: aws ecs account settings

roleyfoley:
## Description
Implements https://github.com/hamlet-io/engine-plugin-aws/issues/123

Provides support for setting ECS account level settings which control account ( region ) wide services

AWS Details on the available services are available here: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html 

As the settings are reasonably safe in terms of hamlet deployments we have chosen to enable all settings with the exception of container Insights which has a cost implication and can be controlled at the cluster level. 

Names of the settings that we support are based on the AWS cli definition of the settings - https://docs.aws.amazon.com/cli/latest/reference/ecs/put-account-setting.html 

## Motivation and Context
AWS has had to introduce new functionality and potentially breaking changes to the ECS service. To ensure smooth transitions and minimal impact some of the changes have been introduced through an opt in/opt out process and this has been implemented as account settings. Managing these through hamlet allows us to assist with smooth transitions to these new services with minimal impact

## How Has This Been Tested?
Tested on an active deployment, locally

## Types of changes

## Followup Actions
- engine - https://github.com/hamlet-io/engine/pull/1393
- executor-bash - https://github.com/hamlet-io/executor-bash/pull/85

## Checklist:
- [X] None of the above.

### DIFF
--- a/legacy/account/account_ecs.ftl
+++ b/legacy/account/account_ecs.ftl
@@ -1,0 +1,43 @@
+[#-- ECS Account Settings --]
+[#if getDeploymentUnit()?contains("ecs") || (allDeploymentUnits!false) ]
+    [#if deploymentSubsetRequired("generationcontract", false)]
+        [@addDefaultGenerationContract subsets="epilogue" /]
+    [/#if]
+
+    [@includeServicesConfiguration
+        provider=AWS_PROVIDER
+        services=[ ]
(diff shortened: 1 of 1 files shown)


ml019 (approved):
Follows structure for account templates

PR #1408, opened 2020-10-06
Link Direction Made Case Insensitive

RossMurr4y:
## Description

Link processing functions to force all values lower-case.

## Motivation and Context

The requirement for lowercase is presently undocumented and unnecessary. Otherwise correctly-formatted links do not trigger errors, so template passes succeed, however they will be missing the intended permissions.

## How Has This Been Tested?

Local template generation and deployment.

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/providers/shared/deploymentframeworks/default/legacy.ftl
+++ b/providers/shared/deploymentframeworks/default/legacy.ftl
@@ -146,7 +146,7 @@
         [#return
             internalCreateOccurrenceFromExternalLink(occurrence, link) +
             {
-                "Direction" : link.Direction!"outbound",
+                "Direction" : (link.Direction?lower_case)!"outbound",
                 "Role" : link.Role!"external",
                 "IncludeInContext" : link.IncludeInContext![]
             }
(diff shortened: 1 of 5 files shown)


providers/shared/deploymentframeworks/default/legacy.ftl
```diff
@@ -233,15 +233,15 @@
...
+            [#local direction = (link.Direction?lower_case)!"outbound"]
 
             [#local role =
                 link.Role!getOccurrenceDefaultRole(targetSubOccurrence, direction)]
 
             [#return
                 targetSubOccurrence +
                 {
-                    "Direction" : direction,
+                    "Direction" : direction?lower_case,
```
ml019, line 244: Unnecessary given change above

ml019 (changes requested):

providers/shared/deploymentframeworks/default/legacy.ftl
RossMurr4y: Thanks @ml019 , have updated in 57c7e40

ml019 (approved):

PR #1409, opened 2020-10-08
feat: add support for token validity expressions

ml019:
## Description
For a token based authorizer, allow the integration configuration to include toekn validity expressions.

## Motivation and Context
Try and do as much validation via the api gateway as possible.

## How Has This Been Tested?
Client deployment

## Types of changes

## Followup Actions
- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/engine/openapi.ftl
+++ b/engine/openapi.ftl
@@ -173,11 +173,13 @@
                         },
                         {
                             "Names" : "TTL",
-                            "Type" : NUMBER_TYPE
+                            "Type" : NUMBER_TYPE,
+                            "Default" : 300
                         },
                         {
                             "Names" : "ValidityExpression",
-                            "Type" : STRING_TYPE
+                            "Type" : STRING_TYPE,
+                            "Default" : r"^[ \t]*bearer[ \t]+[a-zA-Z0-9\-_]+\.[a-zA-Z0-9\-_]+\.[a-zA-Z0-9\-_]+[ \t]*$"
                         },
                         {
                             "Names" : "Arns",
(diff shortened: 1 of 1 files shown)


engine/openapi.ftl
```diff
@@ -504,14 +508,16 @@
...
                         [#case "token"]
                             [#local scheme +=
                                 {
                                     "x-amazon-apigateway-authorizer": {
                                         "type": "token",
                                         "authorizerUri" : formatLambdaArnUsingStageVariable(context, Authorizer.Variable),
-                                        "authorizerResultTtlInSeconds" : Authorizer.TTL ! 300
-                                    }
+                                        "authorizerResultTtlInSeconds" : Authorizer.TTL ! 300,
+                                        "identityValidationExpression" : Authorizer.ValidityExpression ! r"^[ \t]*bearer[ \t]+[A-Za-z0-9+\/=]+\.[A-Za-z0-9+\/=]+\.[A-Za-z0-9+\/=]+[ \t]*$"
```
roleyfoley, line 519: Shoudn't these be defaults on the configuration schema above instead of hard coded? We can't document them where they are now

roleyfoley (changes requested):

engine/openapi.ftl
ml019: Fair point - I just copied what was done for the TTL. If I add a default to the schema, then it will always set the attribute which is probably advantageous anyway

ml019: Note that these settings are on the integrations file so atm, it won't show up in the component configuration. Maybe we should move this to a reference of some type.

roleyfoley (approved):

PR #1417, opened 2020-10-14
Feature - APIGateway LogToOpsData

RossMurr4y:
## Description

New attribute on the `apigateway` component - LogToOpsData. A boolean value that defines whether logs are directed to the OpsData bucket or not.

## Motivation and Context

Need to be able to consolidate logs somewhere. This allows the logs to be placed in a known location (ops data) where other profiles can handle the replication of data elsewhere if required.

* Ultimately the LogProfiles should be able to define such settings however currently they do not. Further discussion around how logging profiles should look need to be had before that could be implemented however, so this attribute is an interim measure.

## How Has This Been Tested?

Local template generation and deployment (alongside matching changes in the AWS provider plugin).

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/providers/shared/components/apigateway/id.ftl
+++ b/providers/shared/components/apigateway/id.ftl
@@ -225,7 +225,8 @@ object.
             {
                 "Names" : "LogStore",
                 "Description" : "The logging destination for the API Gateway.",
-                "Type" : STRING_TYPE
+                "Type" : STRING_TYPE,
+                "Values" : []
             },
             {
                 "Names" : "BasePathBehaviour",
--- a/providers/shared/components/apigateway/id.ftl
+++ b/providers/shared/components/apigateway/id.ftl
@@ -223,6 +223,11 @@ object.
                 "Children" : logMetricChildrenConfiguration
             },
             {
+                "Names" : "LogStore",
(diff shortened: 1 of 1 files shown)


providers/shared/components/apigateway/id.ftl
```diff
@@ -222,6 +222,11 @@ object.
                 "Subobjects" : true,
                 "Children" : logMetricChildrenConfiguration
             },
+            {
+                "Names" : "LogStore",
+                "Description" : "The logging destination for the API Gateway.",
+                "Type" : STRING_TYPE
+            },
```
roleyfoley, line 229: What is the `destination` defined as? Is this the link Id? a specific type?

roleyfoley (changes requested):

RossMurr4y:
@roleyfoley  I have re-implemented this new attribute to specifically direct the logs to OpsData. 

Whilst I don't love the attribute name, I intended for this to only be implemented until we can have a meeting with @ml019 and discuss loggingProfiles further.

ml019 (approved):
Based on this being subject to further refinement, no issue in merging.

PR #1432, opened 2020-10-21
feat: support for centralised service resources

roleyfoley:
## Description
Adds support for a centralised definition of services and their resources configuration. The main functionality is the introduction of the `serviceResourceConfiguration` hash which accepts updates to sections of data within the structure based on the Provider, Service and Resource that you want to provide details about 

Sections allow for multiple sources of data to be added to a given resource, for example output mappings and diagram class mappings. This also allows for other providers to extend existing resources and services which might be implemented by another provider 

With the data in this structure it is also possible to search for the provider and service that a resource belongs to based on a first returned basis ( some resource names currently overlap within the same provider ). 

## Motivation and Context
Currently we implement some mappings for resources ( output mappings ) and we have defined that services belong to resources through our dynamic loading process, but once they have been loaded its not possible to determine a given resources service or its provider. This allows us to complete these kinds of queries and to also maintain a central point of resource and service configuration when multiple providers can offer configuration for a given resource 

## How Has This Been Tested?
Tested locally

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/engine/bootstrap.ftl
+++ b/engine/bootstrap.ftl
@@ -35,6 +35,9 @@
 [#-- Component handling --]
 [#include "component.ftl" ]
 
+[#-- Service Handling --]
+[#include "services.ftl" ]
+
 [#-- View handling --]
 [#include "view.ftl" ]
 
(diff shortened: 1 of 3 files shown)


ml019 (approved):
In pattern with our other config macros.

PR #1435, opened 2020-10-26
feat(apigw + waf): separate logging enabled from logging profiles

RossMurr4y:
## Description

* Separated out the enabling of logging from the logging profile for the `apigateway` component and `WAF` configurations.

Note: I've used objects for the `apigateway` attributes because the intent is for providers to provide any further control they require at the provider level. A bool would not have been sufficient control here but any further configuration here will be provider specific.

## Motivation and Context

The APIGateway provides several kinds of logs - Access Logs, Execution Logs and also WAF Logs. However they are not always able to be configured the same way. Separating out the concerns of "do you want logs?" from "where do you want them?" (`LoggingProfiles`) and creating an attribute structure likes this allows providers to add further control should they wish through namespaced attributes, whilst remaining clear as to the intent of each attribute.

## How Has This Been Tested?

Local template testing.

## Types of changes

## Followup Actions

- [ ] Schema regeneration will be required.

## Checklist:

- [x] My change requires a change to the [documentation](https://github.com/hamlet-io/docs).
- [ ] None of the above.

### DIFF
--- a/providers/shared/components/apigateway/id.ftl
+++ b/providers/shared/components/apigateway/id.ftl
@@ -79,17 +79,6 @@ object.
                 "Names" : "Logging",
                 "Children" : [
                     {
-                        "Names" : "Execution",
-                        "Children" : [
-                            {
-                                "Names" : "Enabled",
-                                "Description" : "Logging of the API Gateway Execution Logs.",
-                                "Type" : BOOLEAN_TYPE,
-                                "Default" : false
-                            }
-                        ]
-                    },
-                    {
                         "Names" : "Access",
                         "Children" : [
                             {
--- a/providers/shared/components/apigateway/id.ftl
+++ b/providers/shared/components/apigateway/id.ftl
@@ -76,18 +76,13 @@ object.
                 "Default" : ""
             },
             {
-                "Names" : "Logging",
+                "Names" : "AccessLogging",
                 "Children" : [
                     {
-                        "Names" : "Access",
-                        "Children" : [
-                            {
-                                "Names" : "Enabled",
-                                "Description" : "Logging of the API Gateway Access Logs.",
-                                "Type" : BOOLEAN_TYPE,
-                                "Default" : true
-                            }
-                        ]
+                        "Names" : "Enabled",
+                        "Description" : "Logging of the API Gateway Access Logs.",
+                        "Type" : BOOLEAN_TYPE,
+                        "Default" : true
                     }
                 ]
             },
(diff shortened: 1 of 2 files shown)


providers/shared/components/apigateway/id.ftl
```diff
@@ -75,6 +75,33 @@ object.
...
+                        "Names" : "Execution",
+                        "Children" : [
+                            {
+                                "Names" : "Enabled",
+                                "Description" : "Logging of the API Gateway Execution Logs.",
+                                "Type" : BOOLEAN_TYPE,
+                                "Default" : false
+                            }
+                        ]
+                    },
```
roleyfoley, line 91: I don't think we should include this one. The execution log is a troubleshooting log and should be mandatory for all deployments. The access logging is optional as that can have significant cost implications

providers/shared/components/component.ftl
```diff
@@ -592,11 +592,18 @@
             "Type" : BOOLEAN_TYPE,
             "Default" : false
         },
+        {
+            "Names" : "EnableLogging",
+            "Description" : "WAF Logs in the OpsData DataBucket.",
+            "Type" : BOOLEAN_TYPE,
+            "Default" : true
+        },
```
roleyfoley, line 600: To keep it constistent with the changes above should this be 

```suggestion
        {
            "Names" : "Logging",
            "Children" : [ 
                {
                    "Names" : "Enabled",
                    "Type" : BOOLEAN_TYPE,
                    "Default" : true
                }
            ]
        },
```

providers/shared/components/apigateway/id.ftl
roleyfoley: If this is removed you could also make it an Object shorter and make it 

AccessLogging with children of Enabled

roleyfoley (changes requested):

providers/shared/components/apigateway/id.ftl
RossMurr4y: Thanks, updated in b973911 and 8bff92b

providers/shared/components/component.ftl
RossMurr4y: Updated in bc14a90

RossMurr4y: @roleyfoley should we also update the `EnableLogging` attr on `apigateway` within the `cloudfront` attribute (i.e `solution.CloudFront.EnableLogging` to `solution.CloudFront.Logging.Enabled` and similarly the one on the CDN component?

roleyfoley (approved):

PR #1436, opened 2020-10-26
Occurrence Updates

roleyfoley:
## Description
Adds a couple of changes to the occurrence and its processing 
- Adds an occurrence cache to reduce the repetition of occurrence building during link processing  
- Moves the setting namespace configuration for the occurrence into it's own section of the occurrence. This keeps the settings lookups consistent and also allows them to be visible to users without having to modify the hamlet code itself ( with debug statements ) 

## Motivation and Context
Occurrence caching is for performance reasons 
SettingNamespace is for better user understanding

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/engine/occurrence.ftl
+++ b/engine/occurrence.ftl
@@ -611,6 +611,7 @@
 
 [#function internalCreateOccurrenceBuildSettings occurrence]
     [#local namespaces = occurrence.Configuration.SettingNamespaces ]
+    [#local deploymentUnit = getOccurrenceDeploymentUnit(occurrence) ]
 
     [#local occurrenceBuild =
         internalCreateOccurrenceSettings(
--- a/engine/occurrence.ftl
+++ b/engine/occurrence.ftl
@@ -4,6 +4,47 @@
 -- Public functions for occurrence processing --
 ------------------------------------------------]
 
+
+[#-- Occurrence Caching controls - Provides a collection of occurrences to save the processing time in looking up the occurrence details --]
+[#assign occurrenceCache = {}]
+
+[#macro addOccurrenceToCache tierId componentId subComponentId instanceId versionId occurrence ]
+    [@internalMergeOccurrenceCache
+        tierId=tierId
+        componentId=componentId
+        subComponentId=subComponentId
+        instanceId=instanceId
+        versionId=versionId
+        occurrence=occurrence
+    /]
+[/#macro]
+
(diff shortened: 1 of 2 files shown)


engine/occurrence.ftl
```diff
@@ -461,6 +515,25 @@
     [#return result]
 [/#function]
 
+[#function internalGetOccurrenceSettingNamepsaces occurrence ]
```
RossMurr4y, line 518: typo here in `Namespaces`

engine/occurrence.ftl
```diff
@@ -167,6 +208,19 @@
             }
         ) ]
 
+    [#-- define the setting namespaces for the component --]
+    [#local occurrence =
+        mergeObjects(
+            occurrence,
+            {
+                "Configuration" : {
+                    "SettingNamespaces" : internalGetOccurrenceSettingNamepsaces(occurrence)
```
RossMurr4y, line 217: Type here in `Namespaces`, copied from the function definition I'd say

RossMurr4y:
Looks good, should really speed things up.

Had a couple questions that I couldn't see covered by the code:

1) How does this handle an existing cache but altered solution? If I update my solution I'd expect it not to retrieve the solution from cache, or perhaps the cache to be flushed because I updated the solution.

2) With the cache now in the user home directory, during CI builds, wont the cache always exist past the deployment unit's first run?

engine/occurrence.ftl
roleyfoley: Fixed

engine/occurrence.ftl
roleyfoley: Fixed

roleyfoley:
> Looks good, should really speed things up.
> 
> Had a couple questions that I couldn't see covered by the code:
> 
>     1. How does this handle an existing cache but altered solution? If I update my solution I'd expect it not to retrieve the solution from cache, or perhaps the cache to be flushed because I updated the solution.
> 
>     2. With the cache now in the user home directory, during CI builds, wont the cache always exist past the deployment unit's first run?

This doesn't cache outside of a given freemarker pass, it only caches occurrences which are called in the same pass multiple times 

Caching to a file would require us to be able to write multiple files at once. Something we are working on as part of the freemarker wrapper. That would be a significant performance increase over what is there now

ml019 (approved):
No Issues. Making namespaces part of occurrence is a nice optimisation.

PR #1439, opened 2020-10-27
Scenarios as part of blueprint

roleyfoley:
## Description
Adds support for loading scenarios as part of blueprint file and to provide documentation and validation of scenario configuration with a collection of changes that provide a flexible approach to scenarios while also supporting existing functionality 

- `Scenarios` now need to be defined using the `addScenario` macro. This defines the name, provider ( of the scenario rather than the components ) and a collection of properties. The properties provided will be evaluated through our standard CompositeObject processing and then passed to the macro using the freemarker [with_args](https://freemarker.apache.org/docs/ref_builtins_expert.html#ref_builtin_with_args) builtin 

- `ScenarioProfiles` define a collection of Scenarios and the properties/parameter values for a given instance of a scenario. When adding a scenario to a profile the user needs to define the name and provider of the scenario. This  makes it easy to understand where the scenario came from as they are quite powerful in how they can modify a solution 

- Scenario Profiles can be used in Tenant, Account, Product, Environment, Segment and Solution objects allowing full control over when the scenario will be used. For all of these objects the scenario profiles can be configured under 
```
   { "Profiles" : { "Scenarios" : [ <id of the scenario profile to load ] } }
```

When loading scenarios the data is loaded after the users input data has been applied, however to allow for users to override the configuration provided via a scenario we reload the users input data after the scenarios have been loaded. As a result of this users still have full control over the settings and component configuration that have been introduced by the scenario 

## Motivation and Context
Our previous implementation of scenarios was only available to providers which meant that users couldn't pick when they wanted a given scenario to be loaded. With the introduction of scenario profiles providers can create recommended scenarios that users can choose to implement into their solutions. 

This also has the ability to create a library provider of different scenarios that a tenant might want to use. Including this as pat of the hamlet providers list would make these available to any solution in the tenancy 

## How Has This Been Tested?
Tested locally 

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [X] None of the above.


roleyfoley:
**Example - a standard topic and subscription for alerts** 

Create the scenario - the scenario is part of my own provider (myprovider) and requires an EmailAddress
- The scenario adds two deployment units to the mgmt tier alarm-topic and monitoringemail which will setup a subscription 
- We then update the default alert profile to include this topic 
- the EmailAddress parameter is used to set a property of the monitoringemail component 

`myprovider/scenarios/scenario.ftl` 

```
[#ftl]

[@addScenario
    name="monitoringalerts"
    description="Send all alerts to monitoring"
    provider="myprovider"    
    properties=[
        {
            "Names" : "EmailAddress",
            "Description" : "The email address to receive the alerts",
            "Type" : STRING_TYPE,
            "Mandatory" : true
        }
    ]
/]

[#macro aws_scenario_monitoringalerts EmailAddress ]
    [@loadScenario
        blueprint={
            "Tiers" : {
                "mgmt" : {
                    "Components" : {
                        "alarm-topic" : {
                            "topic" : {
                                "Instances" : {
                                    "default" : {
                                        "deployment:Unit" : "alarm-topic"
                                    }
                                },
                                "Subscriptions" : {
                                    "Links" : {
                                        "monitoringemail" : {
                                            "Tier" : "mgmt",
                                                "Component" : "monitoringemail",
                                                "Instance" : "",
                                                "Version" : ""
                                        }
                                    }
                                }
                            }
                        },
                        "monitoringemail" : {
                            "external" : {
                                "Instances" : {
                                    "default" : {
                                        "deployment:Unit" : "monitoringemail"
                                    }
                                },
                                "Properties" : {
                                    "endpoint" : {
                                        "Key" : "SUBSCRIPTION_ENDPOINT",
                                        "Value" : EmailAddress
                                    },
                                    "protocol" {
                                        "Key" : "SUBSCRIPTION_PROTOCOL",
                                        "Value" : "email""
                                    }
                                }
                                "Profiles": {
                                    "Placement": "external"
                                }
                            }
                        }
                    }
                }
            },
            "AlertRules" : {
                "Notify" : {
                    "Severity" : "info"",
                    "Destinations" : {
                        "Links" : {
                            "alarm" : {
                                "Tier" : "mgmt",
                                "Component" : "slackalarm-topic",
                                "Instance" : "",
                                "Version" : ""
                            }
                        }
                    }
                }
            },
            "AlertProfiles" : {
                "default" : {
                    "Rules" : [ "Notify" ]
                }
            }
        }
    /]
[/#macro]
```

At our tenant we define a new scenario profile which uses the scenario from my provider and configures the email address
Then in the tenant object we include this profile which will load the scenario with the parameters provided

mytenant/tenant.json
```
    "ScenarioProfiles" : {
        "monitoring" : {
            "Scenarios" : {
                "slackalarms" : {
                    "Enabled" : true,
                    "Provider" : "myprovider",
                    "Name" : "monitoringalerts",
                    "Parameters" : {
                        "EmailAddress" : {
                            "Key" : "EmailAddress",
                            "Value" : "broken@mycompany.me"
                        }
                    }
                }
            }
        }
    },
    "Tenant" : {
        "Id" : "mytenant",
        "Profiles" : {
            "Scenarios" : [ "monitoring" ]
        }
    }
```

ml019 (approved):
Amazing what one little extra language feature can provide.

I do think the name "scenario" is too test related and we should think about renaming it to something more generic like "module" or equivalent.

Issue #1445, opened 2020-10-29
Scenario rename to modules 


roleyfoley:
## Current Behaviour

Scenarios were initially introduced with a specific purpose, to load testing scenarios that will be used to generate tests for our provider plugins. 

## Expected Behaviour
With #1439 scenarios can be used as part of blueprints by users which allows for a much broader scopes of usage. Scenarios are now more closely aligned with the concept of modules, which are common in other continuous deployment tools like terraform

## Possible Solution
Rename the scenarios implementation to modules

## Context
The main reason for this is to fit in with other tools and to provider better understanding of the concept

Issue #1444, opened 2020-10-29
Component Profile Deprecation 


roleyfoley:
## Current Behaviour
We currently have a large collection of profiles and associated configuration which were intended to provide an easy way to share standard configuration across multiple occurrences which use the same component type. I.e. the Processor Profile allows you to define the instance type for all ec2 components or ecs hosts. Each profile is for a specific purpose and implements configuration only for supported component types

Since profiles are so specific in how they are implemented, when developing a component you often need to make changes to both the component configuration and to the profile configuration. This adds maintenance issues as you need to maintain code in two places but can also cause collisions between components which use the same profile. 

## Expected Behaviour
With the introduction of Deployment Profiles and Policies we can essentially implement typed based shared component configuration across a whole solution which is centred around individual components. If all configuration for a component was available through the components configuration, you could use Deployment Profiles and Policies to essentially implement the profiles configuration but in a more flexible and generic style of configuration. 

## Possible Solution
There are two sides to this solution 
- Migrate all configuration that is performed via profiles to the components which consume the profiles currently 
   - In some cases we can use meta parameters #1429 to provide shared configuration blocks between components. These would need to be considered to avoid potential collisions but the use of meta parameters is a documented service to show that this is possible 
   - For more specific configuration we just need to add the configuration parameters and update setup macros as appropriate 

- Deprecate the existing profile usage and help users transition to the new approach 
   - Remove all master data provided profiles and ensure that macros which call the profiles return empty objects 
   - Create a scenario which emulates the master data profiles as equivalent deployment profiles to provide backwards compatibility 
   - If the profile is empty use the configuration added as part of the migration which will be merged with the master data profile 
  - If a profile isn't empty use the configuration from the profile, but provide deprecation warnings that this functionality will be removed 
  - Once are happy that customers have moved to deployment profiles remove the deprecation warnings and remove profile support 

## Context
This would be a great rationalisation of how we apply shared configuration, deployment profiles/policies apply at the component/ component type level, scenarios allow for loading in whole pieces of functionality and the user defined blueprint adds the finest level of control for your specific product

PR #1448, opened 2020-10-31
Layer support and Scneario Rename

roleyfoley:
## Description
This PR contains a collection of changes to the hamlet engine focussing around the way which modules are loaded into a given template generation process 

- Renames Scenarios to Modules - Closes: #1445 
- Removes the Scenario Profile and instead uses configuration directly applied to layer 
- Introduces the concept and handling of layers - Layers represent the context a template generation is running under, the combination of a given set of layer instances is the context that we are using to run a generation. Current layer Types are Tenant, Account, Product, Solution, Environment  and Segment. Each layer type can have one or more instances of they layer.
- Moves reference data loading to the bootstrap process 
- Use the shared configuration and data loading process for layers to define standard configuration across the layers and to reduce the need for empty value handling in setContext 

Further descriptions for each commit are in the long commit details

## Motivation and Context
The initial motivation with this PR was to rename scenarios to modules. However while performing this update I noticed that Scenarios are invoked using the profile approach which has now been deprecated #1444 . To implement applying the more complex configuration required to move away from profiles meant updating configuration in a number of locations and in the current state would have required a lot of code duplication. 

So to find a way to consolidate the configuration the idea of Layers popped into my head as a way to define the contexts we run deployments under

## How Has This Been Tested?
Tested locally - Using awstest test suite and template generation on an active product 

## Types of changes

## Checklist:
- [X] None of the above.


roleyfoley:
Please rebase and merge this commit to keep the individual commits

ml019 (approved):
No issues. It is a significant change but making the layer processing explicit makes sense and I can see where it will plug into the cmdb processing. 

Not 100% on districts but can see value of having a name of the top levels of the layer hierarchy. We say the long form often enough.

Issue #1462, opened 2020-11-16
Refactor - Account Layer Attribute Names


RossMurr4y:
## Expected Behaviour

The `Account` layer definition should establish only provider-independent values.

## Current Behaviour

Currently the `Account` layer defines a number of provider-specific values. These include:

- `AWSId`
- `AzureId`
- `aws:ecsAccountSettings`
- `Access.AWSId`
- `Registry.ShareAccess.AWSAccounts`

## Possible Solution

- [ ] deprecate `AWSId` and `AzureId` in favour of `ProviderId`, setting both as alt. names
- [ ] Move `aws:ecsAccountSettings` into the AWS provider & have provider-loading perform the name-spacing itself.
- [ ] Refactor `Access.AWSId` to use `ProviderId` and its new aliases
- [ ] Refactor `Registry.ShareAccess.AWSAccounts` to use `ProviderId` and its aliases`

PR #1466, opened 2020-11-18
updated account layer by adding provider attribute.

### LINKED ISSUE #1465: Refactor - Account-layer Provider Attribute
## Expected Behaviour

The Account-layer should be both provider independent in its available attribute names, yet contain configuration to indicate which provider a given Account is to be used with.

## Current Behaviour

Currently the Account Id's provide this information, through the attributes `AWSId` and `AzureId` [which are about to be removed.](hamlet-io/engine#1463)

## Possible Solution

Implement a new `Account`-layer attribute `Provider` that can be instead used to determine the intended provider. The attribute should have a default of `aws` to ensure continued operation of existing projects once the executor is updated to use the new attribute.

## Context

[These attributes are being consolidated into `ProviderId`](hamlet-io/engine#1463), however some logic within the executor performs account-switching functionality based on the usage of one or the other (set the corresponding CLI tool to use the correct account). The refactor alone would break this functionality without another way to determine the provider.

gumbaba-gs:
Implement a new Account-layer attribute Provider

## Motivation and Context
issue #1465

## How Has This Been Tested?

Not yet tested. will be with the incoming executor repo changes.

## Types of changes

## Followup Actions

* contributes to hamlet-io/engine#1462
* unblocks hamlet-io/executor-bash/issues/128
* Closes  #1465

## Checklist:

- [x] None of the above.

### DIFF
--- a/providers/shared/layers/Account/id.ftl
+++ b/providers/shared/layers/Account/id.ftl
@@ -15,6 +15,12 @@
             "Type" : STRING_TYPE
         },
         {
+            "Names" : "Provider",
+            "Type" : STRING_TYPE,
+            "Default" : "aws",
+            "Description" : "Define which provider this account applies to"
+        },
+        {
             "Names" : "Name",
             "Type" : STRING_TYPE
         },


ml019 (approved):
No issue.

PR #1473, opened 2020-11-25
fix(schema): metaparameters should not ref links

RossMurr4y:
## Description

Corrected the `metaparameter` scheme generation.

Note: I restructured the input object to reflect how it would appear as a `reference` or `component` composite object. This is in preparation for a subsequent PR to define metaparameters as their own data structure.

## Motivation and Context

The schema generation macro for this data type was incorrectly being told to reference metaparameters (itself) instead of defining its schema.

## How Has This Been Tested?

Local template generation

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/providers/shared/views/schema/setup.ftl
+++ b/providers/shared/views/schema/setup.ftl
@@ -88,11 +88,21 @@
         [#case "metaparameter"]
 
             [#-- Key Value Pairs of Metaparameter Name : Configuration --]
-            [#assign metaparametersConfiguration = {
-                "Links" : linkChildrenConfiguration
-            }]
+            [#assign metaparametersConfiguration =
+                {
+                    "Links" : {
+                        "Names" : "Links",
+                        "Type" : {
+                            "Singular" : "Link",
+                            "Plural" : "Links"
+                        },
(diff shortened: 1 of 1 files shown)


providers/shared/views/schema/setup.ftl
```diff
@@ -88,11 +88,21 @@
...
+                        },
+                        "Properties" : [],
+                        "Attributes" : linkChildrenConfiguration
+                    }
+                }
+            ]
 
-            [#list metaparametersConfiguration as id,childrenConfiguration]
+            [#list metaparametersConfiguration as id,configuration]
             [@addSchema
```
ml019, line 106: needs indenting

ml019 (changes requested):

providers/shared/views/schema/setup.ftl
RossMurr4y: Fixup in 252b302

RossMurr4y:
@ml019 think this ones good to go?

roleyfoley (approved):

PR #1476, opened 2020-11-26
Feature - Metaparameters as their own data type

### LINKED ISSUE #1429: Feature Request - Make Metaparameters Official
## Description
All data types in Hamlet should be defined using the composite object model established for existing data types (components, reference data).

## Context

[Now that `Metaparameters` are their own data type](https://github.com/hamlet-io/engine/pull/1406), they should be defined in the same manner (a composite object that added to with a `@addMetaparameter` macro, similar to `@addReference` or `@addComponent`)

## Current Behaviour
Metaparameters (currently just the "Links" object) are defined as a common data structure on a global variable. This makes their behaviour different to that of other data structures in Hamlet.

# Other
Comes from this comment https://github.com/hamlet-io/engine/pull/1406#pullrequestreview-506993733

RossMurr4y:
## Description

* Defines new data type - `metaparameters`
* Refactors schema generation to work with new data type instead of pretending to
* Refactor the way schemas reference eachother to better handle referencing other schema types than `metaparameters`

## Motivation and Context

Implements and Closes #1429 

## How Has This Been Tested?

Local schema generation

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.


engine/inputdata/metaparameter.ftl
```diff
@@ -0,0 +1,188 @@
...
+                "Type" : BOOLEAN_TYPE,
+                "Default" : true
+            },
+            {
+                "Names" : "IncludeInContext",
+                "Type" : ARRAY_OF_STRING_TYPE
+            }
+        ]
+]
+
```
roleyfoley, line 160: This should probably be part of the shared provider rather than in the engine. The engine is more responsible for implementing the storage and I/O for the inputdata type rather than the data that is injected.

providers/shared/metaparameters/links/id.ftl
```diff
@@ -0,0 +1,12 @@
...
+[@addMetaparameter
+    type=LINK_METAPARAMETER_TYPE
+    pluralType="Links"
+    properties=[
+        {
+                "Type"  : "Description",
+                "Value" : "A relationship between Components."
+        }]
+    attributes=linkChildrenConfiguration
+/]
```
roleyfoley, line 12: This is probably more where linkChildrenConfiguration should exist, as long as we load these early on then it shouldn't impact other usage of linkChildrenConfugrration

roleyfoley (changes requested):

engine/inputdata/metaparameter.ftl
RossMurr4y: Had a good discussion with @roleyfoley regarding this. It's difficult to move these assignments into `metaparameters/links/id.ftl` because of the load-order of configuration. Messing with the load order causes other complications.

This led to a discussion around how metaparameters should be referenced in the first place - instead of being assigned as variables like this and referenced in many places, Metaparameters more directly align with `AttributeSets` - its a collection of attributes that is itself a composite object structure, however it can be used in many places and "referenced" by components or other types (blueprints etc). This would solve issues relating to load-order because a reference value that is just a string ie. `{ "Ref" : "links" }` is resolvable as files are loading in, but then during composite object processing (which is all post-load of freemarker files) this can resolve to an actual full schema where "ref" values are replaced with `{"Children": [ <insert-links-composite-object> ]}`. 

This also perfectly aligns with JSONSchema "ref" attribute which is starting to become [difficult to use due to its implementation only within JSONSchema](https://github.com/hamlet-io/engine/issues/1477) 

I have moved `linkChildConfiguration` back to its original location in `component.ftl` and will log a new ticket to implement the "Ref" attribute within `getCompositeObject` and rename `Metaparameters` to `AttributeSets`.

RossMurr4y:
@roleyfoley  this one should be ready to merge now if you're happy with it.

roleyfoley:
@RossMurr4y Looks like https://github.com/hamlet-io/engine/pull/1473 created a conflict

RossMurr4y:
@roleyfoley rebased and tested

roleyfoley (approved):

PR #1482, opened 2020-11-30
feat(network): flow log control

roleyfoley:
## Description
Adds the ability to define and create your own network flow logs. Logs can either be sent to s3 or a log, s3 being our current component type name for object store and log for the providers native logging service 

Users can create different flows for different actions which have been applied to the flow ( accept, reject, all ) and then route this to an appropriate destination. 

This deprecates the existing boolean flow logs configuration 

## Motivation and Context
Flow logs can be used for a number of different purposes so allowing for them to be controlled and managed by the user makes sense 

## How Has This Been Tested?
tested locally and on development deployment 

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/providers/shared/components/network/id.ftl
+++ b/providers/shared/components/network/id.ftl
@@ -28,8 +28,46 @@
                 "Children" : [
                     {
                         "Names" : "EnableFlowLogs",
+                        "Description" : "Deprecated: Please use FlowLogs",
                         "Type" : BOOLEAN_TYPE,
                         "Default" : true
+                    },
+                    {
+                        "Names" : "FlowLogs",
+                        "Description" : "Log flows across the network",
+                        "Subobjects" : true,
+                        "Children" : [
+                            {
+                                "Names" : "TrafficType",
+                                "Type" : STRING_TYPE,
+                                "Description" : "The type of traffic to capture in the flow log",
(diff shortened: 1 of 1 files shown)


providers/shared/components/network/id.ftl
```diff
@@ -28,8 +28,46 @@
...
                         "Type" : BOOLEAN_TYPE,
                         "Default" : true
+                    },
+                    {
+                        "Names" : "FlowLogs",
+                        "Description" : "Log flows across the network",
+                        "Subobjects" : true,
+                        "Children" : [
+                            {
+                                "Names" : "TrafficType",
```
ml019, line 41: I was expecting this to be TCP, UDP etc. Think type isn't the right word here - maybe "Action"?

roleyfoley: Only problem with that is the `all` type, not really an action

roleyfoley: I'll make it any instead of all that will fit action

roleyfoley: Fixed and ready for review

ml019 (approved):

PR #1483, opened 2020-11-30
Data feed deployment specific prefixes

roleyfoley:
## Description
Adds support for including deployment specific prefixes to the user defined prefix

## Motivation and Context
Making paths globally unique for datafeeds is a requirement when creating consolidated data stores or having a path with identifiable details in it. This allows for these details to be injected in the datafeed deployment based on where the datafeed has been deployed

## How Has This Been Tested?
Tested locally

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/providers/shared/components/datafeed/id.ftl
+++ b/providers/shared/components/datafeed/id.ftl
@@ -56,6 +56,24 @@
                         "Type" : STRING_TYPE,
                         "Description" : "The prefix appeneded to the object name when processing failed",
                         "Default" : "error/"
+                    },
+                    {
+                        "Names" : "Include",
+                        "Description" : "Details to add to prefix",
+                        "Children" : [
+                            {
+                                "Names" : "Order",
+                                "Description" : "The order of the included prefixes",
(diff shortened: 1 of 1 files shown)


providers/shared/components/datafeed/id.ftl
```diff
@@ -56,6 +56,30 @@
                         "Type" : STRING_TYPE,
                         "Description" : "The prefix appeneded to the object name when processing failed",
                         "Default" : "error/"
+                    },
+                    {
+                        "Names" : "Include",
```
ml019, line 61: Will need the same on other sources of logging? If so, maybe worth making a separate configuration variable that can be used elsewhere?

roleyfoley: Its really only optional here when you are using a datafeed as an application component where it might not be useful

ml019: OK - fair call.

PR #1485, opened 2020-12-02
Queuehost component

roleyfoley:
## Description
initial definition of the queuehost component which is aimed to support managed queue broker services such as AmazonMq from AWS. The initial implementation focuses on a rabbitmq broker which requires a single admin user and is managed as a dedicated instance of the broker

## Motivation and Context
Queuehosts differ from a managed queue in that they are actually offering the hosting services to create any number of queues or associated services which can be defined as part of an application 

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/providers/shared/components/queuehost/id.ftl
+++ b/providers/shared/components/queuehost/id.ftl
@@ -133,6 +133,12 @@
                         "Default" : "root"
                     },
                     {
+                        "Names" : "EncryptionScheme",
+                        "Type" : STRING_TYPE,
+                        "Description" : "A prefix appended to link attributes to show encryption status",
+                        "Default" : ""
+                    },
+                    {
                         "Names" : "Secret",
                         "Children" : secretConfiguration
                     },
--- a/providers/shared/components/queuehost/id.ftl
+++ b/providers/shared/components/queuehost/id.ftl
@@ -1,0 +1,147 @@
+[#ftl]
+
+[@addComponentDeployment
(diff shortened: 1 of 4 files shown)


roleyfoley:
Also adds a minor change to move the SECRET_ATTRIBUTE_TYPE from the Azure provider to the shared provider

providers/shared/components/queuehost/id.ftl
```diff
@@ -0,0 +1,153 @@
...
+                "Type" : "Description",
+                "Value" : "Managed message queue hosting"
+            }
+        ]
+    attributes=
+        [
+            {
+                "Names" : "Engine",
+                "Type" : STRING_TYPE,
+                "Values" : [ "rabbitmq" ],
```
RossMurr4y, line 22: Should this value be added by the provider implementing it?

RossMurr4y (approved):

providers/shared/components/queuehost/id.ftl
roleyfoley: rabbitmq would be a standard implementation of a queue host and would be supported across multiple providers. So we could do it per provider but we don’t currently do that for dB and cache which have a similar configuration

ml019: Agree with @roleyfoley - if a standalone product like rabbit, its reasonable to have a standard value in engine

Issue #1443, opened 2020-10-29
Dynamic loading of fragments


roleyfoley:
## Current Behaviour
Fragments provide access to the hamlet engine for users and can be added as part of a CMDB. They are currently implemented as case statement which is assembled as part of the CMDB composite loading process. For each component this case statement is executed and if the fragment id matches the statement the fragment details are executed during the component processing. 

While fragments do provide a powerful way to modify your solution to meet the needs of a specific solution their processing is inefficient ( the case statement gets longer and longer ), brittle ( a break in one fragment breaks the whole process ) and they can't be shared or loaded through our dynamic loading processes ( the freemarker file with the case statement is built outside of the engine ) As a result of this they can't be included in Scenarios #1426  which supports including all other input data except for this 

## Expected Behaviour
- To be able to define documented details on a given fragment, and to execute the fragment as an isolated function for a given component ( i.e. the component will only call the macro that is has been told to ) 
- Fragments can be offerred by CMDB's and Providers 
- A component can invoke multiple fragments
- Components can define supported component types which will be checked to ensure fragments are appropriate for a component 

## Possible Solution
- Define a loader process inline with our dynamic loading process, the load process should support loading from CMDB or providers
- A component will be able to control the location through its configuration of the component ( must include a provider with a special _cmdb provider to load from a local cmdb  ). This reduces the risk of name collisions 
- A fragment is defined as a function, the function will receive a _context and occurrence parameter and is expected to return the _context after it has been appropriately modified 
- To implement a fragment, the fragment developer will need to include an addFragment macro which defines the macro and its parameters and also then include the function definition itself 

## Context
This would align fragments with the rest of our function loading processes and also provide documentation and context to fragments

RossMurr4y:
Does the name `Fragment` still match this then? I like your suggested solution, but `Fragment` sounds like it was named after the "which part of the case statement do you want to run" whereas this new function - that accepts a context and outputs a context - is more of a "context transformation".

Of course they do the same thing but if I was just learning Hamlet, I can't infer much from "Fragment" but something with "Context" in the name that indicates pre-processing context manipulation would allow me to apply pre-existing knowledge from other languages on to this feature.

ml019:
A fragment can basically do anything, as @roleyfoley has ably demonstrated. 

The name came from the idea of a code fragment - a part of the configuration process but not all of it.

Maybe "snippet" might make more immediate connection for the reader?

roleyfoley:
[...]

PR #1494, opened 2020-12-08
Component extension support

roleyfoley:
## Description
Initial implementation of https://github.com/hamlet-io/engine/issues/1443 which aligns fragments ( now extensions ) with our dynamic macro invocation process.

instead of relying on a switch statement and a context variable assigned within a setup routine the context is now handled as part of the extension processing. This still maintains a globally assigned  `_context` variable but it has been moved into the core engine configuration where it is contents is managed through the extension. This maintains backwards compatibility with our existing fragment processing. 

Extensions also add some extra features over the existing fragments 
- Isolated macros reduce the processing time and only invokes the extension required for a component 
- Definition and documentation 
    - Extensions must be added to the configuration before they can be used, this includes defining a collection of properties that are used to validated the fragments
        - Aliases: Support using the same extension with multiple fragment names 
        - Component Type Support: Each extension can specify component types that it supports, if the component that is invoking the extension doesn't match the processing fails ( a wild card can be provided ) 
       - Descriptions: allow you to say what the extension does and to reflect this in documentation
- Inclusion as part of modules and plugins: extensions are loaded as part of our plugin process allowing them to be included in plugins which are loaded as required and don't require a CMDB 

## Motivation and Context
Covered in https://github.com/hamlet-io/engine/issues/1443 but essentially the main motivation for this is to support adding extensions in plugins 

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.


engine/extension.ftl
```diff
@@ -0,0 +1,251 @@
...
+            ]]
+
+        [#local extensionMacro = getFirstDefinedDirective(extensionMacroOptions)]
+        [#if extensionMacro?has_content ]
+
+            [#-- Validate the extenion usage --]
+            [#if ! extensionDetails?has_content ]
+                [@fatal
+                    message="Could not find extension details"
+                    detail="An extension function was found by could not find its configuration details"
```
ml019, line 190: typo

engine/extension.ftl
```diff
@@ -0,0 +1,251 @@
+[#ftl]
+
+[#-- Component Extensions --]
+[#-- Allows for the modification of a components state and its setup routine by users --]
+[#-- Generally used for advanced configuration scenarios including programitc settings and adding specialised resources --]
```
ml019, line 5: typo

engine/extension.ftl
```diff
@@ -0,0 +1,251 @@
...
+[#-- Extensions are defined as part plugin but can also be added to users CMDBs as an embeded extension --]
```
ml019, line 6: typo

ml019 (changes requested):
Very elegant. Ability to apply multiple extensions an a mix of plugin provided and local will be good.

engine/extension.ftl
roleyfoley: Fixed

engine/extension.ftl
roleyfoley: Fixed

roleyfoley:
@ml019 Typos fixed up - you can also use the suggest changes option in the comment for stuff like that ( https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request ) 

In a PR comment box the small piece of paper with the + sign lets you suggest udpates which can be merged straight from the github console

engine/bootstrap.ftl
```diff
@@ -34,6 +34,9 @@
 [#-- Module Loading --]
 [#include "module.ftl" ]
 
+[#-- Extension loading --]
+[#include "extension.ftl" ]
```
ml019, line 38:
```suggestion
[#include fred.ftl]
```

ml019:
@roleyfoley thanks for the tip. Will help in future.

PR #1498, opened 2020-12-14
Refactor - Rename Metaparameters to AttributeSets

RossMurr4y:
## Description

* Rename the `Metaparameter` composite object to `AttributeSet`

## Motivation and Context

Contributes to #1491 

## How Has This Been Tested?

Local schema generation.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Refactor (non-breaking change which improves the structure or operation of the implementation)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Followup Actions

- [x] None

## Checklist:

- [x] My change requires a change to the [documentation](https://github.com/hamlet-io/docs).
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] None of the above.

Issue #1491, opened 2020-12-04
Composite Object Structure Reference Capability


RossMurr4y:
## Expected Behaviour

Any composite object (`component`, `referenceData` etc) should be able to specify "Children" of an Attribute by referencing another composite object structure.

## Current Behaviour

Currently the `component.ftl` file contains a number of global variable assignments that conform to the composite object structure by convention. They are then used wherever that common structure is needed so that multiple composite objects can share a common set of attributes. This allows re-use of these attributes but creates a series of dependency problems and obstructs the ability to create schemas that follow the logic of these lookups.

## Possible Solution

* Rename the recently implemented `Metaparameters` as `AttributeSets` - a collection of common attributes that are used in two or more composite objects.
* Update `getCompositeObject` to handle an "Reference" (or "Ref") property, which works the same as the existing "Children" however it is defined as a string. 
* Implement the global variables named in the format `*ChildConfiguration` that exist in `component.ftl` as `AttributeSets` so they are now loaded as standalone composite objects
* Update uses of those variables with a "Reference" to the `AttributeSets` type i.e `{ "Reference" : LINK_ATTRIBUTESET_TYPE }` 

During the initial loading of providers (which is when all the composite objects get defined) there are no problems with "which data type needs to load first" as instead of referencing a global variable that must already exist, each "Reference" is simply a string. Once all files are loaded into the engine and it gets to actually processing a composite object, the engine can perform an appropriate lookup and replace the "Reference" with a "Children" with the found composite.

## Context

* Provide more clarity around the intention for the current `Metaparameters` composite type
* Enable re-use of composite-structures
* Enable complete documentation of composite-structures
* Remove complexity related to the sequence of events that files are loaded by the engine

RossMurr4y:
`AttributeSets` have been implemented within the engine now.

Some items remain to be converted over from global variables however this ticket was only to include the capability.

Closing.

PR #1502, opened 2020-12-16
Feature - Composite Object AttributeSet Attribute

RossMurr4y:
## Description

* Implement new properties `AttributeSet` and `Component` to the composite object structure.
  * `AttributeSet` defines a reference to an AttributeSet type.
  * `Component` defines a reference to another component. Intended for internal use only (the Engine will apply it where required).

## Motivation and Context

* Contributes to https://github.com/hamlet-io/engine/issues/1491
* Enable composite structure re-use
* Remove file dependency issues at file load time

**Example of Implementation:**
The attribute `Links` on this example component will at the time of loading require no other composite object structures exist in order to be defined. During engine runtime, was the `getCompositeObject()` function is called, it will search the structure of the `attributeset` schema for the `link` type and add its structure as children to the `Links` attribute. 

```ftl
[@addComponent
    type=EXAMPLE_COMPONENT_TYPE
    properties=[... ]
    attributes=
        [
            {
                "Names" : "Links",
                "Subobjects" : true,
                "AttributeSet" : LINK_ATTRIBUTESET_TYPE
            }
        ]
/]
```

## How Has This Been Tested?

Local template generation.

## Types of changes

## Followup Actions

- [ ] Requires https://github.com/hamlet-io/engine/pull/1498

## Checklist:

- [ ] None of the above.


RossMurr4y:
Marked as draft until https://github.com/hamlet-io/engine/pull/1498 is merged, then needs a rebase before unmarking as draft.

roleyfoley:
Looks like a good refactor and should make understanding easier 

Only thing I would ask is do we need to support multiple schemas? I think the idea is that that AttributeSets are our answer to shared schema. Do we need to support sharing all schemas and how would that be implemented for something like bueprint?

engine/base.ftl
```diff
@@ -765,17 +773,74 @@ are added.
...
+            [#if attribute.Reference.Type?has_content ]
+            
+                [#-- Determine composite object --]
+                [#switch attribute.Reference.Schema]
+                    [#case "attributeset" ]
+                        [#local attributeRefValue = attributeSetConfiguration[attribute.Reference.Type]!{} ]
+                        [#break]
+                    [#case "blueprint" ]
+                        [#local attributeRefValue = blueprintObject[attribute.Reference.Type]!{} ]
+                        [#break]
```
roleyfoley, line 805: Not sure I can see when you would ref a blueprint, that would become a bit of inception. Since anything inside of this composite object is most likely going to be in the blueprint.

engine/base.ftl
```diff
@@ -765,17 +773,74 @@ are added.
...
+                    [#case "component" ]
+                        [#local attributeRefValue = componentConfiguration[attribute.Reference.Type]!{} ]
+                        [#break]
+                    [#case "reference" ]
+                        [#local attributeRefValue = referenceConfiguration[attribute.Reference.Type]!{} ]
+                        [#break]
```
roleyfoley, line 811: Probably good to remove reference to, since components should reference an instance of reference data rather than include reference configuration in the component

roleyfoley (changes requested):

RossMurr4y:
> Looks like a good refactor and should make understanding easier
> 
> Only thing I would ask is do we need to support multiple schemas? I think the idea is that that AttributeSets are our answer to shared schema. Do we need to support sharing all schemas and how would that be implemented for something like bueprint?
> 
> edit: After conversation I think we can still use an object, there is a strong requirement to show the relationships between components and subcomponents so this can stay as an object, but we should limit the available reference sets

After further conversation, @roleyfoley and I have determined that the "Reference" composite object attribute does not need to be an object at all. There is no need to reference types other than AttributeSet, and referencing child components is handled independent of the composite object structure.

I will update this PR to rename "Reference" attribute as "AttributeSet", which provides more clarity as to what is happening and why.

RossMurr4y:
@roleyfoley I have updated this draft and it is now ready for review.

As discussed, I have replaced the "Reference" composite object attribute with "AttributeSet" which is far simpler to assign and doesn't introduce unnecessary complexity.

At schema generation time, there is a need to be able to differentiate between "I am constructing a reference to an AttributeSet" and "I am constructing another kind of reference" (such as from one component to another for subComponents). For this I have added the "Component" composite object attribute, however this is not intended for use within definition files (id.ftl), it's used only by the engine itself as is recognises a subComponent and constructs a reference to another component.

roleyfoley (dismissed):

roleyfoley:
@RossMurr4y I think this PR will need a rebase as it does include some changes made outside of this PR and is currently conflicting 

@ml019 Could you pass an eye over this one given the potential impact around getCompositeObject?

RossMurr4y:
Conflicts resolved and matching PR submitted for AWS provider. Azure does not require an update.

RossMurr4y:
Rebased and updated with changes from #1521

RossMurr4y:
@ml019 would you mind reviewing? Other repositories are beginning to have features merged that expect the changes implemented here.

engine/base.ftl
```diff
@@ -721,7 +721,8 @@ are added.
                     "Values" : [],
                     "Children" : [],
                     "SubObjects" : false,
-                    "PopulateMissingChildren" : true
+                    "PopulateMissingChildren" : true,
+                    "AttributeSet" : ""
```
ml019, line 725: Also needs an entry for Component?

engine/base.ftl
```diff
@@ -745,7 +746,9 @@ are added.
                         "Values" : asArray(attribute.Values![]),
                         "Children" : asArray(attribute.Children![]),
                         "SubObjects" : attribute.SubObjects!attribute.Subobjects!false,
-                        "PopulateMissingChildren" : attribute.PopulateMissingChildren!true
+                        "PopulateMissingChildren" : attribute.PopulateMissingChildren!true,
+                        "AttributeSet" : attribute.AttributeSet!"",
+                        "Component" : attribute.Component!""
```
ml019, line 751: There is no mention of component in the PR description. I understand what an attributeSet is doing, but what is component doing?

providers/shared/components/component.ftl
```diff
@@ -131,165 +131,6 @@
 [#assign USERPOOL_AUTHPROVIDER_COMPONENT_TYPE = "userpoolauthprovider" ]
 [#assign USERPOOL_RESOURCE_COMPONENT_TYPE = "userpoolresource" ]
 
-[#assign
-    filterChildrenConfiguration = [
```
ml019, line 135: A filter is basically what is now being referred to as a district. A link is basically a district with some extra attributes, which is why there were two separate definitions here.

It doesn't need addressing now but at some point, the link attribute set may need to be separated into two structure again.

ml019 (changes requested):

ml019:
Will need to be rebased as well

engine/base.ftl
RossMurr4y: Component defines a SubComponent within the composite object model. It is not intended for further use, or manual assignment within `id.ftl` files, rather it separates a reference to another component schema out from a reference to an AttributeSet.

A method to differentiate the two is necessary to provide accurate schema references, that define both the "SchemaType" (component, reference, attributeset) and the Instance (baseline, alertprofile, link, etc) values.

engine/base.ftl
RossMurr4y: updated in https://github.com/hamlet-io/engine/pull/1502/commits/8e88a00f6a4cd65b69bf555e6c2325a03f45d7ed, cheers

engine/base.ftl
RossMurr4y: I've updated the PR description to include this.

providers/shared/components/component.ftl
RossMurr4y: We can do that easily enough when the time comes, no?

PR #1510, opened 2021-01-13
feat(s3): add s3vpcaccess

kshychko:
## Description

An extension that provides access for publically available files in an s3 bucket from the VPC.
## Motivation and Context

For one of the tenants, we faced the situation when SourIP condition wasn't working for accessing the publically available files from the docker containers running on ECS within a VPC, so SourceVPC condition needed to be added.
## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/providers/shared/extensions/s3vpcaccess/extension.ftl
+++ b/providers/shared/extensions/s3vpcaccess/extension.ftl
@@ -9,7 +9,7 @@
         "Set up an access within VPC to s3 items"
     ]
     supportedTypes=[
-        "*"
+        S3_COMPONENT_TYPE
     ]
 /]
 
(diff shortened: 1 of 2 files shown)


providers/shared/extensions/s3vpcaccess/extension.ftl
```diff
@@ -0,0 +1,89 @@
...
+[@addExtension
+    id="s3vpcaccess"
+    aliases=[
+        "_s3vpcaccess"
+    ]
+    description=[
+        "Set up an access within VPC to s3 items"
+    ]
+    supportedTypes=[
+        "*"
```
roleyfoley, line 12:
```suggestion
        S3_COMPONENT_TYPE
```

Since this is really specific to S3 might be a good idea to restrict this one

roleyfoley (changes requested):

PR #1531, opened 2021-02-02
fix: flowlog configuration

ml019:
## Description
Remove/disable flag based configuration settings for flowlogs.

## Motivation and Context
Support for flowlogs now needs to be explicitly included on the network component. These changes ensure the default configuration does not trigger the checks in place to report usage of the legacy format of flowlog configuration.

## How Has This Been Tested?
Local template generation

## Types of changes

## Followup Actions
- [x] AWS implementation - https://github.com/hamlet-io/engine-plugin-aws/pull/226

## Checklist:

- [x] None of the above.

### DIFF
--- a/providers/shared/layers/Environment/id.ftl
+++ b/providers/shared/layers/Environment/id.ftl
@@ -144,7 +144,7 @@
             "Children" : [
                 {
                     "Names" : "Enabled",
-                    "Description" : "Don't use - here to override automatically added enabled attribute",
+                    "Description" : "Deprecated - here to override automatically added enabled attribute",
                     "Types"  : BOOLEAN_TYPE,
                     "Default" : false
                 },
--- a/providers/shared/layers/Environment/id.ftl
+++ b/providers/shared/layers/Environment/id.ftl
@@ -143,6 +143,12 @@
             "Names" : "FlowLogs",
             "Children" : [
                 {
+                    "Names" : "Enabled",
+                    "Description" : "Don't use - here to override automatically added enabled attribute",
+                    "Types"  : BOOLEAN_TYPE,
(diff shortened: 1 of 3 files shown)


providers/shared/layers/Environment/id.ftl
```diff
@@ -142,6 +142,12 @@
         {
             "Names" : "FlowLogs",
             "Children" : [
+                {
+                    "Names" : "Enabled",
+                    "Description" : "Don't use - here to override automatically added enabled attribute",
+                    "Types"  : BOOLEAN_TYPE,
+                    "Default" : false
+                },
```
roleyfoley, line 150: Might be better to say Deprecated instead of Don't use

providers/shared/layers/Segment/id.ftl
```diff
@@ -87,7 +87,9 @@
                     "Children" : [
                         {
                             "Names" : "Enabled",
-                            "Types"  : BOOLEAN_TYPE
+                            "Description" : "Don't use - here to override automatically added enabled attribute",
+                            "Types"  : BOOLEAN_TYPE,
+                            "Default" : false
                         },
```
roleyfoley, line 93: Might be better to say Deprecated instead of Don't use

roleyfoley (changes requested):

providers/shared/layers/Segment/id.ftl
ml019: done

providers/shared/layers/Environment/id.ftl
ml019: done

roleyfoley (approved):

PR #1532, opened 2021-02-02
Shared provider testing

roleyfoley:
## Description

Adds initial support for testing the shared provider has its own plugin similar to what we have for the AWS and Azure plugin. 

The support is based on the externalservice component which is a general purpose component that we use for attribute mappings, this is a good candidate as a testing component since we can also use this to test our extension macros 

In order to do this the externalservice now supports a "deployment" in the sharedtest provider which creates a config file representing the Occurrence of the external service. 

## Motivation and Context
Allows for testing and catching issues in the shared provider before they are released 

## How Has This Been Tested?
Tested locally and will use this PR to test CI support

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.


providers/sharedtest/components/externalservice/setup.ftl
```diff
@@ -1,10 +1,10 @@
 [#ftl]
 
-[#macro shared_externalservice_default_deployment_generationcontract occurrence ]
```
RossMurr4y: I don't think we want to rename this file here, as it will leave the `externalservice` component without a setup routine

roleyfoley: I moved it from the shared provider into the sharedtest prvovider in 81a7046. The externalservice component hasn't had a setup routine before and doesn't really need one. 

The shared provider looks after the state and the sharedtest provider looks after the setup for testing purposes

RossMurr4y: Gotcha. No worries then :)

RossMurr4y (approved):

Issue #1535, opened 2021-02-04
Mockoutput input provider


roleyfoley:
## Expected Behaviour
When building a solution I can run the deployment output creation without having to have all resources deployed

## Current Behaviour
When creating a solution the expectation is that you are deploying everything as you add the components into the solution. When you want to test out the solution or draw diagrams of the solution this requires you to completely deploy the solution in order to draw the diagram of the solution

## Possible Solution
We currently have two input providers mock and composite, a possible solution would be to add another one the `mockcomp` provider which sits in between these two providers 

blueprint, settings and definitions are all provided through our standard composite input processing 
outputs have an additional layer where getStackOutput checks the composite outputs and if the output can't be provided it instead offers a mock output response in place of the actual response.

This would allow for template testing and generation of new solutions 

## Context
There are a couple of scenarios that this is useful 
- In tutorials we would like to reduce the work required to use hamlet and show its benefits. Being able to build a whole solution without needed to deploy anything would be really handy. We can look through the templates or draw diagrams at each stage of the tutorial 
- When developing a new part of a solution it would be nice to see what the templates look like before running through the deployment process. I.e. shorten the solution development loop

PR #1538, opened 2021-02-04
feat: whatif input provider

roleyfoley:
## Description
Closes: https://github.com/hamlet-io/engine/issues/1535

- Creates a new input provider, the whatif provider 
- Refactors some of the input processing functions so they can be shared across providers 

## Motivation and Context
The whatif input provider allows you to use a composite CMDB to create and query deployments before actually deploying the changes you've made in your CMDB

Input sources are primarily handled the same way as the composite input source except for StackOutputs. The whatif provider checks to see if a composite stack output can be found and if it can't then it will return a mocked output instead

This allows you to add components to a CMDB to see what they will create before deploying any of their required components. This is really helpful when generating diagrams or running through demonstrations where you don't want to wait for deployments to complete

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/providers/shared/inputsources/shared/setting.ftl
+++ b/providers/shared/inputsources/shared/setting.ftl
@@ -1,0 +1,98 @@
+[#ftl]
+
+[#---------------------------------------------------------------
+-- Internal support functions for composite setting processing --
+-----------------------------------------------------------------]
+
+[#function internalReformatSettings objects fileRegex=".*"]
+    [#local result = {}]
+    [#list objects!{} as key,value]
+        [#local result =
+            mergeObjects(
+                result,
+                internalReformatFiles(key, value!{}, fileRegex)
+            )]
+    [/#list]
+    [#return result]
+[/#function]
+
+[#function internalReformatFiles key settingsFiles fileRegex]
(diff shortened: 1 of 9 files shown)


providers/shared/inputsources/shared/setting.ftl
```diff
@@ -0,0 +1,98 @@
...
+        [#local result =
+            mergeObjects(
+                result,
+                internalReformatFiles(key, value!{}, fileRegex)
+            )]
+    [/#list]
+    [#return result]
+[/#function]
+
+[#function internalReformatFiles key settingsFiles fileRegex]
```
RossMurr4y, line 19: A comment here on the purpose of the reformat would be helpful. A lot of reading is required to understand the desired final format.

roleyfoley: I moved these functions from composite to the shared provider so they would be available to both providers. Writing these functions wasn't part of this change

RossMurr4y (approved):

PR #1548, opened 2021-02-16
Feat: New Entrance - validate (the blueprint)

### LINKED ISSUE #1450: Composite Object extra parameter detection
## Expected Behaviour
If an object that is passed to the composite object processing has more parameters than expected an error should be provided to the user outlining the extra parameters 

## Current Behaviour
The extra parameters are stripped from the object without knowledge

## Possible Solution
During the getCompositeObject function call check for parameters which haven't been defined in the configuration schema 

## Context
This is useful for a couple of purposes 
- Provides better feedback to users when they are working in a CMDB 
- Allows us to catch configuration migrations which might have been missed when implementing a schema on a configuration item

RossMurr4y:
## Description

Add new `entrance` to the shared provider which will perform a blueprint validation, warning of unexpected values.

## Motivation and Context

Closes #1450

BREAKING CHANGE: The composite object model in Hamlet is used throughout hamlet to
define the various data types used by the engine. These data types
are then used within Hamlet Tenants/Accounts/Products by end-users
in order to configure their CMDB's.

Currently, to determine a given attribute's value the composite object
model simply says : here is the composite object, give me attribute
"x", normalise it, evalute its attriubute sets and return it to me.
The remainder of the composite object is not validated. In this way
it is trivial for a hamlet user to input a typo, an attribute in the
wrong object level or be otherwise wrong and not know it - hamlet
will work as expected so long as all the mandatory Attributes are
found.

This provides a mechanism to ensure the integrity of all of your input sources whilst not breaking existing blueprints that may have a typo-or-two.

## How Has This Been Tested?

Local template testing alongside matching Azure and AWS changes.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Followup Actions

- [x] None 

## Checklist:

- [x] All new and existing tests passed.
- [x] None of the above.

### DIFF
--- a/engine/base.ftl
+++ b/engine/base.ftl
@@ -722,6 +722,7 @@ are added.
                     "Children" : [],
                     "SubObjects" : false,
                     "PopulateMissingChildren" : true,
+                    "AllowAdditionalParameters" : false,
                     "AttributeSet" : "",
                     "Component" : ""
                 } ]
@@ -748,6 +749,7 @@ are added.
                         "Children" : asArray(attribute.Children![]),
                         "SubObjects" : attribute.SubObjects!attribute.Subobjects!false,
                         "PopulateMissingChildren" : attribute.PopulateMissingChildren!true,
+                        "AllowAdditionalParameters" : attribute.Children?has_content?then(attribute.AllowAdditionalParameters!false, false),
                         "AttributeSet" : attribute.AttributeSet!"",
                         "Component" : attribute.Component!""
                     } ]
@@ -770,6 +772,7 @@ are added.
                         "Children" : [],
                         "SubObjects" : false,
                         "PopulateMissingChildren" : true,
+                        "AllowAdditionalParameters" : false,
                         "AttributeSet" : "",
                         "Component" : ""
                     }
@@ -830,6 +833,50 @@ are added.
                     [/#list]
                 [/#if]
             [/#list]
+
+            [#-- Process "AllowAdditionalParameters" value   --]
+            [#-- false: the matched candidated object cannot --]
+            [#--        have undefined parameters provided.  --]
+            [#if providedCandidate?has_content]
+                [#if !(attribute.AllowAdditionalParameters)]
+                    [#local validKeys = 
+                        getUniqueArrayElements(
+                            asFlattenedArray(
+                                evaluatedRefAttributes?map(a -> a.Names)
+                            )
+                        )
+                    ]            
+                    [#local parameters = providedCandidate?keys]
+                    [#-- Common Parameters that are used throughout        --]
+                    [#-- but are as-yet unaccounted for in the             --]
+                    [#-- composite object definitions.                     --]
+                    [#-- TODO(rossmurr4y): add commonParams to definitions --]
+                    [#local commonParams = [
+                        "Id",
+                        "Name",
+                        "Instances",
+                        "Description",
+                        "Configured",
+                        "Extensions"]
+                    ]
+                    [#if !(validKeys?seq_contains("*"))]
+                        [#list parameters as parameter]
+                            [#if !(validKeys?seq_contains(parameter)) &&
+                                !(commonParams?seq_contains(parameter))]
+                                [@fatal 
+                                    message="Attribute Object does not allow additional parameters." 
+                                    context=
+                                        {
+                                            "ValidKeys" : validKeys,
+                                            "DiscoveredKey" : parameter,
+                                            "ProvidedObject" : providedCandidate
+                                        }
+                                /]
+                            [/#if]
+                        [/#list]
+                    [/#if]
+                [/#if]
+            [/#if]
 
             [#-- Name wildcard means include all candidate objects --]
             [#if providedName == "*"]
--- a/engine/base.ftl
+++ b/engine/base.ftl
@@ -837,47 +837,50 @@ are added.
             [#-- Process "AllowAdditionalParameters" value   --]
             [#-- false: the matched candidated object cannot --]
             [#--        have undefined parameters provided.  --]
-            [#if providedCandidate?has_content]
+            [#if attribute.Children?has_content && providedCandidate?has_content && providedValue?is_hash]
                 [#if !(attribute.AllowAdditionalParameters)]
-                    [#local validKeys = 
-                        getUniqueArrayElements(
-                            asFlattenedArray(
-                                evaluatedRefAttributes?map(a -> a.Names)
-                            )
-                        )
-                    ]      
-                    [#local parameters = providedCandidate?keys]
-                    [#-- Common Parameters that are used throughout        --]
-                    [#-- but are as-yet unaccounted for in the             --]
-                    [#-- composite object definitions.                     --]
-                    [#-- TODO(rossmurr4y): add commonParams to definitions --]
-                    [#local commonParams = [
-                        "Id",
-                        "Name",
-                        "Instances",
-                        "Description",
-                        "Configured",
-                        "Extensions",
-                        "multiAZ",
-                        "deployment:Unit",
-                        "deployent:Group"]
-                    ]
-                    [#if !(validKeys?seq_contains("*"))]
-                        [#list parameters as parameter]
-                            [#if !(validKeys?seq_contains(parameter)) &&
-                                !(commonParams?seq_contains(parameter))]
-                                [@fatal 
-                                    message="Attribute Object does not allow additional parameters." 
-                                    context=
-                                        {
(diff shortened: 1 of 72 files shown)


RossMurr4y:
Ready for review after testing corresponding Azure and AWS provider plugin PRs. 

They should be reviewed/committed first, in order to ensure the test suite continues to work (or starts to work, in Azure's case).

AWS: hamlet-io/engine-plugin-aws#235
Azure: hamlet-io/engine-plugin-azure#223

engine/base.ftl
```diff
@@ -831,6 +834,56 @@ are added.
...
+                        [#-- but are as-yet unaccounted for in the             --]
+                        [#-- composite object definitions.                     --]
+                        [#-- TODO(rossmurr4y): add commonParams to definitions --]
+                        [#local commonParams = [
+                            "Id",
+                            "Name",
+                            "Title",
+                            "Enabled",
+                            "Description",
+                            "Instances",
```
roleyfoley, line 860: Would this also need to include `Versions`

engine/base.ftl
```diff
@@ -831,6 +834,56 @@ are added.
...
+                            "Configured",
+                            "Extensions",
+                            "multiAZ",
+                            "deployment:Unit",
+                            "deployent:Group"]
```
roleyfoley, line 865: These should be added to the appropriate composite objects, weird that they aren't 
It also includes `deployment:Priority` 

Should this also include `DeploymentUnits`

engine/base.ftl
```diff
@@ -831,6 +834,56 @@ are added.
...
+                        ]
```
roleyfoley, line 866: There would also be `Qualifiers` which is used in some cases ( certificates I think )

roleyfoley (changes requested):

roleyfoley:
> False by default, this will provide fatal error and sufficient debugging information in the event that the composite object being evaluated contains configuration that is undefined.
> Where necessary, this can then just be disabled on a per-attribute basis.

Would it be worth controlling the the default value for this using a commandLineOption? This would allow a user to include it and run something like a validate pass?

engine/base.ftl
RossMurr4y: Hmm yes, good idea. This (and others you've outlined below) are likely never used in the test-suites and the dev Solution I was working with so never came up. Will add them.

engine/base.ftl
RossMurr4y: I think this is already accounted for:

https://github.com/hamlet-io/engine/blob/fc0c3fd4a1e0261c772cce4b7290b5be4d7ecc89/providers/shared/components/component.ftl#L676-L693
https://github.com/hamlet-io/engine/blob/fc0c3fd4a1e0261c772cce4b7290b5be4d7ecc89/engine/common.ftl#L801-L814

RossMurr4y:
@roleyfoley I have resolved the comments you've made earlier as they should be addressed in the latest update, though there are also significant changes.

@ml019 I've made changes here that impact the work you're doing on Qualifiers - depending on which gets sorted first, the updates to them I've made here are merely a stop-gap toward your changes.

engine/component.ftl
```diff
@@ -126,18 +126,59 @@
...
+            "Names" : "deployment:Priority",
+            "Description" : "The Deployment Priority",
+            "Types" : NUMBER_TYPE,
+            "Default" : 5
+        },
+        {
+            "Names" : [ "deployment:Group", "DeploymentGroup" ],
+            "Description" : "The component Deployment Group",
+            "Types" : STRING_TYPE
+        },
```
roleyfoley, line 157: I don't think these should be included as part of this. They are now only added where they are actually used, rather than being permitted everywhere. Where were you seeing these as not being set?

engine/component.ftl
```diff
@@ -126,18 +126,59 @@
...
+                    "Names" : "Versions",
+                    "Description" : "Versions of the components instance.",
+                    "SubObjects" : true,
+                    "Children" : [
+                        {
+                            "Names" : [ "Extensions", "Fragment", "Container" ],
+                            "Description" : "Extensions to invoke as part of component processing",
+                            "Types" : ARRAY_OF_STRING_TYPE,
+                            "Default" : []
+                        }
```
roleyfoley, line 173: Extensions are only supported on select components, not sure why they would need to be explicitly included on Versions. That might be a bug from a test CMDB

engine/inputdata/blueprint.ftl
```diff
@@ -21,6 +21,238 @@
     [/#if]
 [/#macro]
 
+[#function getBlueprintAttributes ]
```
roleyfoley, line 24: This might be better as getBlueprintConfiguration in line with the the collections of configuration objects you are merging in the function

engine/inputdata/blueprint.ftl
```diff
@@ -21,6 +21,238 @@
...
+            [#local components = []]
+            [#if config.Components??]
+                [#list config.Components as child]
+                    [#local components += [{
+                        "Names" : child.Component,
+                        "SubObjects" : true,
+                        "Children" : asFlattenedArray(componentConfiguration[child.Type]["ResourceGroups"]["default"].Attributes?values) + coreComponentChildConfiguration
+                    }] ]
+                [/#list]
+            [/#if]
```
roleyfoley, line 74: I think any resource group can provide configuration so might be better to loop over all potential resource groups ( currently there is only the default one )

engine/inputdata/blueprint.ftl
```diff
@@ -21,6 +21,238 @@
...
+                    "Children" : [
+                        {
+                            "Names" : "*",
+                            "Types" : ANY_TYPE
+                        }
+                    ]
+                }
+            ]
+        }
+    ]]
```
roleyfoley, line 179: I don't know of a way to include Settings in the blueprint. Not sure this one should be here

engine/inputdata/blueprint.ftl
```diff
@@ -21,6 +21,238 @@
...
+                    "Names" : "*",
+                    "Types" : OBJECT_TYPE,
+                    "Children" : [
+                        {
+                            "Names" : "Stem",
+                            "Types" : STRING_TYPE,
+                            "Mandatory" : true
+                        }
+                    ]
+                },
```
roleyfoley, line 208: I think domain has more properties than just stem. It has a few like Name, Primary etc.  I thought there was some validation for this somewhere

providers/shared/entrances/validate/entrance.ftl
```diff
@@ -0,0 +1,40 @@
...
+                "Framework" : {
+                    "Name" : DEFAULT_DEPLOYMENT_FRAMEWORK
+                }
+            },
+            "Flow" : {
+                "Names" : [ "views" ]
+            },
+            "View" : {
+                "Name" : BLUEPRINT_VIEW_TYPE
+            }
```
roleyfoley, line 26: I don't think that is quite right your view shared_view_default_validate_config should have a type as well I think

roleyfoley (changes requested):

engine/component.ftl
RossMurr4y: Ah yep I added this because of an instance of it in one of the provider masterdata files. I'll remove it here and amend the original.

engine/inputdata/blueprint.ftl
RossMurr4y: Updated here: 51ca5d4

engine/inputdata/blueprint.ftl
RossMurr4y: I'm really not sure how / why you would validate them either... 

I've removed them now.

engine/inputdata/blueprint.ftl
RossMurr4y: Name is added as a part for the Framework Attributes to everything, and I found the other configuration defined in `getDomainObjects()` - I've refactored that now in b463fd8 for your review.

providers/shared/entrances/validate/entrance.ftl
RossMurr4y: I created a new view type for `validate` and moved existing macros out of the blueprint view.

31e3353

RossMurr4y:
Rebased and ready again for review

engine/inputdata/blueprint.ftl
roleyfoley: Stem is used everywhere? What does it do? that seems really weird

roleyfoley: Sorry didn't expand where the domain value was being used. All good

roleyfoley (approved):

PR #1549, opened 2021-02-18
feat: refactor input sources

ml019:
## Description
First set of refactors for input sources

## Motivation and Context
As part of supporting dynamic cmdb loading, input sources are being refactored to be more dynamic in the way they load state.

This change adds the necessary support for this refactor without changing existing processing to break up the process of transition.

Input sources now need to be explicitly defined. They have a sequence of stages, with each stage then having seeders that provide state to a stage.

Providers can contribute any of source, stages and seeders. Directory trees for `inputSources`, `inputStages` and `inputSeeders` are loaded as part of provider loading, which in turn explicitly define the corresponding items via `addInputSource`, `addInputStage` and `addInputSeeder` macros.

The shared provider implements a base set of input sources, stages and seeders. Other providers will commonly contribute seeders to existing stages.

Rather than the input state being static, it is dynamically calculated when requested by invoking the seeder functions. Seeder
functions are provided with a filter which they can use to determine what, if any state they should contribute. Because they receive the results of the previous seeder, they can override each other.

Both the input source and the filter can be modified at any time and the next request for the state will reflect this change.

On startup, a `bootstrap` input source is available that has a single stage to seed command line options. In turn, this may then define an option for the input source be used for processing.

All input sources include the command line option stage.

Within the object representing the state, the key categories of input are then identified by top level attributes - `CommandLineOptions`, `Masterdata`, `Blueprint`, `Settings`, `StackOutputs` and `Definitions`.

## How Has This Been Tested?
Local template generation

## Types of changes

## Followup Actions
- [x] AWS seeder implementation - https://github.com/hamlet-io/engine-plugin-aws/pull/236
- [X] Azure seeder implementation - https://github.com/hamlet-io/engine-plugin-azure/pull/224

## Checklist:

- [x] None of the above.


RossMurr4y:
> Input sources now need to be explicitly defined. 

Could you include in the description an example of this?

RossMurr4y:
I feel both Filters and Qualifiers would benefit from becoming AttributeSets.

As they currently are they are a "black box" item that a user needs to know about, but will have a difficult time sourcing documentation for. As an AttributeSet they will have JSONSchema created for them and be kept up-to-date automatically. With the increased importance to them now it seems like a good time to make the change over.

I was thinking this whilst working on #1548 - the Filter and Qualifiers are really quite different from the rest of the values assigned to the "frameworkObjectAttributes" global variable.

providers/shared/inputseeders/shared/id.ftl
```diff
@@ -0,0 +1,133 @@
...
+                    "MinDepth" : 1,
+                    "MaxDepth" : 1,
+                    "FilenameGlob" : "masterdata.json"
+                }
+            )[0].ContentsAsJSON
+        )!{}
+    ]
+[/#macro]
+
+[#function shared_input_masterdata_seeder filter state]
```
roleyfoley, line 40: Our convention for dynamic functions is probably better as 

`<provider>_<function>_<qualifers>` 

Since the seeder is a functionality wouldn't it be better as 

`shared_inputseeder_masterdata` 
`sahred_inputloader` 

etc

engine/base.ftl
```diff
@@ -1470,3 +1470,303 @@ A comparator is an operator and a version
...
+        },
+        {
+            "Filter" : {"Any" : "industry"},
+            "MatchBehaviour" : ANY_FILTER_MATCH_BEHAVIOUR,
+            "Value" : 23,
+            "CombineBehaviour" : MERGE_COMBINE_BEHAVIOUR
+        }
+    ]
+}
+
```
roleyfoley, line 1708: Can I use both approaches? The second approach while more verbose is far easier to understand. Looking at the qualifiers in the simplified form is really hard to follow. 

What is prod? what is industry? How as a user can I find out what is available for qualification?

roleyfoley: Any provides a real potential for danger or misunderstanding I think

ml019:
By the time getCompositeObject() sees the input, Qualifiers will have been applied and removed. You will note there is a stage for applying qualifiers.

We could still define their structure from a documentation perspective.

roleyfoley (changes requested):

providers/shared/inputseeders/shared/id.ftl
roleyfoley: I've put examples on the azure provider 
https://github.com/hamlet-io/engine-plugin-azure/pull/224#pullrequestreview-594903506

engine/base.ftl
ml019: Only one approach per `Qualifiers` attribute as the content (hash or sequence) then drives what structure is expected.

any is what we have now - we can transition away from it but it is there to supported existing qualification in cmdbs.

providers/shared/inputseeders/shared/id.ftl
ml019: OK - I wasn't aware we were that strict but I can align to that, noting that "provider" is a seeder. For instance, with the next set of changes re command line options, an entrance can provide a seeder.

roleyfoley:
> By the time getCompositeObject() sees the input, Qualifiers will have been applied and removed. You will note there is a stage for applying qualifiers.
> 
> We could still define their structure from a documentation perspective.

To keep the implementation inline with the documentation I don't think just having it in the documentation is going to work. What if Qualifiers were there own high level configuration similar to how we use Deployment Profiles? Or they were applied as part of the Deployment Profile processing? 

That way the profile would hold the configuration and it could be documented as part of the Deployment Profile implementation?

ml019:
> > Input sources now need to be explicitly defined.
> 
> Could you include in the description an example of this?

@RossMurr4y I've added further explanation to the PR.

ml019:
> > By the time getCompositeObject() sees the input, Qualifiers will have been applied and removed. You will note there is a stage for applying qualifiers.
> > We could still define their structure from a documentation perspective.
> 
> To keep the implementation inline with the documentation I don't think just having it in the documentation is going to work. What if Qualifiers were there own high level configuration similar to how we use Deployment Profiles? Or they were applied as part of the Deployment Profile processing?
> 
> That way the profile would hold the configuration and it could be documented as part of the Deployment Profile implementation?

I was trying to avoid having to add Qualifiers to every attribute definition and having every usage aware of them. They are also recursive so may end up spanning conventional boundaries we have now for getCompositeObject.

providers/shared/inputseeders/shared/id.ftl
ml019: Updated.

ml019:
@roleyfoley @RossMurr4y I've reflected our discussions into this PR.  Ready for another pass. The aws and azure PRs have also been updated to reflect the desired seeder macro naming.

roleyfoley (approved):

engine/base.ftl
```diff
@@ -1470,3 +1470,373 @@ A comparator is an operator and a version
...
+            [#-- Filters must have the same attributes --]
+            [#if
+                (matchBehaviour == EXACTLY_ONETOONE_FILTER_MATCH_BEHAVIOUR) &&
+                removeObjectAttributes(currentFilter, matchFilter?keys)?has_content]
+                [#return false]
+            [/#if]
+            [#return true]
+            [#break]
+
+            [#-- Unknown behaviour --]
```
RossMurr4y, line 1592: Perhaps a debug message here to help with troubleshooting when filter isn't being found and you aren't sure why?

ml019: done - throwing a fatal as we shouldn't ever get here

RossMurr4y (approved):

PR #1558, opened 2021-03-02
fix: allow for loading provider input replacement

roleyfoley:
## Description

This fix uses the unique combine configuration for providers to ensure that we build a list of providers instead of replacing it

## Motivation and Context
the input seeder process uses the mergeObject command to apply state changes as required. When merging the mergeObject behaviour replaces arrays as its hard to merge an array

The main reason for this fix is to handle the update of the provider names command line options when loading plugins from a layer/blueprint. Previously this was being overridden with whatever the shared provider set the value to 

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/providers/shared/inputseeders/shared/id.ftl
+++ b/providers/shared/inputseeders/shared/id.ftl
@@ -79,10 +79,14 @@
                     [#-- Deployment Details --]
                     "Deployment" : {
                         "Provider" : {
-                            "Names" : (providers!"")?has_content?then(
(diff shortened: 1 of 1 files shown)


ml019 (approved):
The shared provider is normally the first in the chain so shouldn't cause an issue but I think because we have yet to transition plugins into the chain, their input is being lost when the chain is rerun.

Issue #1577, opened 2021-03-13
Occurrence State Output


roleyfoley:
## Expected Behaviour

I can get a lit of all occurrences and subOccurrences as a single list which can be used to present data on occurrences and their states 

## Current Behaviour

The blueprint entrance does have some of this information but its nested within a district representation. SubOcccurences are listed as children of Occurrences which can make it difficult to include in queries where you want to include all Occurrences that exist

## Possible Solution

Create a new entrance which uses the component flow to output all occurrences 

## Context

Potential fix for https://github.com/hamlet-io/executor-python/issues/97

PR #1578, opened 2021-03-14
feat: add occurrences entrance

roleyfoley:
## Description

Closes: https://github.com/hamlet-io/engine/issues/1577

- Adds a new entrance to provide information specifically around occurrences.
- Adds a new output which provides the state of each occurrence in the district. To remove repetition subocurrences are removed from any occurrence which has a sub occurrence

## Motivation and Context

For occurrence queries in the python executor we use the blueprint output which is quite complicated to find all the occurrences. Providing a specific output makes it easy to query the occurrence details including finding suboccurrence details 

## How Has This Been Tested?

Tested locally

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [X] None of the above.

### DIFF
--- a/providers/shared/entrances/occurrences/id.ftl
+++ b/providers/shared/entrances/occurrences/id.ftl
@@ -1,0 +1,18 @@
+[#ftl]
+
+[@addEntrance
+    type=OCCURRENCES_ENTRANCE_TYPE
+    properties=
+        [
+            {
+                "Type"  : "Description",
+                "Value" : "Provides the state of all occurrences within a district"
+            }
+        ]
+    commandlineoptions=[
+        {
+            "Names" : "*",
+            "Types" : ANY_TYPE
+        }
+    ]
+/]
--- a/providers/shared/components/shared/setup.ftl
+++ b/providers/shared/components/shared/setup.ftl
@@ -17,6 +17,30 @@
     /]
 [/#macro]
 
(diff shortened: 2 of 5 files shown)


providers/shared/entrances/occurrences/id.ftl
```diff
@@ -0,0 +1,18 @@
+[#ftl]
+
+[@addEntrance
+    type=OCCURRENCES_ENTRANCE_TYPE
+    properties=
+        [
+            {
+                "Type"  : "Description",
+                "Value" : "Provides the state of all occurrences within a district"
```
RossMurr4y, line 9: It sounds like the use-case for this is primarily internal to Hamlet Deploy. 

The naming pattern we use for internal macros is not available to us here because of the dynamic macro evaluation but perhaps the Description here should reflect this?

RossMurr4y (approved):

providers/shared/entrances/occurrences/id.ftl
roleyfoley: I think the macros you might be referring too are actually internal to functionality within the engine not necessarily to hamlet deploy or even the engine itself. 

The occurrence structure is a fairly standard interface now so I think it would be safe to call its structure a stable public interface that people can use outside of hamlet if they want.

This isn't really that different to the blueprint entrance which includes the occurrence information anyway

PR #1583, opened 2021-03-16
feat: add support for engine output writing

roleyfoley:
## Description

Adds support for naming and writing outputs directly from the engine instead of having to invoke the hamlet engine for each output file 

The key change is the introduction of the output writer. The output writer is responsible for determining the file based details of the output being generated, which are the file name, directory and output type. The writer is a collection of output handlers which update a global property state of the file properties. 
   - The handlers are invoked at two stages of the output process, before generation which allows for file names to be used in outputs themselves, such as the generation contract and after output generation which is for writing the actual file and its contents to disk 

To start off we have 4 handlers 

- 2 cmdb functions which use the new toCMDB functions to write the content and discover the available file system 
- 2 which handle filename and directory assignment based on provided command line options 

Output file naming is now  controlled completely as part of the engine and the filename specified in the generation contract instead of just including the suffix and having the executor fill in the rest.

The generation contract has now been extended to provide all of the parameters required by the engine to generate an output. So by generating the filename within the engine we can then include this in the contract and the engine can read the output file name in the contract to output the new subset file. 

Output and output serialiser macros are now functions so that we can capture the content to pass to the outputwriter instead of using the standard freemarker output process 

Log messages are output using the freemarker output process and have been removed from all generated outputs. This is to ensure that outputs are always consistent with what they are expected to generate and allows us to always use JSON based formatting for log messages, including when generating sh based outputs 

A new task rename_file has been added to support naming the generation contract to align with the other outputs. This ensures that we can provide a generic name for the initial generation of the contract and align it with the engine defined name as required

## Motivation and Context

This gives the engine essentially full ownership over the CMDB and its contents in line with ADR-0001 
Also allows the engine to write multiple outputs on a single invocation which should reduce the time it takes to generate and handle outputs

## How Has This Been Tested?

Tested locally

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [X] None of the above.


roleyfoley:
**Note:** When using the engine to write outputs the order of the outputs is different to using the standard freemarker output. This means that there will be differences from a file perspective between old and new files

RossMurr4y:
> **Note:** When using the engine to write outputs the order of the outputs is different to using the standard freemarker output. This means that there will be differences from a file perspective between old and new files

So they will look the same, but git will interpret some invisible-to-the-eye difference between the two so on first upgrade, a CMDB will decide that every file has changed?

roleyfoley:
It wont be invisible the JSON ordering of keys will be completely different. Our diff process at the moment is doing a diff based on the contents of the file rather than JSON structure. So they will be all considered different and will look different to someone looking in the file. But the actual JSON content is still the same

RossMurr4y:
Works well alongside the corresponding PR's.

Looks like there's some conflicts to resolve though.

RossMurr4y (approved):

roleyfoley:
The test failure result is now as expected. In order to set the directory and filename the changes in https://github.com/hamlet-io/executor-bash/pull/177 are now required

RossMurr4y (approved):

PR #1576, opened 2021-03-13
feat: validate deployment mode provided

roleyfoley:
## Description

Fixes: https://github.com/hamlet-io/executor-python/issues/83

Enforces the use of defined deployment modes

- Any deployment mode that is used must now be defined in the blueprint before they can be used
- Adds two common deployment modes hibernate, and maintenance which are used in existing deployment profiles
- Removes the default deployment mode and the fall back used when getting details

## Motivation and Context

This provides better understanding of how deployment modes works and what the behaviour will be when they are used

## How Has This Been Tested?

Tested locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Refactor (non-breaking change which improves the structure or operation of the implementation)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Followup Actions

- [x] None

## Checklist:

- [ ] My change requires a change to the [documentation](https://github.com/hamlet-io/docs).
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [X] None of the above.

PR #1586, opened 2021-03-17
fix: define deploymentgroups for schemas

RossMurr4y:
## Description

The introduction of validation for DeploymentGroups at runtime has
caused schema generation to fail. Unless a DeploymentGroup is defined
within a blueprint, the option is invalid as a commandLineOption.

This commit introduces definitions for the groups used for JSONSchema
generation - component, reference, module and attributeset.

It also adds schema generation into the Jenkins pipeline to act as a test for catching this sort of thing in the future.

## Motivation and Context

#1576 

## How Has This Been Tested?

Locally - can now generate schemas the same as before introduction of #1576 

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,6 +12,20 @@ pipeline {
     }
 
     stages {
+
+        stage('Run Schemas Generation as Test') {
+            environment {
+                GENERATION_ENGINE_DIR="${WORKSPACE}"
+                SCHEMA_OUTPUT_DIR="${WORKSPACE}"
+            }
+
+            steps {
+                sh '''#!/usr/bin/env bash
+                    hamlet -i mock -p shared -p aws -p azure schema create-schemas -o "${SCHEMA_OUTPUT_DIR}"
+                '''
+            }
+        }
+
         stage('Run Shared Provider Tests') {
             environment {
                 GENERATION_ENGINE_DIR="${WORKSPACE}"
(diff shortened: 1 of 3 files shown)


roleyfoley:
I think its probably better that we move these checks to only run in the deployment entrance. Having extra deployment groups to work around the schema setup doesn't feel right. 

As an extra step they should probably be their own command line options to make it easier to follow.

roleyfoley (changes requested):
I think its probably better that we move these checks to only run in the deployment entrance. Having extra deployment groups to work around the schema setup doesn't feel right. 

As an extra step they should probably be their own command line options to make it easier to follow.

RossMurr4y:
> I think its probably better that we move these checks to only run in the deployment entrance. Having extra deployment groups to work around the schema setup doesn't feel right.

I've moved it over now.

> As an extra step they should probably be their own command line options to make it easier to follow.

Note sure what you are referring to. Do you mean that the schema group/unit should be a unique set of commandLineOptions so as to be distinct from component generation? Wont that be applicable to pretty much anything we generate that isn't components?

roleyfoley:
> Note sure what you are referring to. Do you mean that the schema group/unit should be a unique set of commandLineOptions so as to be distinct from component generation? Wont that be applicable to pretty much anything we generate that isn't components?

Yeah essentially as the commandline option says the value you are using is  deploymentUnit and deploymentGroup which aren't actually relevant to what you are doing. So instead of overloading one option and making it difficult to validate we should validate each one as required.

roleyfoley (approved):

PR #1622, opened 2021-04-07
feat(apigateway): image sourcing support

roleyfoley:
## Intent of Change

## Description

Adds support for specifying alternate image sources for apigateway

Also includes a none option which can be used in modules which specify the apidefintion and inject it as part of the blueprint processing

## Motivation and Context

Allows for external sources to be used for API Gateway definitions which simplifies the build process and also removes unnecessary pregeneraiton scripts when they aren't required

## How Has This Been Tested?

Tested locally

## Followup Actions

- [x] None

### DIFF
--- a/providers/shared/components/apigateway/id.ftl
+++ b/providers/shared/components/apigateway/id.ftl
@@ -244,6 +244,37 @@ object.
             {
                 "Names" : "Tracing",
                 "Children" : tracingChildConfiguration
+            },
+            {
+                "Names" : "Image",
+                "Description" : "Control the source of the image that is used for the function",
+                "Children" : [
+                    {
+                        "Names" : "Source",
+                        "Description" : "The source of the image - registry: the local hamlet registry - url: an external public url - none: no source image",
+                        "Types" : STRING_TYPE,
+                        "Mandatory" : true,
+                        "Values" : [ "registry", "url", "none" ],
+                        "Default" : "registry"
+                    },
+                    {
+                        "Names" : "Source:url",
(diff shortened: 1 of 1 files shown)


providers/shared/components/apigateway/id.ftl
```diff
@@ -244,6 +244,37 @@ object.
             {
                 "Names" : "Tracing",
                 "Children" : tracingChildConfiguration
+            },
+            {
+                "Names" : "Image",
+                "Description" : "Control the source of the image that is used for the function",
```
RossMurr4y, line 250: This description was applicable to `lambda` implementation of these attributes, but could use updated wording for the apigateway

providers/shared/components/apigateway/id.ftl
```diff
@@ -244,6 +244,37 @@ object.
...
+                        "Values" : [ "registry", "url", "none" ],
+                        "Default" : "registry"
+                    },
+                    {
+                        "Names" : "Source:url",
+                        "Description" : "Url Source specific Configuration",
+                        "Children" : [
+                            {
+                                "Names" : "Url",
+                                "Description" : "The Url to the lambda zip file",
```
RossMurr4y, line 266: here too

RossMurr4y (approved):
Ready to merge once you're happy with the description.

providers/shared/components/apigateway/id.ftl
roleyfoley: Fixed

PR #1625, opened 2021-04-08
feat: add attirbutes during extension processing

roleyfoley:
## Intent of Change

## Description

Updates extensions to handle the addition of extra attributes under a shared attribute along with extra values.
Adds a generic function to prefix attribute names, values and defaults across a whole set of attributes.

## Motivation and Context

Our current prefixing solution doesn't allow you to add new attributes under a shared attribute along and you can't pick and choose which values you support for a given attribute. This allows you to achieve both of these. 

The change has a little magic ( in the use of the prefix) but I think its straightforward enough to understand what is happening 

## How Has This Been Tested?

Tested locally 

## Followup Actions

- [x] None

### DIFF
--- a/engine/component.ftl
+++ b/engine/component.ftl
@@ -206,94 +206,6 @@
 
 [/#macro]
 
-[#function addPrefixToAttributes providerAttributes provider ]
-
-    [#-- This handles prefixing attributes and how they interact with shared attributes --]
-    [#-- The following rules are applied --]
-
-    [#-- If a provider adds an attribtue the name will be prefixed with the provider Id --]
-    [#-- If an attribute name is prefixed with the shared attribute then the prefix will be stripped and provider attribtue omitted --]
-    [#-- If values have been added to a provider attribute and the value is a string ( or list of ) then the provider is prefixed --]
-    [#-- If the value has been prefixed with the shared attribute then the prefix will be stripped and provider attribtue omitted --]
-    [#-- If the default value has been added on the provider attribute and the value is a string then the provider is prefixed --]
-    [#-- if the default has been prefixed with the shared attribute then the prefix will be stripped and the provider attribute omitted --]
-    [#-- When handling a list of values the shared and provider attributes need to be specified --]
-
-    [#local attributes = []]
-    [#list providerAttributes as providerAttribute ]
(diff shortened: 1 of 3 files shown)


roleyfoley:
Converting back to draft as this conflicts with https://github.com/hamlet-io/engine/pull/1391

engine/component.ftl
```diff
@@ -206,6 +206,94 @@
...
+                    [#else]
+                        [#local prefixedDefault = providerAttribute.Default ]
+                    [/#if]
+                [/#if]
+
+            [#else]
+
+                [#local prefixedNames += [ name?ensure_starts_with("${provider}:")]]
+
+                [#local prefixedValues = (providerAttribute.Values?map( x -> x?is_string?then( x?remove_beginning("${SHARED_ATTRIBUTES}:"), x ) ))![] ]
```
ml019, line 265: So if a provider specific name, should all the values be forced to be prefixed with the provider or is the assumptions that the values are unprefixed because the attribute is provider specific?

Also wondering what the intent of removing any "shared:" on the front was - when would this be provided? Or is this just ensuring any occurrence of the "shared:" prefix is always stripped off reagrdless?

ml019 (commented):
Wanting to understand a couple of aspects of the implementation.

Also, please some examples of what will break when PR is merged..

engine/component.ftl
roleyfoley: anything without the `shared:` prefix will be prefixed with the provider prefix. If you append an attribute with shared your essentially saying I want to use attributes from the shared provider but include my own. 

On further review of our code @RossMurr4y has already implemented his behaviour in extensions. So most of this code isn't actually required. I'm working on some updates in the extensions which will probably replace this PR and use extensions instead

roleyfoley:
Reworked this to now remove any broken changes. The functionality is essentially the same with the addition of being able to add extra attributes as required to a shared on. The function is a bit overkill for what the end result was, but it could be useful in the future

RossMurr4y (approved):
Looks good @roleyfoley

Looks like this will handle any that have been manually prefixed in the past (as you've used `?ensure_starts_with` so there shouldn't be any blockers to merging.

PR #1630, opened 2021-04-13
feat: Attribute set extensions

roleyfoley:
## Intent of Change

## Description

Adds support for creating extended attribute sets, extended attribute sets allow for overriding attribute Values, Defaults and AttributeSets along with adding prefixed attributes if they aren't part of the attribute set used as the base. 

## Motivation and Context

Attribute sets are really useful to share configuration and show that the same configuration is being used. However there are times when you might want to override default values at the provider level or add additional attributes across a provider. This allows you to do that and only add the overrides you require. 

## How Has This Been Tested?

Tested locally 

## Followup Actions

- [x] None

### DIFF
--- a/engine/attributeset.ftl
+++ b/engine/attributeset.ftl
@@ -18,6 +18,34 @@
     /]
 [/#macro]
 
+[#macro addExtendedAttributeSet baseType provider type pluralType properties attributes ]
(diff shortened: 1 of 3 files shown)


ml019 (approved):
Looks ok. I'll need to see some usages of this to fully understand the intent.

PR #1638, opened 2021-04-13
chore: add node 14 support for lambda

ml019:
## Intent of Change

## Description
Add the specific runtime value needed for using node 14 with lambda.

## Motivation and Context
Keeping up with AWS ...

## How Has This Been Tested?
Local template generation

## Followup Actions
- [x] None

### DIFF
--- a/providers/shared/components/lambda/id.ftl
+++ b/providers/shared/components/lambda/id.ftl
@@ -96,6 +96,7 @@
                     "nodejs8.10",
                     "nodejs10.x",
                     "nodejs12.x",
+                    "nodejs14.x",
                     "python2.7",
                     "python3.6",
                     "python3.7",


roleyfoley:
Should we just remove this value? seems like this is a bit of a maintenance burden, that we could fix with a note on the component or set a description. 


`For supported values see the Identifier value here ( https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html )`

roleyfoley (changes requested):
Should we just remove Values here? seems like this is a bit of a maintenance burden, that we could fix with a note on the component or set a description. 


`For supported values see the Identifier value here ( https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html )`

ml019:
@roleyfoley I think a sanity check that we are ok with the new version is still worthwhile. They are not changing so fast we can't keep up.

roleyfoley:
I can see where you are coming from at the hamlet development side, but as a user I think that just gets in the way, and has a potential impact on that first 20 minute window of using hamlet. 

if we miss a type then people will get a bad impression that we aren't keeping up. 

Are we actually doing anything to test that each version works? does any of our code actually worry about what the version is?

ml019:
Equally, if they get the value wrong, they have to wait until stack generation time. 

I've raised https://github.com/hamlet-io/engine/issues/1640 to cover off the broader issue/discussion with the team.

In the interim, I'd like to get this in so I can deploy stuff for ICS.

roleyfoley (approved):

Issue #1635, opened 2021-04-13
Move template log messages into engine


roleyfoley:
## Expected Behaviour

Log messages generated by the engine should be written by the engine removing the need for a secondary process to look for the messgaes

## Current Behaviour

Log messages are written to an output file which is then scanned to find messages. 

## Possible Solution

The hamlet engine-core ( freemarker wrapper ) now includes support for writing to stdout or stderr ( https://github.com/hamlet-io/engine-core/issues/45 ) 

When calling the message macro we can then write output as required

## Context

Part of the implementation of ADR-0001 ( https://github.com/hamlet-io/architectural-decision-log/blob/main/adr/0001-executor-engine-separation.md ) Which aims to isolate the engine from the execution process as much as possible

roleyfoley:
Completes as part https://github.com/hamlet-io/engine/pull/1645 

They have now been renamed to in-line logs and integrated into our current logging process

PR #1639, opened 2021-04-14
feat: console output and engine logging 

roleyfoley:
## Intent of Change

## Description

Adds support for the engine to handle the processing of log messages that it creates. Logging is handled through the output writer process using its own writer configuration. This allows for the writing of log messages to different locations than the output that is being requested. 

Adds a new output writer which can write messages to the console using either stdout or stderr 

The log message processing by default logs to both the console and to a log file provided by the executor and formats messages to align with the output type 

The outputProperties have been re-factored to support different output types, currently file and console to support outputs under a single process 

Example console output showing a header outlining what is being generated and a collection of logs that have been generated during processing : 

```
[*] entrance: deployment | output: contract | subset: generationcontract

 Hamlet Engine Logs
----------------------

{
  "Timestamp": "2021-04-14T11:21:32.188Z",
  "Severity": "fatal",
  "Message": "All is borked"
}

-----
```

## Motivation and Context

Implementation of #1635 

## How Has This Been Tested?

Tested locally 

## Followup Actions

- Requires: https://github.com/hamlet-io/engine-core/pull/51
- Requires: https://github.com/hamlet-io/engine/pull/1643


providers/shared/outputhandlers/write_to_cmdb/outputhandler.ftl
```diff
@@ -12,29 +12,60 @@
...
-    [#local fileProperties = getOutputFileProperties() ]
-    [#local fileFormat = (fileProperties["format"])!""]
-
-    [#-- An empty JSON object is considered empty content --]
-    [#-- Even if the content is empty we still write an empty file --]
-    [#if ! content?has_content ]
-        [#local content=""]
-        [#local fileFormat=""]
-    [/#if]
+    [#if properties["type"] == "file" ]
```
ml019, line 15: Is this a transitional thing or for internal use only? We've discussed the engine owning the cmdb so if coming from an external trigger, then should be through a specific entrnace e.g. for updating a setting or build reference.

I guess the generation contract needs this option anyway ...

ml019 (commented):
Use of writers makes for a nice abstraction.

Just need to confirm intention of cmdb writer is basically only for use in generation contracts?

providers/shared/outputhandlers/write_to_cmdb/outputhandler.ftl
roleyfoley: This will always be used but more than likely moved into the cmdb provider as a handler that it owns 

With the logging support I think it highlights the need for non CMDB methods i.e. toFile cpFile etc.

ml019: Agreed. I saw the new ticket for them.

RossMurr4y (approved):

PR #1645, opened 2021-04-17
feat: handle log based exceptions in engine

### LINKED ISSUE #1636: Move inline template error detection into the engine
## Expected Behaviour

When an inline error is found in an output ( `HamletFatal:` ) the engine should detect these errors and handle them appropriately 

## Current Behaviour

Once the engine has created the output an executor needs to search for this message in the output

## Possible Solution

Add an output handler which will search for this value in generated output. If its found an error should be logged the output discarded 

## Context

Part of the implementation of ADR-0001 ( https://github.com/hamlet-io/architectural-decision-log/blob/main/adr/0001-executor-engine-separation.md ) Which aims to isolate the engine from the execution process as much as possible

### LINKED ISSUE #1637: Handle fatal errors as part of engine processing
## Expected Behaviour

When fatal messages are generated during template generation the engine should handle logging the error, determining the output location and setting the appropriate status code for the engine pass

## Current Behaviour

All log events which occur during template generation are sent to a log file. Once template generation has completed an executor needs to look through this file to determine the result of the generation.

## Possible Solution

Combine #1635 #1636  and https://github.com/hamlet-io/engine-core/issues/44 to the engine to handle all errors as they come up during template generation processing. 

## Context

With hamlet-io/executor-bash#168 and hamlet-io/architectural-decision-log#5 we've identified the need for more complex error handling across hamlet in general. The current implementation relies on the executor to classify messages generated by the engine.

roleyfoley:
## Intent of Change

## Description

- Adds support for handling exceptions generated through log messages in the engine itself. This allows for the engine to stop processing if it finds logs which match the stop level
- Adds support for in-line log messages. Once an output has been generated the content of the output is scanned for the presence of Hamlet*: messages which will be included in the logMessages found during processing
- Updates the stop message action to use the logWriter process to output the context of the stop. This allows for extra logs to be included as required.

## Motivation and Context

This allows for the engine to define when things went wrong instead of relying on the executor to handle this. The executor does have some control over this through the stopLevel command line option but ultimately the engine says what happens during its processing.

Closes #1637  
Closes #1636

## How Has This Been Tested?

Tested locally 

## Followup Actions

- [x] None

### DIFF
--- a/client/invokeEntrance.ftl
+++ b/client/invokeEntrance.ftl
@@ -30,3 +30,9 @@
 [@writeLogs
     writers=getCommandLineOptions().Logging.Writers
 /]
+
+[#-- Find the logs for any that match the stop level or above --]
+[#-- Set the exit code if they are found --]
+[@setExitStatusFromLogs
+    failureLevel=getCommandLineOptions().Logging.StopLevel
+/]
--- a/engine/logging.ftl
+++ b/engine/logging.ftl
@@ -81,6 +81,11 @@
     [/#if]
 [/#macro]
 
+[#-- Level/Severity Handling --]
+[#function getLogLevelFromDescription severity ]
+    [#return (logLevelDescriptions?seq_index_of(severity))!0 ]
+[/#function]
+
 [#function willLog level ]
     [#return currentLogLevel <= level ]
 [/#function]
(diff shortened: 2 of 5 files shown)


roleyfoley:
When inline logs are detected the log message contains the immediate parent within the content where the object is found. 

- So if a value in a nested object is provided to be a log the object the value belongs to is logged instead of all the contents of the output. 
- When an array contains an inline message just the array is shown

ml019 (approved):

PR #1648, opened 2021-04-20
feat: allow configuration of the schema  property

### LINKED ISSUE #1551: [BUG] JSONSchemas have a hardcoded Id
## Expected Behaviour

Generated JSONSchemas should have an easily configurable Id value, as well as a sensible default

## Current Behaviour

- the current Id is incorrect as it points to the docs site prior to its move to docs.hamlet.io
- Generating other schemas for things like Modules that are not necessarily "official" will still contain links to the docs site

## Possible Solution

- update default value that points at the docs site
- set the value based on settings - that way the user can overwrite them for some, whilst not others as necessary.
  - this means there is no additional effort on behalf of the module author

RossMurr4y:
## Intent of Change

- new feature
- 
## Description

Allow the setting of commandLineOptions to control the JSONSchema `$id` property.

Defaults to the existing $id if unprovided.

## Motivation and Context

Closes #1551 

## How Has This Been Tested?

Locally

## Followup Actions

- [x] None

### DIFF
--- a/engine/schema.ftl
+++ b/engine/schema.ftl
@@ -82,12 +82,12 @@
     [#return asArray(composite.Names)[0]]
 [/#function]
 
-[#function formatSchemaId section unit version="latest"]
-    [#switch section]
-        [#default]
-            [#return formatPath(false, rootSchemaPath, version, "blueprint", "schema-" + section + "-" + unit + "-schema.json")]
-            [#break]
-    [/#switch]
+[#function formatSchemaId section unit version="latest" prefix="" ]
+    [#local filename = "schema-" + section + "-" + unit + "-schema.json" ]
+    [#if prefix?has_content ]
+        [#return formatPath(false, prefix, filename) ]
+    [/#if]
+    [#return formatPath(false, rootSchemaPath, version, "blueprint", filename)]
 [/#function]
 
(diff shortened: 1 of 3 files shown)


RossMurr4y:
@ml019  I am having trouble identifying where else I need to introduce these commandLineOptions such that they could be actually configured. 

It used to be reasonably straight forward but the recent introduction of configseeders and loaders has made adding new commandLineOptions quite an involved process. I would appreciate some direction here.

roleyfoley:
@Ross I think you want the input seeders for this and there are two places I could see it going 

- On the entrance seeder - https://github.com/hamlet-io/engine/blob/29bd56116e7b3180c5669bc43a59027126fd15b8/providers/shared/entrances/schema/entrance.ftl 
   - This would make that option available to the schema entrance only and make it obvious what this entrance expects 
   - You can also default the value in the Id for the schema entrance 
   
- As a default command line option for all seeding 
   - https://github.com/hamlet-io/engine/blob/9b7789329c155ce77fdef2f3e92c645928b3c074/providers/shared/inputseeders/shared/id.ftl#L93 
     - This would make it available across all entrances and included at load time 
     
Given this is really specific I'd recommend putting it on the entrance.

ml019 (approved):

PR #1654, opened 2021-04-22
refactor: move alert config into attribute set

roleyfoley:
## Intent of Change

## Description

Moves the alert configuration from a childConfiguration global variable to an attribute set

## Motivation and Context

Cleaner usage and understanding of the configuration 

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/component.ftl
+++ b/providers/shared/components/component.ftl
@@ -74,6 +74,8 @@
 
 [#assign GLOBALDB_COMPONENT_TYPE = "globaldb" ]
 
+[#assign HEALTHCHECK_COMPONENT_TYPE = "healthcheck" ]
+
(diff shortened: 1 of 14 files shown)


providers/shared/components/component.ftl
```diff
@@ -74,6 +74,8 @@
 
 [#assign GLOBALDB_COMPONENT_TYPE = "globaldb" ]
 
+[#assign HEALTHCHECK_COMPONENT_TYPE = "healthcheck" ]
```
ml019, line 77: Looks like this might have bled through from another change. Will let it through as the definition is harmless.

ml019 (approved):

PR #1658, opened 2021-04-22
feat: add raw id and name formatting

roleyfoley:
## Intent of Change

## Description

- Adds formatting based on the raw id and name values which don't split the id and name values based on - in the id

## Motivation and Context

This ensures unique names and Ids can be generated for all occurrences as these values must be unqiue within the JSON structure of the data

Allows for ensuring there is a unique set of identifiers for each occurrence which is required for  https://github.com/hamlet-io/executor-python/issues/117

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/occurrence.ftl
+++ b/engine/occurrence.ftl
@@ -218,19 +218,28 @@
             {
                 "Core" : {
                     "Id" : formatId(occurrence.Core.Extensions.Id),
+                    "RawId" : formatId(occurrence.Core.Extensions.RawId),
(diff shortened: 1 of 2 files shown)


ml019 (approved):
Based on the naming conventions that we follow, TypedId and raw id are generally the same thing. 

If a search shows the typedID variants not being used (my suspicion is they aren't), then I'd suggest removing them in favour of the raw id which then gives the consumer more freedom on component identifiers.

PR #1661, opened 2021-04-26
feat(ecs): add compute provider support on service

roleyfoley:
## Intent of Change

## Description

- Adds support for specifying the compute provider on the ecs service. When using compute providers the providers available depends on the engine. So we set the default provider to the provider recommended by the engine
- Removes vendor specific attributes from the container definitions and instead uses attribute extensions to define them in the provider

## Motivation and Context

This configuration allows for services to specify the compute provider configuration that they require specifically allowing for different services within the same container host to be hosted using different strategies 

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/component.ftl
+++ b/providers/shared/components/component.ftl
@@ -1315,7 +1315,7 @@
                         "Children" : [
                             {
                                 "Names" : "Provider",
-                                "Description" : "The default container compute provider - _engine uses he default provider of the engine",
+                                "Description" : "The default container compute provider - _engine uses the default provider of the engine",
                                 "Types"  : STRING_TYPE,
                                 "Values" : [ "_engine" ],
                                 "Default" : "_engine"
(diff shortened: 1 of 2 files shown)


providers/shared/components/component.ftl
```diff
@@ -1298,6 +1305,55 @@
...
+            {
+                "Names" : "ComputeProvider",
+                "Description" : "The compute provider placement policy",
+                "Children" : [
+                    {
+                        "Names" : "Default",
+                        "Children" : [
+                            {
+                                "Names" : "Provider",
+                                "Description" : "The default container compute provider - _engine uses he default provider of the engine",
```
ml019, line 1318:
```suggestion
                                "Description" : "The default container compute provider - _engine uses the default provider of the engine",
```

ml019 (changes requested):

ml019 (approved):

Issue #1666, opened 2021-04-28
Generate mutliple outputs during single generation pass


roleyfoley:
## Expected Behaviour

When invoking the hamlet engine for a particular entrance I can provide a collection of outputs to be generated within the scope of the entrance and they are all generated at once

## Current Behaviour

We currently create the generation contract and using this generation contract we invoke the engine for each required output. This requires reprocessing all input data each time and doesn't allow us to cache values that are used by all outputs

## Possible Solution

Support providing the generation contract to the engine itself. It would then look through the contract to find steps that it can do and run them. This would allow the engine to create multiple outputs as part of the same engine invocation. 

## Context

The primary reason for this change is to make the template generation process faster. There is a lot of setup and duplication in the template generation process that we could remove with this change.

PR #1671, opened 2021-05-01
feat: extend multipass generation support

roleyfoley:
## Intent of Change

## Description

- Adds a status field to contract steps to indicate  if they have already been run. Available values are completed, available and failed 
- Adds an output clear macro which is invoked between entrance passes to ensure the output isn't tainted between passes
- Adds support for adding entrancePasses during the generation contract pass. This allows for the engine to create extra outputs as part of the same call to create the generation contract
- refactors contract steps to allow for them to be passed to the entrancePasses as well as written to file
- removes command line input parameters which are managed by the contract input

## Motivation and Context

The main motivation here is to increase performance in output generation by reducing the repetition in the processing of outputs. By reusing the engine call as much as possible we can have more efficient cache usage and reduce startup processes 

This closes https://github.com/hamlet-io/engine/issues/1666 as we are now generating as many templates as possible within a single pass

This change brings the aws test suite timing down to 

```
real    8m2.570s
user    11m9.057s
sys     2m8.182s
``` 

from 

```
real    9m38.630s
user    11m33.315s
sys     2m16.764s
```

in https://github.com/hamlet-io/executor-bash/pull/204

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- None

### Dependent PRs:

- https://github.com/hamlet-io/executor-bash/pull/208

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/deploymentframeworks/default/output.ftl
+++ b/providers/shared/deploymentframeworks/default/output.ftl
@@ -481,7 +481,7 @@
 
             /]
 
-            [#-- If pregenration is not reqiured we can complete this pass as part of this entrance invoke --]
+            [#-- If pregeneration is not required we can complete this pass as part of this entrance invoke --]
             [#if ! pregeneration ]
                 [@addEntrancePass
                     contractStep=
--- a/providers/shared/deploymentframeworks/default/output.ftl
+++ b/providers/shared/deploymentframeworks/default/output.ftl
@@ -298,6 +298,7 @@
                                                     "Id" : step.Id,
                                                     "Type" : step.Type,
                                                     "Priority" : step.Priority,
+                                                    "Status" : step.Status,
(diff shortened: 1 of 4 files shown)


providers/shared/deploymentframeworks/default/output.ftl
```diff
@@ -422,16 +461,40 @@
...
-                    )
+                parameters=stepParameters
+                status=(pregeneration)?then(
+                            "available",
+                            "completed"
+                        )
+
             /]
+
+            [#-- If pregenration is not reqiured we can complete this pass as part of this entrance invoke --]
```
ml019, line 484:
```suggestion
            [#-- If pregeneration is not required we can complete this pass as part of this entrance invoke --]
```

ml019 (dismissed):

PR #1672, opened 2021-05-02
Yaml config contract properties

roleyfoley:
## Intent of Change

## Description

- Adds support for generation converters which allow for generating outputs based on one format and then have the output writer convert it to another one. The  primary usage for this is converting from json to yaml since the generated output is the same but just needs to be converted 
- Adds a more specific configuration option for alternative outputs to allow for applying alternatives to specific subsets instead of all of them 
- Adds a properties section to contracts which contain contextual level information that can be applied to all stages in the contract or provide executors with extra information 
- Refactors the process_template_pass outputs to only include the ones required 
- refactors standard generation steps into a single stage instead of multiple ones per alternative. This makes the processing of outputs more efficient and easier to follow 
- Adds image support for adapators

## Motivation and Context

converters allow for JSON and yaml based config files to be created to support different tools that would use them. The refactor of the contract output makes it faster to generate outputs and to reduce the number of outputs required 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- https://github.com/hamlet-io/engine-plugin-aws/pull/291

### Consumer Actions:

- None

### DIFF
--- a/engine/output.ftl
+++ b/engine/output.ftl
@@ -554,7 +554,7 @@
         "outputConversion"       : outputMappings["OutputConversion"],
         "pass"                   : subset,
         "passAlternative"        : alternative,
-        "deploymentUnitSubset"   : outputMappings["Subset"],
+        "deploymentUnitSubset"   : deploymentUnitSubset,
         "outputFileName"         : getOutputFileName(subset, alternative, converter)
     }]
 [/#function]
--- a/engine/output.ftl
+++ b/engine/output.ftl
@@ -367,6 +367,7 @@
 
 [#-- Generation Contracts --]
 [#assign generationcontractStepOutputMappings = {} ]
+[#assign generationcontractStepOutputMappingConverters = {}]
 
 [#macro addGenerationContractStepOutputMapping provider subset outputType outputFormat outputSuffix]
     [#assign generationcontractStepOutputMappings = mergeObjects(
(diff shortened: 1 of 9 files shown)


engine/output.ftl
```diff
@@ -458,22 +499,43 @@
...
+        {
+            "entrance"               : getCLOEntranceType(),
+            "flows"                  : getCLOFlows()?join(","),
+            "providers"              : (getLoaderProviders()?join(","))!SHARED_PROVIDER,
+            "deploymentFramework"    : getCLODeploymentFramework(),
+            "deploymentUnit"         : getCLODeploymentUnit(),
+            "deploymentGroup"        : getCLODeploymentGroup(),
+            "account"                : getCommandLineOptions().Layers[ACCOUNT_LAYER_TYPE],
+            "accountRegion"          : contentIfContent(getCLOAccountRegion(), getAccountLayerRegion()),
+            "region"                 : contentIfContent(getCLOSegmentRegion(),getProductLayerRegion()),
```
ml019, line 522: Is this just interim? With multi-placement, the account/region can be different per resourceGroup so better to leave on the pass. Same for region.

roleyfoley: I went with reality, we can just move it when that is the case

ml019: Ok - I expect that will be the case for the placement stuff but happy for this to go through for now.

ml019 (approved):

PR #1674, opened 2021-05-05
fix: remove global settings assignement

roleyfoley:
## Intent of Change

## Description

removes a global assignment of the settings var which was overriding the composite settings global variable

## Motivation and Context

This was overriding the composite settings var that is used during input processing 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019 (approved):
oops

PR #1678, opened 2021-05-07
fix: fix device id for linked EBS volumes

kshychko:
## Intent of Change

## Description

Add "/dev/" prefix to DeviceId in DataVolumeMount macro.
## Motivation and Context

Specified DeviceId in macro goes without "/dev/" prefix, but when VolumeAttachment resource created, it is appended there.
In order to align the values, the DataVolumeMount macro needs to add a prefix to the specified deviceId. 
## How Has This Been Tested?

Locally.
### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/extensions/extension.ftl
+++ b/providers/shared/extensions/extension.ftl
@@ -375,7 +375,7 @@
                 (_context.VolumeMounts!{}) +
                 {
                     volumeLinkId : {
-                        "DeviceId" : deviceId,
+                        "DeviceId" : "/dev/" + deviceId,
                         "MountPath" : mountPath
                     }
                 }


roleyfoley:
@kshychko I'm not quite sure this is the best way to go and might be better to require that the id provided is the full cloud providers Id. So for aws this would be `/dev/sds` 

This allows the datavolume macro to work for providers which might use different syntaxes. 
https://github.com/hamlet-io/engine-plugin-aws/pull/296 includes a fix for the volume attachment  in AWS which was always adding `/dev/ to the start of the id and now just does an ensure_starts_with 

So I think we can close this one in favour of that

PR #1716, opened 2021-06-25
refactor: remove direct references to region

ml019:
## Intent of Change

## Description
Replace references to the command line option `region` with references to the `regionId` global variable set in setContext.

## Motivation and Context
There is a mix of use of `region` and `regionId` reflecting the fact that the composite assembly process initialises `region` while
`getContext` populates regionId.

With dynamic loading, references to region need to be made more dynamic so removal of direct references to the input value is the first step in preparing for this.

Once all known uses are modified, the variables used to pass in the layer values will be changed which will hopefully flush out any unidentified uses via null/missing exceptions.

## How Has This Been Tested?
The change is straightforward but the most efficient way to confirm is the running of templates in the normal course of business and rapidly addressing any exceptions.

There are a number of situations where `region` is actually a local variable in a function so global replace isn't an option.

## Related Changes
- https://github.com/hamlet-io/engine-plugin-aws/pull/349
- https://github.com/hamlet-io/engine-plugin-azure/pull/267

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/legacy/account/account_console.ftl
+++ b/legacy/account/account_console.ftl
@@ -110,7 +110,7 @@
                         r'  create|update)',
                         r'      info "Cleaning documents which might not have been created by CFN"',
                         r'      cleanup_ssm_document' +
-                        r'      "' + region + r'" ' +
+                        r'      "' + regionId + r'" ' +
                         r'      "' + consoleSSMDocumentName + r'" ',
                         r'      ;;',
                         r'esac'
--- a/legacy/account/account_ecs.ftl
+++ b/legacy/account/account_ecs.ftl
@@ -16,7 +16,7 @@
 
     [#assign ecsAccountCommands = [] ]
     [#list ecsAccountSettings as setting,state ]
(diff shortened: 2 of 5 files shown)


roleyfoley (approved):

ml019:
@roleyfoley tests failing. 

A couple of errors during install of the cli - don't think they are the issue
```
ERROR: aws-sam-translator 1.36.0 has requirement six~=1.15, but you'll have six 1.14.0 which is incompatible.
ERROR: cfn-lint 0.51.0 has requirement pyyaml>=5.4; python_version != "3.4" and python_version != "3.5", but you'll have pyyaml 5.3.1 which is incompatible.
```

Then it can find the jar file.  IS this related to the new ENV VAR needed to locate the jar file?

ml019:
@roleyfoley this needs another review. Tests passing now so should merge as soon as you approve.

roleyfoley (approved):

PR #1725, opened 2021-07-01
feat: subcomponent linking

ml019:
## Intent of Change

## Description
Introduce a `SubComponent` attribute on links that will replace the use of component type specific link attributes. For now, either will work.

Also expand the checking of component types and the use of the `Type` attribute, particularly as a check on the type of the expected link target and to differentiate between subcomponents with the same id
but different types.

During testing, a bug was discovered in the cacheing of occurrences which wasn't type aware. This has been added to support the subcomponent linking changes.

## Motivation and Context
- avoid an ever growing list of subcomponent attributes
- more stable link schema
- aligns with 

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/occurrence.ftl
+++ b/engine/occurrence.ftl
@@ -224,6 +224,7 @@
                     "Name" : formatName(occurrence.Core.Extensions.Name),
                     "RawName" : formatName(occurrence.Core.Extensions.RawName),
                     "TypedName" : formatName(occurrence.Core.Extensions.Name, type),
+                    "TypedRawName" : formatName(occurrence.Core.Extensions.RawName, type),
                     "FullName" : formatSegmentFullName(occurrence.Core.Extensions.Name),
                     "RawFullName" : formatSegmentFullName(occurrence.Core.Extensions.RawName),
                     "TypedFullName" : formatSegmentFullName(occurrence.Core.Extensions.Name, type),
--- a/engine/occurrence.ftl
+++ b/engine/occurrence.ftl
@@ -40,11 +40,12 @@
         [#local instanceId = parentOccurrence.Core.Instance.Id ]
         [#local versionId = parentOccurrence.Core.Version.Id ]
         [#local subComponentId = occurrence.Core.SubComponent.Id ]
+        [#local subComponentType = occurrence.Core.Type ]
         [#local subInstanceId = occurrence.Core.Instance.Id ]
         [#local subVersionId = occurrence.Core.Version.Id ]
 
         [#local occurrenceDetails =
-            (occurrenceCache[tierId][componentId][instanceId][versionId]["SubComponents"][subComponentId][subInstanceId][subVersionId])!{} ]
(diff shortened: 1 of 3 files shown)


ml019:
@roleyfoley would you mind trying this with one of the solutions where you expect id clashes could be an issue?

roleyfoley:
Turns out I hadn't actually done it before. But there is a bigger issue the two components can't be named the same. One overrides the other so only one subcomponent under either type can exist. 

Example component

```
                "mgmt-ecs" : {
                    "ECS" : {
                        "Instances" : {
                            "default" : {
                                "Enabled" : true,
                                "deployment:Unit" : "mgmt-ecs"
                            }
                        },
                        "Services" : {
                            "hello" : {
                                "deployment:Unit" : "hello",
                                "Instances" : {
                                    "default" : {
                                    }
                                },
                                "NetworkMode" : "awsvpc",
                                "Engine" : "fargate",
                                "DesiredCount" : 2,
                                "Placement" : {
                                    "ComputeProvider" : {
                                        "Default" : {
                                            "Provider" : "_engine"
                                        }
                                    }
                                },
                                "Containers" : {
                                    "_hello" : {
                                        "Image": {
                                            "Source": "containerregistry",
                                            "Source:containerregistry": {
                                                "Image": "hamletio/hello:latest"
                                            }
                                        },
                                        "Extensions" : [ "_noenv"],
                                        "Cpu" : 256,
                                        "MemoryReservation" : 512,
                                        "MaximumMemory" : 512
                                    }
                                }
                            }
                        },
                        "Tasks" : {
                            "hello" : {
                                "deployment:Unit" : "hello-task",
                                "Instances" : {
                                    "default" : {
                                    }
                                },
                                "NetworkMode" : "awsvpc",
                                "Engine" : "fargate",
                                "Containers" : {
                                    "_hello" : {
                                        "Image": {
                                            "Source": "containerregistry",
                                            "Source:containerregistry": {
                                                "Image": "hamletio/hello:latest"
                                            }
                                        },
                                        "Extensions" : [ "_noenv"],
                                        },
                                        "Cpu" : 256,
                                        "MemoryReservation" : 512,
                                        "MaximumMemory" : 512
                                    }
                                }
                            }
                        }
                    }
                },
```


When generating the list of occurrences I can only find the service 

```
hamlet component list-occurrences 
```

only lists one
```
 | mgmt     | mgmt-ecs      | management-mgmt-ecs         | ecs                |
| mgmt     | mgmt-ecs      | management-mgmt-hello       | service            |
```

And the occurrence entrance that lists that only has one entry as well

ml019:
@roleyfoley the occurrence cache wasn't type aware. I've added that and the cli now correctly lists the instances.

I also tested the detection of ambiguous subcomponent linking with the addition of a type correctly resolving the link.

I think this is ready for review.

NOTE: we will need to revisit the occurrence cache as part of implementation of ADR0007.

ml019:
@roleyfoley looking at the python executor, it is currently relying on the RawName in the occurrence structure but to avoid subcomponent conflicts, needs a typed variant of this. I've added that in this PR.

roleyfoley (approved):

PR #1728, opened 2021-07-06
feat: define registries on account layer

roleyfoley:
## Intent of Change

## Description

- Adds definitions of registries to layers so that they are always defined and know to the engine. This includes defining a storage type on each named registry to know how each of the images are stored. The initial implementation aligns with our current structure where only one storage backend is supported for each registry 
- Adds a releaseinfo entrance and corresponding view which uses the registry information to generate the environment variables used by the automation executor scripts to determine where images are hosted

## Motivation and Context

The main motivation behind this change is to simplify the configuration required by users when dealing with image management processes and removes the need to configure the properties file to provided this information

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


roleyfoley:
This will need to be reworked to align with the changes in 
https://github.com/hamlet-io/architectural-decision-log/blob/main/adr/0007-placement-support.md 

But this will provide a cleaner migration process and solves immediate issues

ml019 (approved):

PR #1730, opened 2021-07-07
fix: qualification of layer data

### LINKED ISSUE #1714: [BUG] Layer configuration ignoring qualifiers
## Current Behaviour
Qualifiers are ignored within layer objects

## Expected Behaviour
Qualifiers should apply to layers

## Possible Solution
Apply same logic as currently applied to the overall solution when determining layer configuration

## Steps to Reproduce
1. Add a `Qualifiers` attribute to the product layer to change the certificate behaviour 
2. Run template generation
3. Qualifiers will be ignored and default values used
4.

ml019:
## Intent of Change

## Description
Permit the use of qualification within layer data.

Rework the handling of layers to permit their data to provide the filter inputs (Id, Name) to their own qualification.

Layer data is processed once to work out the values to use when filtering it, then a second time to work out the effects of qualification on the remaining configuration data.

As part of the refactoring, the active state of layers in now part of the input state and calls to `getActiveLayers` now returns the data in the input state.

## Motivation and Context
Readiness preparation for [ADR0007](https://github.com/hamlet-io/architectural-decision-log/blob/main/adr/0007-placement-support.md)
Fixes #1714 

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/provider.ftl
+++ b/engine/provider.ftl
@@ -27,10 +27,10 @@
     [/#if]
 [/#macro]
 
-[#function getActivePluginsFromLayers  ]
+[#function getActivePluginsFromLayers  layersState ]
 
     [#local plugins = [] ]
-    [#local possiblePlugins = asFlattenedArray(getActiveLayerAttributes( [ "Plugins" ] )) ]
+    [#local possiblePlugins = asFlattenedArray(getActiveLayerAttributes( [ "Plugins" ], ["*"], [], layersState )) ]
 
     [#list possiblePlugins as pluginInstance ]
         [#list pluginInstance as id, plugin ]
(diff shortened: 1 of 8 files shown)


roleyfoley (dismissed):

roleyfoley:
@ml019 Looks some conflicts from an existing PR

engine/provider.ftl
```diff
@@ -158,46 +158,6 @@
     [#return result]
 [/#function]
 
-[#function getPluginsFromLayers pluginState ]
```
roleyfoley, line 161: Where has this moved to? Couldn't see a reference to it anymore?

ml019: I couldn't find any use for it - I think it got superceded when the installation  of providers moved to the input pipeline.

I removed it as unused code but can add back in if you think there is a need. I will need to adjust it to correctly deal with the layerState depending on whether it is call inside or outside the input pipeline

roleyfoley: All good didn't realise that it was subsumed. All good

roleyfoley (approved):

PR #1737, opened 2021-07-17
feat: set districts for deployment groups

roleyfoley:
## Intent of Change

## Description

- Adds a district property to deployment groups to scope the deployments included as part of the group
- Includes district details in the management contract
- Includes all deployment groups in the management contract instead of filtering on those that have created outputs
- Adds support for resource sets that do not act as subsets and instead are treated as standalone deployment units
- reorder default deployment group priorities to accommodate other groups

## Motivation and Context

This allows for discovering and listing all available deployments across the hamlet districts and to select which ones you want to use

## How Has This Been Tested?

Tested locally and on development environment

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/references/DeploymentGroup/id.ftl
+++ b/providers/shared/references/DeploymentGroup/id.ftl
@@ -29,6 +29,12 @@
             "Mandatory" : true
         },
         {
+            "Names" : "District",
+            "Description" : "The id of the district for components included in the deployment group",
+            "Types" : STRING_TYPE,
+            "Mandatory" : true
+        },
+        {
             "Names" : "OutputPrefix",
             "Description" : "Overrides the prefix used when generating outputs - Defaults to the Id of the Group",
             "Types" : STRING_TYPE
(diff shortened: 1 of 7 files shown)


providers/shared/references/DeploymentGroup/id.ftl
```diff
@@ -28,6 +28,12 @@
             "Values" : [ "", "account", "segment", "solution", "application" ],
             "Mandatory" : true
         },
+        {
+            "Names" : "District",
+            "Description" : "The id of the district for components included in the deployment group",
```
ml019, line 33: Need to be consistent with our terminology. "The id of the district" is "segment" not the id of the solution within the district right? I think this is how we've used it up until now and how it is reflect in recent changes to the cli.

When we come to a solution within a district, that's when we talk about the id of the solution.

roleyfoley: Good point, currently deployment groups live outside of the solution as part of the reference data. District seemed the best fit here so that the groups can be used across multiple solutions but they are still tied to a given district. This ensures that you don't get overlap and can define the deployment easily

ml019: all good.

ml019 (approved):

Issue #159, opened 2018-03-09
Add Maintenance Window configuration to RDS instances


roleyfoley:
Maintenance windows for RDS define when minor work can be scheduled and performed by AWS. 
By default the maintenance window is set by AWS and can be different for each RDS instance. So that we can report and identify maintenance windows from the template configuration we should set the window using the PreferredMaintenanceWindow parameter. 

Since maintenance windows can be used on other services, we should have a codeontap maintenance window configuration which staggers across the environments, segments and tiers. 

- Environments: staggered across days of the week
- Segments: staggered across half day of the environment day (morning/night) 
- Tiers: staggered by the hour of the segment half day 

This way updates are applied on a predictable, rolling basis.

roleyfoley:
@ml019  Thoughts? 
Not sure if this is too complicated for what we would want, but this is the maintenance window configuration I was thinking of defaulting to for the SSM components as well.

ken9ross:
Regarding the RDS maintenance windows
Current state is:
- aws:RDS will be updated on a random day at a random 30 minute block starting time in the range 12:00–20:00 UTC (if in ap-southeast-2)
- azure:RDS will be updated any day 1700-0800 local time
- aws:QueueHost already has a MaintenanceWindow defined (which while not the correct format for RDS, can be translated so I think should be used as is as a structure to be consistent)

In the case of an Aurora cluster, for the cluster, with each DB instance being scheduled for 1 hour thereafter. If not set it would use the current AWS default as per above.
In the case of azure:RDS will only make use of the day as there are only two options (other than default) Mon-Thu 1700-0800 or Fri 1700-0800 local time
This is not what you were asking for in the ticket, but I don't understand your objective with the segment and tier being included in the calculation of the timewindows

BTW The current default for QueueHost is Sunday 1000AEST, is that intended?

roleyfoley:
So having looked at this again ( 3 years down the track ) I think what was suggested is too complicated for implementation within hamlet. 

We should ensure that components that have maintenance windows can be configured in a standard way. Users can then use deployment profiles to develop their own scheduling processes 

So I think requirements here are 

- Times set in UTC across the board 
- Determine a common pattern to use for scheduling inline with what the providers offer. If this isn't the case we should make a broader configuration pattern within hamlet to define schedules which would give us the ability transform this into the appropriate formats the providers require.

roleyfoley:
So I think we could merge this into an epic around maintenance window for managed IaaS services 

- db 
[...]

PR #1741, opened 2021-07-20
Add standard date functions and (optional) MaintenanceWindow to RDS

ken9ross:
## Intent of Change

## Description
Standard date manipulation functions added
Standard MaintenanceWindow child added to RDS

## Motivation and Context
Precondition for PR to address https://github.com/hamlet-io/engine/issues/159

## How Has This Been Tested?
Locally

### Prerequisite PRs:

- None

### Dependent PRs:

- TBA

### Consumer Actions:

- Functionality to be supported within provider plugins for MaintenanceWindow

### DIFF
--- a/providers/shared/components/cache/id.ftl
+++ b/providers/shared/components/cache/id.ftl
@@ -50,6 +50,10 @@
                 ]
             },
             {
+                "Names" : "MaintenanceWindow",
+                "AttributeSet" : MAINTENANCEWINDOW_ATTRIBUTESET_TYPE
+            },
+            {
                 "Names" : "Profiles",
                 "Children" :
                     [
--- a/engine/base.ftl
+++ b/engine/base.ftl
@@ -689,6 +689,72 @@ are added.
     [#return (end?long - start?long)]
 [/#function]
 
+[#function addDateTime dateTime unit amount]
+    [#switch unit]
+        [#case "dd"]
+            [#local offset=1000*60*60*24]
+            [#break]
+        [#case "hh"]
+            [#local offset=1000*60*60]
+            [#break]
+        [#case "mm"]
+            [#local offset=1000*60]
+            [#break]
+    [/#switch]
+    [#local retval = (dateTime?datetime?long + amount?long*offset?long)?number_to_datetime]
+    [#return retval ]
+[/#function]
+
+[#function showDateTime dateTime format timezone="UTC" ]
+    [#local old_time_zone = .time_zone]
+    [#setting time_zone = timezone]
+    [#local retval = dateTime?datetime?string(format)]
+
+    [#setting time_zone = old_time_zone]
+    [#return retval ]
+[/#function]
+
+[#function convertDayOfWeek2DateTime dayOfWeek timeofDay="00:00" timeZone="UTC"]
+    [#-- 
+    The explicit dates below are used to convert a "dayOfWeek" to a date to allow for date and 
+    TimeZone manipulations.
+    Any date sufficiently far away from leap year impacts can be used 
+    --]
+    [#switch dayOfWeek]
+        [#case "Monday"]
+            [#local startTime = "05-JUL-2021 "+timeofDay+":00 "+timeZone ]
+            [#break]
+        [#case "Tuesday"]
+            [#local startTime = "06-JUL-2021 "+timeofDay+":00 "+timeZone ]
+            [#break]
+        [#case "Wednesday"]
+            [#local startTime = "07-JUL-2021 "+timeofDay+":00 "+timeZone ]
+            [#break]
+        [#case "Thursday"]
+            [#local startTime = "08-JUL-2021 "+timeofDay+":00 "+timeZone ]
+            [#break]
+        [#case "Friday"]
+            [#local startTime = "09-JUL-2021 "+timeofDay+":00 "+timeZone ]
+            [#break]
+        [#case "Saturday"]
+            [#local startTime = "10-JUL-2021 "+timeofDay+":00 "+timeZone ]
+            [#break]
+        [#case "Sunday"]
+            [#local startTime = "11-JUL-2021 "+timeofDay+":00 "+timeZone ]
+            [#break]
+    [/#switch]
+    [#local retval = startTime?datetime("dd-MMM-yyyy HH:mm:ss z") ]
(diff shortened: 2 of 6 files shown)


engine/common.ftl
```diff
@@ -1,5 +1,59 @@
...
+            [#break]
+        [#case "Friday"]
+            [#local startTime = "09-JUL-2021 "+timeofDay+":00 "+timeZone ]
+            [#break]
+        [#case "Saturday"]
+            [#local startTime = "10-JUL-2021 "+timeofDay+":00 "+timeZone ]
+            [#break]
+        [#case "Sunday"]
+            [#local startTime = "11-JUL-2021 "+timeofDay+":00 "+timeZone ]
+            [#break]
```
roleyfoley, line 51: What are these values for with the current date? Could we instead use ?now which gets the current time?

providers/shared/components/cache/id.ftl
```diff
@@ -49,6 +49,32 @@
...
+                            "Saturday"
+                        ]
+                    },
+                    {
+                        "Names" : "TimeOfDay",
+                        "Types" : STRING_TYPE
+                    },
+                    {
+                        "Names" : "TimeZone",
+                        "Types" : STRING_TYPE
```
roleyfoley, line 74: Should we default the TimeZone to UTC to make it easier to follow?

engine/common.ftl
```diff
@@ -1,5 +1,59 @@
 [#ftl]
```
ml019, line 1: If generic, these functions should be in base.ftl rather than common.ftl.

providers/shared/components/cache/id.ftl
```diff
@@ -49,6 +49,32 @@
                     }
                 ]
             },
+            {
```
ml019, line 52: Feels like this should be an attribute set given it is turning up in multiple places.

ml019 (dismissed):

ml019:
@ken9ross a few suggestions...

engine/common.ftl
ken9ross: These are to convert a "day" to a date which is sufficiently far away from any leap years etc to allow for timezone manipulations and date arithmetic (eg Saturday 01:00AEST to become Friday 15:00UTC)

providers/shared/components/cache/id.ftl
ken9ross: Maybe I configured it incorrectly, but if I set a default on the children o, I was then explicitly setting the maintenance window and couldn't leave it to AWS to configure an "appropriate" window (comment mostly applies to RDS which use the region to determine "appropriate")

engine/common.ftl
```diff
@@ -1,5 +1,59 @@
...
 
+[#-- DateTime functions --]
+[#function DateTimeAdd dateTime unit amount]
```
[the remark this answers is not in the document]
ken9ross: Cool

engine/common.ftl
ken9ross: Done

providers/shared/components/cache/id.ftl
ken9ross: Done but left the queuehost which has a default as an explicit definition

engine/common.ftl
roleyfoley: I think we need a comment on this as to why we have added fixed dates, its a bit weird to see specific dates in the code is all.

providers/shared/components/cache/id.ftl
roleyfoley: Thats the isConfigured check on the Maintenance window right? I.e. if someone has configured a maintenance window the defaut TimeZone is to use UTC.

providers/shared/components/cache/id.ftl
ml019: You can override the default attributeset with a default value in the queuehost.

engine/common.ftl
ken9ross: Cool, done

providers/shared/components/cache/id.ftl
ken9ross: I don't see a function isConfigured. Is it in master? I have a isLayerConfigured and a isConfigurationIncluded but neither seems to be what you're talking about

Also I didn't think there was a difference between something set in a solution file, inherited from a shared solution file or set by a default, when it came to processing. Am I missing something?

ml019: IsConfigured is added at run-time to say whether the parent has explicitly been defined. Do a global search in the aws provider and you'll see the usage.

ken9ross: The function is isPresent and the property is .Configured (had a chat with Roley at the standup). I'm leaning towards .Configured, because isPresent relies on an Enabled:true as well and you can't disable maintenance windows, so I think that'd be misleading

The more I think about this, the more I'm in favour of making the day, time and timezone mandatory with no defaults rather than having someone specify to do maintenance on Wednesday at 01:00 and have it trigger at 11:00 due to not specifying the timezone and picking up the UTC default. Thoughts?

If a child attribute has Mandatory:true does that force the existence of the parent, or is the child attribute only mandatory if the parent is defined? ie Can I enforce using Mandatory:true or do I need to explicitly code a fatal error for a partial maintenance window specification?

roleyfoley: Mandatory applies to the whole configuration object not just a child configuration. So if the property is conditional then you need to handle it outside of the configuration process

ken9ross: Cool, added standard function testMaintenanceWindow to handle including Azure v AWS options based on current preview

ken9ross:
This looks like something's gone screwy with the rebase. @roleyfoley Can I beg for assistance to figure out what's happened here?

roleyfoley:
@ken9ross The branch needed to be rebased to realign the changes made in the master branch and this branch

roleyfoley (approved):

roleyfoley:
@ml019 could you please approve this one if you are happy with it. Since you asked for changes you need to say they are done

ken9ross:
@ml019 Updated to the conventional commit syntax

roleyfoley (approved):

PR #1742, opened 2021-07-25
fix: duplicate fragment processing

ml019:
## Intent of Change

## Description
Remove the direct processing of fragment files.

## Motivation and Context
Input processing now includes fragments in the input state for both composite and dynamic cmdb processing. Currently for composite processing, the fragment is processed twice which can result in errors if the fragment logic is not idempotent.

## How Has This Been Tested?
Local template generation.

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/extension.ftl
+++ b/engine/extension.ftl
@@ -166,11 +166,6 @@
 
         [#-- Legacy support for fragments --]
         [#local fragmentId = id ]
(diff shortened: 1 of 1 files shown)


ml019:
@forcing commit as need to get update to records.digital out tonight.

PR #1784, opened 2021-08-31
refactor: add Hostname as an alternate to Certificate

roleyfoley:
## Intent of Change

## Description
adds an alternate name on all Certificate Configuration sections as Hostname.

## Motivation and Context

This aligns more with the configuration that the configuration offers to the user with the Certificate lookup we perform within the engine a consumer of this configuration

## How Has This Been Tested?

tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/apigateway/id.ftl
+++ b/providers/shared/components/apigateway/id.ftl
@@ -149,7 +149,7 @@ object.
                 ]
             },
             {
-                "Names" : "Certificate",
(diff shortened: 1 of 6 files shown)


ml019 (approved):
Worth reversing the order in a future PR but that will involve implementation changes.

PR #1792, opened 2021-09-07
refactor: use SubComponent in links

ml019:

## Description
Replace use of component specific keys in links with the more generic "SubComponent" attribute.

## Motivation and Context
Component specific link attributes are deprecated.

## How Has This Been Tested?
Local template generation.

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/baseline/id.ftl
+++ b/providers/shared/components/baseline/id.ftl
@@ -73,24 +73,6 @@
 
     [#list baselineProfile as key,value ]
         [#if baselineComponentNames?seq_contains(key)]
-            [#switch key?lower_case ]
-                [#case "opsdata" ]
-                [#case "appdata" ]
-                    [#local subComponentType = "DataBucket" ]
-                    [#break]
-
-                [#case "encryption" ]
-                [#case "sshkey" ]
-                [#case "cdnoriginkey" ]
-                    [#local subComponentType = "Key" ]
-                    [#break]
-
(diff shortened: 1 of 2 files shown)


providers/shared/components/baseline/id.ftl
```diff
@@ -73,34 +73,16 @@
...
-                [#case "cdnoriginkey" ]
-                    [#local subComponentType = "Key" ]
-                    [#break]
-
-                [#default]
-                    [@fatal
-                        message="Unknown baseline subcomponent"
-                        context=key
-                    /]
-            [/#switch]
```
roleyfoley, line 93: I think this would still be required as the values in the case statement aren't sub component types they are more around the types of data within the baseline setup

roleyfoley (changes requested):

providers/shared/components/baseline/id.ftl
ml019: I can add it back in to validate the component type expected for the target but the revised code doesn't use the key - it relies on the value being a unique subcomponent identifier in the baseline component.

roleyfoley: Yeah lets leave it out, it makes more sense that we can just pass through the id and let normal link processing fail instead

roleyfoley (approved):

PR #1794, opened 2021-09-08
refactor: warnings for occurrence links

roleyfoley:
## Intent of Change

## Description

- Adds a check to see if linked occurrences are in the same deployment and doesn't raise a warning if they
- Adds an extra check to see if the occurrence is enabled or not and raises a warning if it is

## Motivation and Context

Removes warnings that can't be fixed ( i.e. an occurrence that is being deployed at the same time as the linking occurrence ) and makes it more obvious why a link had an issue

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/occurrence.ftl
+++ b/engine/occurrence.ftl
@@ -209,6 +209,17 @@
     [#return getOccurrenceChildren(occurrence)?has_content ]
 [/#function]
 
+[#function occurrencesInSameDeployment occurrenceA occurrenceB]
+    [#return
+        getOccurrenceDeploymentUnit(occurrenceA) == getOccurrenceDeploymentUnit(occurrenceB) &&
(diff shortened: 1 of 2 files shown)


ml019:
@roleyfoley adding a warning on disabled components has implications for module design.

If something is optional, then the module can either control the `Enabled` attribute to remove it from the solution or it can use something like the `attributeIfContent()` function to completely remove the content so the target doesn't exist at all.

I think adding the warning is reasonable  - however we also now have the `ActiveRequired` flag which is probably the preferred approach if presence of the target is important, e.g. SPA/mobile app links to APIs.

I see the link warnings as something temporary while getting everything deployed. If they become permanent then their value diminishes. It does somewhat come down to personal style of the module designer. I'll approve the PR and we can iterate if the result proves problematic over time.

ml019 (approved):

Issue #1756, opened 2021-08-24
New Component - Directory 


roleyfoley:
## Expected Behaviour

I can define a new component which represents a user and machine directory authentication service like MS Active Directory. Adding this component would deploy a directory service onto a private network and include the required features to setup and manage the ongoing operations of the directory as a whole. 

User/.Group and machine management would be out of scope of this component. 

The component should be based on a managed service rather than an IaaS based configuration approach

## Current Behaviour

Not currently supported or would need to be managed through ComputeTasks on IaaS platforms 

## Possible Solution

Create and define a new component with the appropriate configuration options for a directory 

## Context

The main reason for this is to implement AWS Directory Services Managed AD as part of an AWS based solution. However Azure offer a similar service with AzureAD DS which provisions and manages domain controllers on a users behalf

PR #1796, opened 2021-09-09
feat(ds): New Component - Directory Services

ken9ross:
## Intent of Change

## Motivation and Context

https://github.com/hamlet-io/engine/issues/1756

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/directory/id.ftl
+++ b/providers/shared/components/directory/id.ftl
@@ -1,0 +1,112 @@
+[#ftl]
+
+[@addComponentDeployment
+    type=DIRECTORY_COMPONENT_TYPE
+    defaultGroup="solution"
+/]
+
+[@addComponent
+    type=DIRECTORY_COMPONENT_TYPE
+    properties=
+        [
+            {
+                "Type"  : "Description",
+                "Value" : "A managed directory services instance"
+            }
+        ]
+    attributes=
+        [
+            {
+                "Names" : "Engine",
+                "Types" : STRING_TYPE,
+                "Values" : [ "simple" ],
+                "Mandatory" : true
+            },
+            {
+                "Names" : "EnableSSO",
+                "Types" : BOOLEAN_TYPE,
+                "Default" : false
+            },
+            {
+                "Names" : "DsName",
+                "Types" : STRING_TYPE,
+                "Description": "FQDN for the directory service",
+                "Mandatory" : true
+            },
+            {
+                "Names" : "Password",
+                "Types" : STRING_TYPE
+            },
+            {
+                "Names" : "ShortName",
+                "Types" : STRING_TYPE
+            },
+            {
+                "Names" : "Size",
+                "Types" : STRING_TYPE
+            },
+            {
+                "Names" : "Links",
+                "SubObjects" : true,
+                "AttributeSet" : LINK_ATTRIBUTESET_TYPE
+            },
+            {
+                "Names" : "Profiles",
+                "Children" :
+                    [
+                        {
+                            "Names" : "Alert",
+                            "Types" : STRING_TYPE,
+                            "Default" : "default"
+                        },
+                        {
+                            "Names" : "Network",
+                            "Types" : STRING_TYPE,
+                            "Default" : "default"
+                        }
+                    ]
+            },
+            {
+                "Names" : "Hibernate",
+                "Children" : [
+                    {
+                        "Names" : "Enabled",
+                        "Types" : BOOLEAN_TYPE,
+                        "Default" : false
+                    },
+                    {
+                        "Names" : "StartUpMode",
+                        "Types" : STRING_TYPE,
+                        "Values" : ["replace"],
+                        "Default" : "replace"
+                    }
+                ]
+            },
+            {
+                "Names" : "RootCredentials",
+                "Description" : "Secret store configuration for the root credentials",
+                "Children" : [
+                    {
+                        "Names" : "Username",
+                        "Types" : STRING_TYPE,
+                        "Default" : "root"
+                    },
+                    {
+                        "Names" : "EncryptionScheme",
+                        "Types" : STRING_TYPE,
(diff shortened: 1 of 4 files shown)


providers/shared/components/directory/id.ftl
```diff
@@ -0,0 +1,112 @@
...
+                "Names" : "EnableSSO",
+                "Types" : BOOLEAN_TYPE,
+                "Default" : false
+            },
+            {
+                "Names" : "DsName",
+                "Types" : STRING_TYPE,
+                "Description": "FQDN for the directory service",
+                "Mandatory" : true
+            },
```
roleyfoley, line 35: might be worth using the Hostname/Certificate configuration for the Domain name. Then it can be configured to include distrct details

providers/shared/components/directory/id.ftl
```diff
@@ -0,0 +1,112 @@
...
+                "Names" : "Engine",
+                "Types" : STRING_TYPE,
+                "Values" : [ "simple" ],
+                "Mandatory" : true
+            },
+            {
+                "Names" : "EnableSSO",
+                "Types" : BOOLEAN_TYPE,
+                "Default" : false
+            },
```
roleyfoley, line 29: This might be better as an AWS specific parameter? This is really for SSO into the console right?

providers/shared/components/directory/id.ftl
```diff
@@ -0,0 +1,112 @@
...
+            {
+                "Names" : "DsName",
+                "Types" : STRING_TYPE,
+                "Description": "FQDN for the directory service",
+                "Mandatory" : true
+            },
+            {
+                "Names" : "Password",
+                "Types" : STRING_TYPE
+            },
```
roleyfoley, line 39: Is this different to the generated credentials?

roleyfoley (changes requested):

providers/shared/components/directory/id.ftl
ken9ross: Makes sense. Done

providers/shared/components/directory/id.ftl
ken9ross: I read this as using the AD as an SSO for other functions within the provider. I don't believe Azure allows you to set this as false, but AWS and other providers probably would so an "anything but Azure" specific parameter, which is why I went global. Thoughts?

providers/shared/components/directory/id.ftl
ken9ross: I don't think it needs to be. Removed

providers/shared/components/directory/id.ftl
roleyfoley: I think this is an AWS only thing. The SSO integration on their directory means that users created with the directory can log into the AWS Console 

With Azure AD DS  ( Their directory servcies implementation ) it works the other way around. Azure AD users can login to computers that on the Azure AD DS Domain. 

So this SSO is just for AWS

providers/shared/components/directory/id.ftl
```diff
@@ -0,0 +1,106 @@
...
+    attributes=
+        [
+            {
+                "Names" : "Engine",
+                "Types" : STRING_TYPE,
+                "Values" : [ "Simple", "ActiveDirectory" ],
+                "Mandatory" : true
+            },
+            {
+                "Names" : [ "Certificate", "Hostname" ],
```
roleyfoley, line 26: Can remove the Certificate option here and maybe just use Hostname. Also worth including a description that this is for the Domain FQDN

providers/shared/components/directory/id.ftl
```diff
@@ -0,0 +1,106 @@
...
+            },
+            {
+                "Names" : "IPAddressGroups",
+                "Types" : ARRAY_OF_STRING_TYPE,
+                "Default" : []
+            },
+            {
+                "Names" : "ShortName",
+                "Types" : STRING_TYPE
+            },
```
roleyfoley, line 37: Probably good to have a description here to. This is the Netbios name right?

providers/shared/components/directory/id.ftl
```diff
@@ -0,0 +1,106 @@
...
+            {
+                "Names" : "Size",
+                "Types" : STRING_TYPE
+            },
```
roleyfoley, line 41: Is this required? Or does it have specific value?

roleyfoley (changes requested):

providers/shared/components/directory/id.ftl
ken9ross: Correct. Added

providers/shared/components/directory/id.ftl
ken9ross: It has different values depending on whether you are looking at SimpleAD (Small or Large), or MicrosoftAD (Standard or Enterprise). Was that not what you were after?

Yes it probably should be mandatory. Done

roleyfoley: As a general user does that value matter? For both services aren't they just the physical size of the deployment? Almost like the Processor profile size? 

So could we just go small, medium, large, and then in the backend handle translating those to the appropriate service?

providers/shared/components/directory/id.ftl
ken9ross: Done

providers/shared/components/directory/id.ftl
ken9ross: For MicrosoftAD, it's also whether it's cross Region, but mostly sizing. Each only has two options - so small and large?

roleyfoley: Yeah I think lets go Small/Large and we can worry about multi region separately 
That would be an AWS specific thing and we could make that a check between the values

ken9ross: Done

roleyfoley (approved):

PR #1808, opened 2021-09-16
feat: subscription, hostingplatform and dnszone components

ml019:
## Intent of Change

## Description
Add support for new `subscription`, `hostingplatform` and `dnszone` components.

## Motivation and Context
Part of [ADR0007](https://github.com/hamlet-io/architectural-decision-log/blob/main/adr/0007-placement-support.md) implementation.

- Add shared definition of these new components
- Add preliminary changes to support `Locations`
  - accessor methods for location information on components and occurrences
  - ability to capture the errors generated by `getCompositeObject()` so alternate actions to generating fatal errors can be supported, mainly so the context of the errors can be better explained
  - More extensive commenting of the process for attribute extension and some refactoring of the implementation
- new component attributes support import of externally defined resources in readiness for the introduction of `Locations`.
- Support for an explicit Type attribute on components rather than relying on the name of a child attribute
  - makes config more explicit
  - permits a flatter JSON file
  - in line with similar changes to links with the removal of subcomponent specific attributes

Subsequent PRs will add explicit support for Locations.
## How Has This Been Tested?
Local template generation

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/component.ftl
+++ b/engine/component.ftl
@@ -196,8 +196,7 @@
                 "ResourceGroups" : {
                     resourceGroup : {
                         "Extensions" : {
-                            provider :
-                                extensions
+                            provider : extensions
                         }
                     }
                 }
@@ -206,7 +205,7 @@
 
 [/#macro]
 
-[#macro addResourceGroupInformation type attributes provider resourceGroup services=[] prefixed=true ]
+[#macro addResourceGroupInformation type attributes provider resourceGroup services=[] prefixed=true locations={} ]
     [#if provider == SHARED_ATTRIBUTES ]
         [#-- Special processing for profiles --]
         [#if resourceGroup == DEFAULT_RESOURCE_GROUP ]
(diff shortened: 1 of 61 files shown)


engine/component.ftl
```diff
@@ -241,24 +271,80 @@
...
+                "Children" : [
+                    {
+                        "Names" : "Mandatory",
+                        "Description" : "Is location always required?",
+                        "Types" : BOOLEAN_TYPE,
+                        "Default" : true
+                    },
+                    {
+                        "Names" : "TargetComponentTypes",
+                        "Description" : "Enable at rest encryption",
```
roleyfoley, line 289: Is that description right?

engine/setContext.ftl
```diff
@@ -3,6 +3,8 @@
 [#-- Shared Provider Configurations --]
 [@includeSharedComponentConfiguration component="shared" /]
 [@includeSharedComponentConfiguration component="baseline" /]
+[@includeSharedComponentConfiguration component="s3" /]
+[@includeSharedComponentConfiguration component="ec2" /]
```
roleyfoley, line 7: Do we need to include new providers? I thought we were trying to drop this?

roleyfoley (changes requested):

engine/setContext.ftl
ml019: Its temporary - they disappear in the districts PR.

engine/component.ftl
ml019: Cut/paste error - I'll fix

roleyfoley (approved):

PR #1810, opened 2021-09-17
feat(locations): location support

ml019:
## Intent of Change

## Description
Add support for Locations on ResourceGroups as a means to determine the placement of resources.

## Motivation and Context
Part of [ADR0007](https://github.com/hamlet-io/architectural-decision-log/blob/main/adr/0007-placement-support.md) implementation.

When determining placements, the resource group id is used as the Location to find the corresponding placement information,
so this means the ids of all resource groups need to be known BEFORE location information is checked.

As a result, when adding a component, any possible resource groups that an implementation might want to use must be
defined. This means that a provider implementer cannot add additional resource groups not already defined when the component was defined.

The existing placement profile approach has been left in place but if present, locations take precedence.

As locations are link based, their introduction increases the chances of creating loops. Code to detect loops has thus
been added when getting Occurrences.

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:
- engine - https://github.com/hamlet-io/engine/pull/1808

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/flows/components/flow.ftl
+++ b/providers/shared/flows/components/flow.ftl
@@ -93,11 +93,11 @@
         }
     ]
 
-    [@fatal
+    [@debug
         message="Getting link Target"
         context=
             {
-                "Occurrence" : occurrence,
+                "Occurrence" : getOccurrenceSummary(occurrence),
                 "Link" : link,
                 "EffectiveLink" : effectiveLink,
                 "ActiveOnly" : activeOnly,
@@ -239,7 +239,7 @@
     [#if targetOccurrences?size > 1]
         [@fatal
             message='Internal error - multiple component matches returned for component "${effectiveLink.Component}"'
-            context=targetOccurrences
+            context=getOccurrenceSummary(targetOccurrences)
             detail=effectiveLink
         /]
         [#return {} ]
@@ -274,7 +274,7 @@
                 [#if subOccurrences?size > 1 ]
                     [@fatal
                         message='Multiple matching subcomponents with id "${effectiveLink.SubComponent}". Use a Type attribute to differentiate them.'
-                        context=subOccurrences
+                        context=getOccurrenceSummary(subOccurrences)
                         detail=effectiveLink
                     /]
                     [#return {} ]
@@ -340,8 +340,8 @@
                     function="getLinkTarget"
                     context=
                         {
-                            "Occurrence" : occurrence,
-                            "TargetOccurrence" : matchedOccurrence,
+                            "Occurrence" : getOccurrenceSummary(occurrence),
+                            "TargetOccurrence" : getOccurrenceSummary(matchedOccurrence),
                             "Link" : link,
                             "EffectiveLink" : effectiveLink
                         }
@@ -372,7 +372,7 @@
         function="getLinkTarget"
         context=
             {
-                "Occurrence" : occurrence,
+                "Occurrence" : getOccurrenceSummary(occurrence),
                 "Link" : link,
                 "EffectiveLink" : effectiveLink
             }
@@ -418,17 +418,6 @@ that doesn't match the link.
 --]
 
 [#function internalGetOccurrences component tier={} link={} parentOccurrence={} parentContexts=[] componentType="" ]
-
-[@fatal
-    message="InternalGetOccurrences"
-    context=
-    {
-        "Component" : component,
-        "Link" : link
-    }
-    detail=internalGetOccurrencesInvocations
-    enabled=false
-/]
 
     [#if !(component?has_content) ]
         [#return [] ]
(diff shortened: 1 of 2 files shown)


providers/shared/flows/components/flow.ftl
```diff
@@ -84,22 +84,29 @@
...
+    [#local effectiveLink =
+        link +
+        {
+            "Instance" : internalConvertDefaultIdName(link.Instance!(getOccurrenceInstance(occurrence).Id)),
+            "Version" : internalConvertDefaultIdName(link.Version!(getOccurrenceVersion(occurrence).Id))
+        }
+    ]
 
-    [@debug
+    [@fatal
```
roleyfoley, line 96: Looks like a hangover from debugging

providers/shared/flows/components/flow.ftl
```diff
@@ -174,169 +181,186 @@
...
+    [#-- A match to a component occurrence will also return --]
+    [#-- any SubOccurrences. A match to a SubOccurrence     --]
+    [#-- only returns the Suboccurrence itself              --]
+    [#local targetOccurrences = internalGetOccurrences(component, getTier(link.Tier), effectiveLink) ]
 
-        [#local potentialOccurrences = [targetOccurrence] ]
+    [#if targetOccurrences?size > 1]
+        [@fatal
+            message='Internal error - multiple component matches returned for component "${effectiveLink.Component}"'
+            context=targetOccurrences
```
roleyfoley, line 237: Can we cut this down? Occurrences are huge! not sure if they will help someone figure out an issue

providers/shared/flows/components/flow.ftl
```diff
@@ -174,169 +181,186 @@
...
-
-                    [#local potentialOccurrences += [potentialOccurrence] ]
+            [#if effectiveLink.SubComponent?has_content]
+
+                [#-- Match is to a SubComponent --]
+
+                [#if subOccurrences?size > 1 ]
+                    [@fatal
+                        message='Multiple matching subcomponents with id "${effectiveLink.SubComponent}". Use a Type attribute to differentiate them.'
+                        context=subOccurrences
```
roleyfoley, line 272: Same here, maybe some specific details rather than the occurrence details

providers/shared/flows/components/flow.ftl
```diff
@@ -570,27 +693,208 @@
                         [/#list]
                     [/#list]
 
+                    [#-- Determine any provided location information permitting profiles to contribute content --]
+                    [#local locations =
+                        getCompositeObject(
+                            [
+                                {
+                                    "Names" : ["deployment:Locations"],
```
roleyfoley, line 701: I think there is actually a configuration section where the deployment  global variable exists. Would it be possible to put this in there to keep the deployment configuration in one spot?

roleyfoley: It also aligns with the model that some component/subcomponents can't be split and some can. The deployment details should only show up on components that have control over their deployment.

providers/shared/flows/components/flow.ftl
```diff
@@ -570,27 +693,208 @@
...
+                                                    "Placement" : placement
+                                                }
+                                            }
+                                        }
+                                    }
+                                ) ]
+                        [#else]
+                            [@fatal
+                                message='Insufficient information to place resource group "${key}". Provider and DeploymentFramework are required'
+                                context=occurrence
```
roleyfoley, line 790: Another big error

roleyfoley (changes requested):
Looks like this might need a rebase? or is it just that things have moved around?

providers/shared/flows/components/flow.ftl
ml019: yep - fixed.

providers/shared/flows/components/flow.ftl
ml019: I've added a helper routine in occurrence.ftl to abstract exactly what needs to be shown - for now its the type and the core ids to identify the occurrence. We can then tweak the function in future - maybe link it to the logging mode (compact/full) so if you really want the detail, you can get it.

providers/shared/flows/components/flow.ftl
ml019: see above

providers/shared/flows/components/flow.ftl
ml019: see above

providers/shared/flows/components/flow.ftl
ml019: There is a special "deployment" provider in the default RG where they all are. Given where this lookup is in the code, I didn't want to be looking up all the deployment info. 

Even if you can't deploy an occurrence, it can still have locations so perhaps I need a different prefix. Open to suggestions ...

roleyfoley: I think it might be better to move it to a different prefix. The deployment options shouldn't be included by default. I think some of them are at the moment

ml019: OK - I'll go with `placement:` as that is the predominant use of locations. We could also use them for things like code repos - not sure if that sort of thing is better done as links on the component or as an addition location. Advantage of locations is they do check that the endpoint is the correct type whereas in the solution it is just whether the link is formatted correctly.

ml019:
@roleyfoley retested with `placement:` with no issues. Ready for review thanks.

roleyfoley (approved):

PR #1813, opened 2021-09-21
fix(user): remove scheme restriction

roleyfoley:
## Intent of Change

## Description

Removes the required values for the encryption scheme so that it can be controlled by users as required

## Motivation and Context

The current value only allowed one value which wouldn't work and removed the ability to configure it

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/user/id.ftl
+++ b/providers/shared/components/user/id.ftl
@@ -32,18 +32,20 @@
                 "Children" : [
                     {
                         "Names" : "Formats",
+                        "Description" : "The type of credentials to generate - system provides api level access, console is user level access",
                         "Types" : ARRAY_OF_STRING_TYPE,
                         "Values" : ["system", "console"],
                         "Default"  : [ "system" ]
                     }
                     {
                         "Names" : "EncryptionScheme",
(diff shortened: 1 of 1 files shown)


providers/shared/components/user/id.ftl
```diff
@@ -32,18 +32,20 @@
...
+                        "Description" : "The type of credentials to generate - system provides api level access, console is user level access",
                         "Types" : ARRAY_OF_STRING_TYPE,
                         "Values" : ["system", "console"],
                         "Default"  : [ "system" ]
                     }
                     {
                         "Names" : "EncryptionScheme",
+                        "Description" : "A prefix added to the start of the encrypted value to show that it is encrypted",
                         "Types" : STRING_TYPE,
-                        "Values" : ["base64"],
```
ml019, line 42: This was based on the support elsewhere for encrypting values. I guess we can still do that but this allows users to roll their own if they wish?

ml019 (approved):

PR #1815, opened 2021-09-26
feat(base): null detection in base routines

ml019:
## Intent of Change

## Description
Add support for null detection in any of the arguments passed to the library routines provided in `base.ftl`.

A few key routines have also been moved into `base.ftl` from `engine.ftl` and `engine.ftl` deleted. It was a leftover from a previous experiment related to dynamic loading which has since been superceded.

## Motivation and Context
The detection relies on the behaviour of Freemarker whereby if a null is passed to a routine, and the routine has a default for an argument, the default will be used in preference to the null value.

For routines with arguments that do not have a natural default, the default is set to the string `Hamlet:Null`, the presence of which then permits detection of a null being passed without triggering Freemarker's normal error detection.

Deep detection is performed, so things like a null value in an array will be detected when the array is first presented rather than in a lower level routine where it is harder to understand the context. For this reason, checking has been added to all routines even though this will mean some arguments are checked at multiple levels. It also means the reporting of the caller who triggered the error is more targeted.

Deep detection is also useful for checking arguments which do have a natural default value, particularly arrays and objects.

Because of the heavy use of the base routines throughout hamlet, null detection is disabled by default. A single function controls whether it is performed so as soon as a null exception is reported, it is a simple matter to rerun the processing with null detection turned on. For now the assumption is that the code will be to hand so this change can be done temporarily but a future change could be to add a cli switch for enabling detection.

The pattern established in this commit can also be used elsewhere in general purpose routines such as id or name formatting, again so the reporting of the source of the issue is more accurate.

## How Has This Been Tested?
Local hamlet cli command execution

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


roleyfoley (approved):
I wish I could approve this twice!

PR #1821, opened 2021-09-29
refactor: setContext wrapper functions (1)

ml019:
## Intent of Change

## Description
Replace the use of global variables in setContext() with accessor functions for the same information.

 NOTE: dependent PRs need to be merged at the same time as this PR to avoid failures.

## Motivation and Context
This is in preparation for full district support, where all config and state access needs to be via the input state.

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:

- None

### Dependent PRs:
- azure - https://github.com/hamlet-io/engine-plugin-azure/pull/278
- aws - https://github.com/hamlet-io/engine-plugin-aws/pull/421

### Consumer Actions:

- None

### DIFF
--- a/engine/common.ftl
+++ b/engine/common.ftl
@@ -291,7 +291,7 @@
     [#elseif processorProfile.MaxPerZone?has_content ]
         [#local maxCount = processorProfile.MaxPerZone ]
         [#if multiAZ]
-            [#local maxCount = maxCount * zones?size]
+            [#local maxCount = maxCount * getZones()?size]
         [/#if]
     [#else]
         [@fatal
@@ -305,7 +305,7 @@
     [#elseif processorProfile.MinPerZone?has_content ]
         [#local minCount = (processorProfile.MinPerZone)!1]
         [#if multiAZ]
-            [#local minCount = minCount * zones?size]
+            [#local minCount = minCount * getZones()?size]
         [/#if]
     [#else]
         [@fatal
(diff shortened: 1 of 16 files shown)


roleyfoley (commented):
A lot of fragments use these values to configure complex settings like SES regions or naming lookups 

We could just call this a breaking change and let fragments fail or we could include them in the invokeExtensions macro as assigned vars?

ml019:
It is a relatively straightforward process to cut them across to the accessor methods. I'd suggest if we are doing that (9.0.0) then we also review all the TODOs which talk about removing legacy code and have a bit of a clean up.

As a mitigation (and partially as a future looking change) we can ensure all pipelines are explicitly setting the hamlet version so if necessary, we can lock to 8.3.0.

The alternative is to deprecate use of absolute variables and do the temporary setting as you suggest. 

Given I don't feel strongly either way, I think backwards compatibility should be the overridding concern. 

I'll add a function in `getContext()` which does the assign thing and call that from invokeExtensions.

ml019:
Given this direction, I'll make the function names more like accessors (`get...`) rather than look like variables and then the actual variables will all be concentrated in that one macro.

ml019:
@roleyfoley refactored to keep global vars for now. Forcing the accessors to be of the form `get...()` makes the changes easier to review I think (given that sometimes local vars of the same name as the previous globals have been defined) and I've allowed for future control as to whether we set the globals by default.

Ready for another review if you would be so kind.

roleyfoley (approved):

PR #1837, opened 2021-10-17
feat: add named ip address groups

roleyfoley:
## Intent of Change

## Description

- Adds support for referencing inbuilt named prefix groups in IPAddressGroups
- Defaults group to closed when handling missing tier configuration - mostly a security best practice thing

## Motivation and Context

In providers like Azure there are a collection of Prefix references that map to Azure Services or Generic IP Collections ( Internet, VirtualNetwork etc. ). When using them in Azure they can be used in the same places that CIDR ranges can be used. 

This aligns the concept of the Named prefix groups with our IPAddress Groups 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/setContext.ftl
+++ b/engine/setContext.ftl
@@ -616,6 +616,28 @@
 [#function getIPAddressGroup group occurrence={}]
     [#local groupId = group?is_hash?then(group.Id, group) ]
 
+    [#if groupId?starts_with("_named") || groupId?starts_with("__named") ]
+        [#local lookupName = groupId?split(":")[1]]
+        [#if ! lookupName?has_content]
+            [@fatal
+                message="Invalid named IP address group"
+                detail="Provide named groups as _named: name of group"
+                context=groupId
+            /]
+
+            [#return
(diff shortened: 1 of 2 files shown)


engine/setContext.ftl
```diff
@@ -629,8 +651,8 @@
             [#return
                 {
                     "Id" : groupId,
-                    "IsOpen" : true,
-                    "IsLocal" : false,
+                    "IsOpen" : false,
```
ml019, line 654: This looks like a bug fix? Issue only said enhancement. Assuming we aren't currently using `_tier` anywhere otherwise it would have not worked correctly?

roleyfoley: Whoops will add a note for that. Just more that if you specify a Tier it shouldn't open if it can't be found and so we shouldn't open up if it can't be found

ml019 (approved):

PR #1851, opened 2021-11-06
fix: handle missing deployment unit

roleyfoley:
## Intent of Change

## Description

- Handles scenarios where the deployment unit value is null or empty with the move to only include the attribute when required

## Motivation and Context

Exceptions being thrown when deploymentUnit properties were empty when they were required

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/common.ftl
+++ b/engine/common.ftl
@@ -5,7 +5,7 @@
     [#if obj["deployment:Unit"]?has_content ]
         [#return obj["deployment:Unit"]]
     [#else]
-        [#if ! obj.DeploymentUnits?? || ! obj.DeploymentUnits?has_content ]
+        [#if ! obj.DeploymentUnits?has_content ]
             [#return ""]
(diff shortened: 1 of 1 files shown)


engine/common.ftl
```diff
@@ -5,6 +5,10 @@
     [#if obj["deployment:Unit"]?has_content ]
         [#return obj["deployment:Unit"]]
     [#else]
+        [#if ! obj.DeploymentUnits?? || ! obj.DeploymentUnits?has_content ]
```
ml019, line 8:
```suggestion
        [#if ! obj.DeploymentUnits?has_content ]
```
`has_content` handles missing attributes

ml019 (changes requested):

ml019 (approved):

PR #1874, opened 2021-12-08
feat: standard configuration storage

roleyfoley:
## Intent of Change

## Description

- Adds a standardised configuration store to provide a definitive store of the configuration available in hamlet
- Updates calls to get configuration to use helper functions instead of direct calls to global variables
- Removes the plural type from the attribute set configuration as it is not required

## Motivation and Context

The key motivation for this is to create a central standardised store of the configuration available to hamlet. This can be used to enforce and document a fixed schema throughout hamlet. On top of this it should be possible to also determine the source of configurations that have been added through plugins, which is where most of our configuration comes from. 

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019:
Need to understand how this interacts with dynamic CMDB loading and cross-solution linking. A lot of this configuration is not a singleton but varies as the context changes with following the link.

roleyfoley:
The next step after this is to add the source plugin for each configuration set. This would allow for each plugin to add their configuration sets and then when the configuration is looked up filters can be applied to control what the effective result would be.

providers/shared/views/schemaset/setup.ftl
```diff
@@ -36,13 +36,13 @@
...
-                [#local schemas = referenceConfiguration?keys ]
+                [#local schemas = getReferenceConfiguration()?keys ]
                 [#break]
             [#case "attributeset" ]
-                [#local schemas = attributeSetConfiguration?keys ]
+                [#local schemas = getAttributeSetIds() ]
                 [#break]
             [#case "module" ]
-                [#local schemas = moduleConfiguration?keys ]
+                [#local schemas = getModuleConfiguration?keys ]
```
ml019, line 45:
```suggestion
                [#local schemas = getModuleConfiguration()?keys ]
```

ml019 (changes requested):
I think I get the idea. No mention of config being provider specific but I guess we can evolve if that seems necessary.

roleyfoley:
Yeah that is the next PR once this is merged. That is potentially the split between provider and plugin...

ml019 (approved):

PR #1878, opened 2021-12-12
feat: add support for component level config settings

roleyfoley:
## Intent of Change

## Description

- Adds support for defining settings at the component level
- Adds a macro for controlling the inclusion of component variables
- Adds a default context property set to make sure these are available

## Motivation and Context

Settings are available across all components and are simple key value config options

The Settings configuration options are on all Components as 

```json
{
   "Settings" : {
         "my_setting_key" : {
               "Value" : "my_setting_property"
          }
}
```

These will be treated as we do with all settings and will be treated as env vars

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- https://github.com/hamlet-io/engine-plugin-aws/pull/484

### Dependent PRs:

- None

### Consumer Actions:

- None


engine/occurrence.ftl
```diff
@@ -900,6 +902,16 @@ already prefixed attribute.
     ]
 [/#function]
 
+[#function internalCreateOccurrenceComponentSettings occurrence]
+    [#local result = {}]
+    [#list (occurrence.Configuration.Solution.Settings)!{} as id, setting]
+        [#if (setting.Enabled)!true]
```
ml019, line 908: This assumes the value provided is an object - if a string is provided, I suspect freemarker will throw an error.

Up until now, settings has been an internal structure which is always an object. If the value in the settings file is a string, it gets converted to an object with a `Value` attribute. This allows for the entry to have other attributes like "Internal" which ensures the setting is ignored when generating an environment from it.

The config definition has the value as an ANY_TYPE so I think some more complex logic is needed here (in line with the existing processing of settings files) to check the type of "setting", build an object per existing code if it isn't one (or it doesn't have a `Value` attribute itself) or just use the object as is if it does have a `Value` attribute

ml019 (changes requested):
Please also add a ticket to deprecate the "Properties" attribute on the externalservice component in favour of the `Settings` attribute added by this PR. It should be marked as "required breaking change" so we can eventually remove the code supporting the `Properties` attribute.

engine/occurrence.ftl
roleyfoley: The Settings here is a component configuration that represents the wrapper for setting them in the solution. 

So if you have 

```
   { 
      "Settings" : {
           "my-setting" : {
                "Value" : {
                    "this is a setting" : "as an object" 
                }
           }
      }
   } 
``` 

Then the setting that is created will be 

```
{
    "my-setting" : { "this is a setting" : "as an object" 
}
```

So that shouldn't mess with any of the settings handling we do once we look at the setting. The Value attribute defined in the component won't be seen by the underlying setting processing.

roleyfoley: Standard string settings would be defined as 

```
{
    "Settings" : {
        "my-string" : { 
                "Value" : "my-value" 
         } 
} 
```

and would be added to settings as 

```
{ "my-string" : "my-value"
```

roleyfoley: The settings are each subobjects to align with the solution configuration and to do that you can't have a key : value pair like we have in settings files at the moment

ml019: `I missed that they were Subobjects. The trouble is that the `Internal` attribute is at the same level as `Value` so we will need to add that to the config schema and handle it when merging the objects.

Each setting is managed as an object and only converted to name: value whenb generating an environment.

roleyfoley: Thats why the Value attribute in the component configuration has an Any type. Its not the Value attribute in the setting. So to use the value syntax you would need to put 


```
{
    "Settings" : {
        "my-string" : { 
                "Value" : {
                     "Value" : { "JSON" : "Stuff" } 
                 }
         } 
} 
```

or 

```
{
    "Settings" : {
        "my-string" : { 
                "Value" : {
                     "Internal" : true,
                     "Value" : { "JSON" : "Stuff" } 
                 }
         } 
} 
```

Are you saying you don't want it that way or that I should incorporate more of the settings schema into this?

roleyfoley: Sorry read this completely wrong, I'm setting all values provided as Value to that root level Value key that we allow for in the settings files. So to align with this I should also include the other root level settings we support for settings values like Senstive or Internal. This will allow for them to be processed without the need for the double Value key.

roleyfoley: Fixed this up and pused in the PR

engine/configuration/component.ftl
```diff
@@ -196,6 +196,108 @@
...
+                },
+                {
+                    "Names" : "Sensitive",
+                    "Description" : "Hide the value of this setting when using it in output",
+                    "Types" : BOOLEAN_TYPE,
+                    "Default" : false
+                },
+                {
+                    "Names" : "Internal",
+                    "Description" : "Don't inlude this setting as part of environment variable generation",
```
ml019, line 290:
```suggestion
                    "Description" : "Don't include this setting as part of environment variable generation",
```

ml019 (approved):

Issue #1917, opened 2022-01-23
Backup Support


ml019:
## Expected Behaviour
It should be possible to use backup capabilities provided by providers to support the backup approaches required by products.

## Current Behaviour
There is currently no support for configuring backup facilities within providers.

## Possible Solution
- add new component `backupstore` with `backupstoreregime` subcomponent
- store offers configuration of encryption keys and notifications whenever a backup to a store occurs
- `backupstoreregime` provides details of a set of rules to apply and what should be targetted by the regime
- a regime can have links to specific resources to back up
- a regime can define a tag that if placed on a resource, will cause it to be included in the regime
- a regime can define a set of tags on the component that must match the same tag on the backupstore
- if a component is linked to a regime (and supports centralised backups), it will have any tags defined by the regime added in order that it will be affected by the regime. 

For AWS, 
- use a prologue script to detect if the required services are enabled for backup, and if not, enable them.

## Context
- support backup requirements of customer projects
- simplify the process of maintaining backups regardless of the project solution

roleyfoley:
Do we need the split between the two components? If you want to have two stores capturing backups from different locations you could 

- Use a deployment profile 
- If the store supports it have a replication rule or option on the store 

Most backup services follow a pull/discovery based process so saying that a store has these rules to pull data into itself, might align with backup processes and make it simpler to configure.

ml019:
Closing. Basic implementation in place. Any additions can be handled as enhancements if required.

PR #1921, opened 2022-01-25
feat(backup): Backup support

ml019:
## Intent of Change

## Description
Add a component representing a store for backups sourced from one or more components.

Regimes are used to determine combinations of schedule and component for which snapshots should be generated and saved in the store.

Components to be backed up can be selected via
- carrying tags that match those of the backup store
- carrying a tag specific to a regime on a backup store
- direct links from the backup store.

Copies of snapshots can be saved in additional backup stores if required.

## Motivation and Context
Part of #1917 

## How Has This Been Tested?
- Local run of the test suite
- Additional verification of completeness as part of implementation for the AWS provider
 
### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


providers/shared/components/backupstore/id.ftl
```diff
@@ -0,0 +1,217 @@
...
+            {
+                "Type"  : "Description",
+                "Value" : "Store for holding backups"
+            }
+        ]
+    attributes=
+        [
+            {
+                "Names" : "Links",
+                "Description" : "Link to the key to be used for encrpytion of snapshots. If not provided, the baseline key will be used",
```
roleyfoley, line 16: I think if you want to have a specific function for a link it might be better as a dedicated attribute?

providers/shared/components/backupstore/id.ftl
```diff
@@ -0,0 +1,217 @@
...
+                        "Default" : "_operations"
+                    },
+                    {
+                        "Names" : "Offline",
+                        "Types" : [NUMBER_TYPE],
+                        "Description" : "Days until backup is moved to longer term storage (if supported)",
+                        "Default" : "_operations"
+                    }
+                ]
+            },
```
roleyfoley, line 154: Types would be STRING_TYPE and NUMBER_TYPE based on the default yeah? 

Also kinda confusing looking at it as the defaults are the same value but it says that the Expiration has to be larger than the Offline

providers/shared/components/backupstore/id.ftl
```diff
@@ -0,0 +1,217 @@
...
+            {
+                "Names" : "Targets",
```
roleyfoley, line 156: Can I use all three at once? If not then might need a TargetType or something like that where you specify for the rule which one is used.

roleyfoley (changes requested):

providers/shared/components/backupstore/id.ftl
ml019: When I checked our key support, it is currently only on the baseline component so I switched to just being able to say which key is required.

providers/shared/components/backupstore/id.ftl
ml019: I've tweaked the defaults and descriptions. With qualification and modules, I've not tried to replicate the `_` style values we have for S3 - I think they would be better handled by deployment profiles.

providers/shared/components/backupstore/id.ftl
ml019: MatchesStore just controls conditions so doesn't actually select any targets. 

By default I add a match all if no tag support or specific resources are provided.

roleyfoley (approved):

PR #1934, opened 2022-02-11
feat: add docdb support

ken9ross:
## Intent of Change

## Description
Add support for DocDB

## Motivation and Context
Add support for DocDB

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


providers/shared/components/ddb/id.ftl
```diff
@@ -0,0 +1,311 @@
...
+        [
+            {
+                "Type"  : "Description",
+                "Value" : "A managed NoSQL database instance"
+            },
+            {
+                "Type" : "Note",
+                "Value" : "Minor Version Upgrades - Are presently ignored by DocDB",
+                "Severity" : "warning"
+            }
```
roleyfoley, line 15: If we aren't using minor version support it should be removed

providers/shared/components/ddb/id.ftl
```diff
@@ -0,0 +1,311 @@
...
+                ]
+            },
+            {
+                "Names" : "MaintenanceWindow",
+                "AttributeSet" : MAINTENANCEWINDOW_ATTRIBUTESET_TYPE
+            },
+            {
+                "Names" : "BackupWindow",
+                "AttributeSet" : BACKUPWINDOW_ATTRIBUTESET_TYPE
+            },
```
roleyfoley, line 167: Should this be under the existing Backup configuration section?

providers/shared/components/ddb/id.ftl
```diff
@@ -0,0 +1,311 @@
...
+            {
+                "Names" : "AutoMinorVersionUpgrade",
+                "Types" : BOOLEAN_TYPE
+            },
```
roleyfoley, line 171: Based on what is stated above this should be removed

providers/shared/components/ddb/id.ftl
```diff
@@ -0,0 +1,311 @@
...
+                            },
+                            {
+                                "Names" : "Value",
+                                "Types" : [ STRING_TYPE, NUMBER_TYPE, BOOLEAN_TYPE],
+                                "Mandatory" : true
+                            }
+                        ]
+                    }
+                ]
+            }
```
roleyfoley, line 304: IS clustering supported on current providers?

roleyfoley (changes requested):

providers/shared/components/ddb/id.ftl
roleyfoley: Yep looks like it is in AWS.

providers/shared/components/ddb/id.ftl
ken9ross: Done

providers/shared/components/ddb/id.ftl
ken9ross: Done

providers/shared/components/ddb/id.ftl
```diff
@@ -0,0 +1,311 @@
...
+                            {
+                                "Names" : "CollectionInterval",
+                                "Description" : "Metric Collection interval ( seconds )",
+                                "Types" : NUMBER_TYPE,
+                                "Values" : [ 0, 1, 5, 10, 15, 30, 60 ],
+                                "Default" : 60
+                            }
+                        ]
+                    }
+                ]
```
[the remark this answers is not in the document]
ken9ross: Removed

providers/shared/components/ddb/id.ftl
ken9ross: I saw it as more closely aligned to the MaintenanceWindow, but not strongly. Done

roleyfoley (changes requested):

providers/shared/components/component.ftl
```diff
@@ -54,6 +54,8 @@
     type=DB_COMPONENT_TYPE
 /]
 
+[#assign DDB_COMPONENT_TYPE = "ddb" ]
```
roleyfoley, line 57: ddb doesn't really mean anything as a Component name. Why not docdb?

providers/shared/references/Processor/id.ftl
```diff
@@ -100,6 +100,16 @@
...
+        {
+            "Names" : [ "ddb", "DDB" ],
+            "Children" : [
+                {
+                    "Names" : "Processor",
+                    "Types" : STRING_TYPE
+                }
+            ] +
+            nodeCountChildConfiguration
+        },
```
roleyfoley, line 112: Don't need to do casing here as we already do a to lower on it, and we don't really need two ways to set it. 

This should also be updated to align with the component name if that is changed.

ken9ross: I just replicated the pattern from other components in case it mattered. Done

providers/shared/components/component.ftl
ken9ross: Done

providers/shared/components/ddb/id.ftl
```diff
@@ -0,0 +1,260 @@
...
+            },
+            {
+                "Names" : "Backup",
+                "Children" : [
+                    {
+                        "Names" : "BackupWindow",
+                        "AttributeSet" : BACKUPWINDOW_ATTRIBUTESET_TYPE
+                    },
+                    {
```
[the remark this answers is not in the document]
ken9ross: True but it was very similar to MaintenanceWindow which is used in multiple locations so though to make it an attribute set on the assumption that it would also appear elsewhere (although given the elapsed time - maybe not)

Is there a downside to it being an attribute set?

roleyfoley (approved):

PR #1936, opened 2022-02-13
refactor: full names based on district

ml019:
## Intent of Change

## Description
Expand the district configuration to control the format of full names for resources.

Remove the corresponding configuration in setContext.

Add an "environment" district that does not have segment for simpler deployments.

## Motivation and Context
By driving the full names from the district, full names can reflect whatever district layer configuration is selected.

## How Has This Been Tested?
- Local template generation
- Run engine and aws test suites

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/inputdata/district.ftl
+++ b/engine/inputdata/district.ftl
@@ -90,14 +90,14 @@
 [#-- validated and normalised district     --]
 [#function getLinkDistrict link logErrors=true]
     [#-- Validate the district configuration --]
-    [#return internalGetLinkDistrict(link, logErrors) ]
+    [#return internalGetLinkDistrict(link, .caller_template_name, logErrors) ]
 [/#function]
 
 [#macro pushDistrict link district={} ]
 
     [#local stackEntry = district ]
     [#if ! district?has_content]
-        [#local stackEntry = internalGetLinkDistrict(link) ]
+        [#local stackEntry = internalGetLinkDistrict(link, .caller_template_name) ]
     [/#if]
 
     [#-- Check if the district has changed --]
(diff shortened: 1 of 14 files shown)


roleyfoley (approved):

roleyfoley:
@ml019 Looks like some attributes are missing or the validation isn't quite working as expected here.

ml019:
@roleyfoley ready for another review Thanks.

engine/inputdata/district.ftl
```diff
@@ -17,16 +17,55 @@
...
+                                    "Type" : STRING_TYPE
+                                },
+                                {
+                                    "Names" : "Ignore",
+                                    "Description" : "An array of values to not include if encountered",
+                                    "Type" : ARRAY_OF_STRING_TYPE
+                                }
+                            ]
+                        }
+                    ],
```
roleyfoley, line 61: Is it possible to make this an Attribute set and then use the attribute set lookup process to get the configuration? That way it can be documented easily with our current schemas?

roleyfoley (changes requested):

engine/inputdata/district.ftl
ml019: Sure - I didn't go that path because I didn't see it being reused but happy to do it to get doco benefits.

roleyfoley: Can make it a configuration set to if you want but reference data felt the quickest approach to the configuration available for doco.

roleyfoley (approved):

PR #1937, opened 2022-02-21
fix(apigateway): handling of null values in definitions

roleyfoley:
## Intent of Change

## Description

- Add a warning but continue to process null values found in definition files

## Motivation and Context

In OpenAPI there is a property called nullable within the specification, this is being caught by the current processing and the API definitions are not loaded even though it is valid. 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/inputseeders/shared/id.ftl
+++ b/providers/shared/inputseeders/shared/id.ftl
@@ -376,8 +376,15 @@
         ]
     [/#if]
 
+    [#if definitions?contains("null")]
+        [@warn
+            message="null JSON values not supportd in hamlet"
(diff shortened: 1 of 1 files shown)


ml019 (approved):

providers/shared/inputseeders/shared/id.ftl
```diff
@@ -376,8 +376,15 @@
         ]
     [/#if]
 
+    [#if definitions?contains("null")]
```
ml019, line 379:
```suggestion
    [#if (definitions!"")?contains("null")]
```

ml019 (changes requested):

PR #1947, opened 2022-03-01
feat: null clean inputstage

roleyfoley:
## Intent of Change

- Documentation (new or update to existing)

## Description

- Adds an input stage that is applied after loading user inputs to strip any null values which aren't supported in freemarker well
- Adds a warning when they are found
- Adds an entrance to show how each input sources are put together

## Motivation and Context

null values aren't handled well in freemarker and its not possible to output them when writing output as freemarker doesn't have a null character. 

This input stage checks for any values that are missing or null and removes either their key:value from dictionaries or their value from a sequence

The inputinfo entrance tries to provide more details on how input stages are working.

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


providers/shared/entrances/inputinfo/id.ftl
```diff
@@ -0,0 +1,18 @@
+[#ftl]
+
+[@addEntrance
+    type=INPUTINFO_ENTRANCE_TYPE
+    properties=
+        [
+            {
+                "Type"  : "Description",
+                "Value" : "Provides information for release management tasks performed by executors"
```
ml019, line 9:
```suggestion
                "Value" : "Provides information about the pipelines used for processing inputs"
```

ml019 (changes requested):

ml019 (approved):

PR #1949, opened 2022-03-04
fix: use raw path for settings path prefix

roleyfoley:
## Intent of Change

## Description

- fixes an issue where settings s3 keys were being overridden when syncing as file settings
- moves the standard policies for IAM from the shared provider to the aws provider

## Motivation and Context

When working with components that have common first sections of their id ( mgmt, mgmt-worker ) the FullRelativePath  value overlaps between the subcomponents. 
When syncing the asFile settings to s3 this can cause the settings to be overriden as the sync command deletes files that don't match the sync. 

Moving to the FullRawRelativePath ensures the paths are unique for each component or subcomponent. 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/common.ftl
+++ b/engine/common.ftl
@@ -85,7 +85,7 @@
 [#-- S3 settings/appdata storage  --]
 
 [#function getSettingsFilePrefix occurrence ]
-    [#return formatRelativePath("settings", occurrence.Core.FullRelativePath) ]
+    [#return formatRelativePath("settings", occurrence.Core.FullRelativeRawPath) ]
 [/#function]
 
 [#function getAppDataFilePrefix occurrence={} ]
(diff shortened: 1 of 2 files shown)


ml019 (approved):
This was intentional but predates settings namespace support.

I think making the settings unique by default makes sense now - if you want them shared then namespaces are available.

roleyfoley:
> This was intentional but predates settings namespace support.
> 
> I think making the settings unique by default makes sense now - if you want them shared then namespaces are available.

The problem was that while the namepsaces overlapped each component still had a unique value. 

So in the case this was found the SETTINGS_PREFIX for the components ended in 

mgmt/
mgmt/worker
mgmt/beat 


The sync function for S3 objects has sync --delete which meant at when the script run it would sync the worker and beat into their own directories, then when the mgmt/ ran the sync it would remove the worker and beat prefixes as the sync command was aligning the local directory with the s3 prefix. 

This meant that the worker and beat were told to look in their prefix but it was empty.

ml019:
Yep - sure. Raw makes sense.

PR #1968, opened 2022-04-10
feat: district based solutions

roleyfoley:
## Intent of Change

## Description

- Adds support for defining solutions which are applied to a matching district
- Adds support for defining and extending the configuration that makes up the solution

## Motivation and Context

Makes it possible for other districts like the account district to have its own solution/components. Extensions of the solution allow for different ways to define how the district might work, the Tiers model is the primary one for defining components but there might be other configuration approaches that would align with being used in solutions. 

Solutions are merged into each other if multiple matches for a given district are found for an available solution, this allow for solutions to be broken up into more relevant chunks but still need to be deployed into the same segment/environment. 

Backwards compatibility is in place with the current root level Tiers object mapping to a segment level solution applied to all layers as a filter. This ensures that the transition to a object based solution is easy. 

## How Has This Been Tested?

Tested locally on existing test suite. 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/bootstrap.ftl
+++ b/engine/bootstrap.ftl
@@ -33,6 +33,7 @@
 [#include "inputdata/reference.ftl" ]
 [#include "inputdata/setting.ftl" ]
 [#include "inputdata/seed.ftl" ]
+[#include "inputdata/solution.ftl" ]
 
 [#-- Plugin data loading --]
 [#include "extension.ftl" ]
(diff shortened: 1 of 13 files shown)


ml019:
@roleyfoley would you mind resolving the conflict? I'll merge this then

ml019 (approved):

PR #1969, opened 2022-04-11
refactor: move multiAZ to standard component configuraiton

roleyfoley:
## Intent of Change

## Description

- Adds a MultiAZ attribute to all components which currently use MultiAZ in both the aws and azure providers
- Removes the MultiAZ flag on layers that had it configured
- Adds a deployment profile for multiAZ environment control to replicate the current behaviour
- Remove layer support for minor version updates
- Adds deployment profile to control minor version updates per env
- Removes NAT controls at the segment level as they are no longer used

## Motivation and Context

This simplifies where component configuration is sourced from and controlled and allows for removing the Solution layer which isn't really a layer following review of the expected implementation of solutions

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- If Deployment Profiles have been defined on the environment object they will override the default ones set for these properties. They will need to be added to the array.

### DIFF
--- a/engine/base.ftl
+++ b/engine/base.ftl
@@ -1718,22 +1718,13 @@ are added.
 
 [#macro validateCompositeObject attributes=[] objects=[] ]
 
-    [#-- Common Parameters that are used throughout        --]
-    [#-- but are as-yet unaccounted for in the             --]
-    [#-- composite object definitions.                     --]
-    [#-- TODO(rossmurr4y): add commonParams to definitions --]
-    [#local commonParams = [
-        "multiAZ", "MultiAZ"]
-    ]
-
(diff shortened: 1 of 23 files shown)


ml019:
Overall this looks good but it wasn't clear where the segment NAT config went? Are we assuming we will always use AWS NATs?

roleyfoley:
> Overall this looks good but it wasn't clear where the segment NAT config went? Are we assuming we will always use AWS NATs?

I think most of this was moved into the Component itself and it was updated to follow the multiAZ config of the component

ml019:
This will also require an update to any customer site with their own environments like HA

ml019 (approved):
Nice to cleanup some of these oddities.

PR #1980, opened 2022-05-03
feat(lambda): provisioned executions

ml019:
## Intent of Change

## Description
Add the config to permit provisioned executions of versioned lambda functions.

## Motivation and Context
Tuning of customer applications for production loads usually requires establishing provisioned capacity.

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/lambda/id.ftl
+++ b/providers/shared/components/lambda/id.ftl
@@ -224,6 +224,11 @@
                 "Default" : -1
             },
             {
+                "Names" : "ProvisionedExecutions",
+                "Types" : NUMBER_TYPE,
+                "Default" : -1
+            },
+            {
                 "Names" : "Image",
                 "Description" : "Control the source of the image that is used for the function",
                 "Children" : [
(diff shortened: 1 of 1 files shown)


providers/shared/components/lambda/id.ftl
```diff
@@ -223,6 +223,11 @@
                 "Types" : NUMBER_TYPE,
                 "Default" : -1
             },
+            {
+                "Names" : "ProvisionedExecutions",
```
roleyfoley, line 227: Should have a description to make it clear in the documentation what this does.

roleyfoley (changes requested):

providers/shared/components/lambda/id.ftl
ml019: Added

roleyfoley (approved):

Issue #1640, opened 2021-04-14
Lambda runtime validation relaxation


ml019:
## Expected Behaviour
Hamlet should accept values for run-times when they are released by AWS.

## Current Behaviour
Hamlet developers must review and explicitly add new run-time values. While this ensures validation of values at template creation time, it imposes a maintenance burden and also means the use may enter a legitimate value as far as AWS is concerned but not supported as far as Hamlet is concerned. 

## Possible Solution
Remove values list and add a reference in the description of the attribute to the location in AWS of the current list.

## Context
Concern is this affects the first 30 minute experience of hamlet.

roleyfoley:
From #1986 

> I'm wondering if we need an option to provide a value set and flag a warning rather than an error if the value provided is outside the set. That way we don't stop new values being used, but do assist when the value has a typo.
> 
> So basically extending the current logic that checks for values to check if say an `AdditionalValues : true` attribute is present and set to true. In this case, allow a user defined value but warn of its use. We can still perform other checks such as type conformance.
> 
> I'm reflecting the pattern used in JSON schema where `additionalProperties : true` controls whether unspecified properties can be present on an object.

Yeah I like it!

PR #1986, opened 2022-05-11
feat(lambda): remove constraint on runtime

ken9ross:
## Intent of Change

## Description
Remove value list on runtime engine selection

## Motivation and Context
Fixes https://github.com/hamlet-io/engine/issues/1640

## How Has This Been Tested?
Locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/lambda/id.ftl
+++ b/providers/shared/components/lambda/id.ftl
@@ -78,22 +78,7 @@
             {
                 "Names" : "RunTime",
                 "Types" : STRING_TYPE,
-                "Values" : [
-                    "dotnet6",
-                    "dotnetcore1.0",
-                    "dotnetcore2.1",
-                    "go1.x",
-                    "java8",
-                    "java11",
-                    "nodejs12.x",
-                    "nodejs14.x",
-                    "python2.7",
-                    "python3.6",
-                    "python3.7",
-                    "python3.8",
-                    "python3.9",
-                    "ruby2.5"
-                ],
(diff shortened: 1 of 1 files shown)


ken9ross:
This seems so simple, that I don't understand raising a new request rather than doing it as part of #1638. Have I missed something in the requirement?

roleyfoley:
> This seems so simple, that I don't understand raising a new request rather than doing it as part of #1638. Have I missed something in the requirement?

The technical implementation is super simple the discussion on if we should do it not was intended for the ticket. That hasn't really happened yet.

ml019:
I'm wondering if we need an option to provide a value set and flag a warning rather than an error if the value provided is outside the set. That way we don't stop new values being used, but do assist when the value has a typo.

So basically extending the current logic that checks for values to check if say an `AdditionalValues : true` attribute is present and set to true. In this case, allow a user defined value but warn of its use. We can still perform other checks such as type conformance.

I'm reflecting the pattern used in JSON schema where `additionalProperties : true` controls whether unspecified properties can be present on an object.

roleyfoley:
> I'm wondering if we need an option to provide a value set and flag a warning rather than an error if the value provided is outside the set. That way we don't stop new values being used, but do assist when the value has a typo.
> 
> So basically extending the current logic that checks for values to check if say an `AdditionalValues : true` attribute is present and set to true. In this case, allow a user defined value but warn of its use. We can still perform other checks such as type conformance.
> 
> I'm reflecting the pattern used in JSON schema where `additionalProperties : true` controls whether unspecified properties can be present on an object.

Yeah I like it!

roleyfoley:
Closing this PR in favour of completing the conversion in https://github.com/hamlet-io/engine/issues/1640 which could involve changes to the engine instead.

PR #1988, opened 2022-05-16
feat(logstore): add dedicated log storage component

roleyfoley:
## Intent of Change

## Description

- Defines the log store component used to create an independent log store from other components 

## Motivation and Context

This allows for the creation of a log store that is written to by another component with content that doesn't relate to the execution of that content. 

This could be something like a lambda function which writes log events it receives, so the lambda log group is for running the function and capturing issues in running the function, where the log store is the formatted event content it was processing. 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/logstore/id.ftl
+++ b/providers/shared/components/logstore/id.ftl
@@ -12,6 +12,12 @@
     attributes=
         [
             {
+                "Names" : "Engine",
+                "Description": "The service used to store the logs",
+                "Types" : [ STRING_TYPE ],
+                "Values" : []
+            },
+            {
                 "Names" : "Links",
                 "SubObjects" : true,
                 "AttributeSet" : LINK_ATTRIBUTESET_TYPE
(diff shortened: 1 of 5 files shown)


providers/shared/components/logstore/id.ftl
```diff
@@ -0,0 +1,59 @@
...
+[@addComponent
+    type=LOGSTORE_COMPONENT_TYPE
+    properties=
+        [
+            {
+                "Type" : "Description",
+                "Value" : "Defines a standalong log store used to capture function base logs"
+            }
+        ]
+    attributes=
```
ml019, line 12: Should this have an engine attribute? Feels like something where the implementations will vary considerably.

ml019, line 12: Should this have an engine attribute? Feels like something where the implementations will vary considerably.

ml019 (approved):

Issue #140, opened 2018-03-04
Notifications for RDS events


roleyfoley:
RDS has the ability to provide SNS notifications when events occur on an RDS cluster, this includes events like failovers, patching, reboots etc. 
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_Events.html 

This could be integrated with KTLG if it supports forwarding events via an HTTP call or post.

roleyfoley:
Related to #139 
This would assist with know that is going on with RDS events when the occur and ensure that people are notified of events

roleyfoley:
This ticket is to discover the configuration required in RDS to enable the sending of these events to a supported destination most likely SNS which aligns with our existing alerts support in hamlet

PR #1991, opened 2022-05-17
Add RDS event support

ken9ross:
## Intent of Change

## Description
Add SystemNotifications attribute to DB component

## Motivation and Context
Part of https://github.com/hamlet-io/engine/issues/140

## How Has This Been Tested?
Locally, deployed to AWS, testcases established

### Prerequisite PRs:
- None

### Dependent PRs:

- None

### Consumer Actions:

- None


providers/shared/attributesets/event/id.ftl
```diff
@@ -0,0 +1,38 @@
...
+
+[@addAttributeSet
+    type=EVENT_ATTRIBUTESET_TYPE
+    properties=[
+        {
+                "Type"  : "Description",
+                "Value" : "Configuration of reported system notifications from component resources"
+        }]
+    attributes=[
+        {
```
roleyfoley, line 11: These are very specific to DB components. I think it would be better that this was on the DB component. If you think it should be an attribute set. EVENT_ATTRIBUTE_SET_TYPE is to generic. 

Maybe DB_SYSTEM_EVENT_ATTRIBUTE_TYPE would be better.

roleyfoley (changes requested):

providers/shared/attributesets/event/id.ftl
ken9ross: Done

providers/shared/attributesets/event/id.ftl
```diff
@@ -0,0 +1,38 @@
...
+                "low storage", 
+                "maintenance", 
+                "notification", 
+                "read replica", 
+                "recovery", 
+                "restoration", 
+                "security"
+            ]
+        },
+        {
```
roleyfoley, line 32: Is this optional? wouldn't default be all of them?

roleyfoley (changes requested):

providers/shared/attributesets/event/id.ftl
ken9ross: Yes on both counts

roleyfoley: So should we just make the default all of them? 
To me an undefined value would mean none not all.

ken9ross: The AWS processing is an absent Categories attribute means all events (presumably to allow for them adding new ones) ie it's a filtering attribute. If we default the value, we lose the "all events even ones not yet defined" status. To have no events processed, you'd need to remove the SystemNotification attribute level. Thoughts?

roleyfoley: Maybe an _all option as the default. Then when you pass that to AWS it is sent as the empty string. 
Probably also need a check that someone hasn't set _all and other values.

ken9ross: For that to make sense, we'd need to make Categories mandatory, correct? So the only way to get no Categories attribute is to specify the "_all" value?

roleyfoley: Include _all in the values ad add a Default property on the account. 
That would mean that its always _all by default and then you can override that with specific categories as required.

ken9ross: Or just to have left it as an empty attribute not passed through, but done

roleyfoley: Then it wouldn't show in documentation. Trying to get the understanding of what it does without making people read the code.

PR #1995, opened 2022-05-21
refactor(apigateway): authorization models

ml019:
## Intent of Change

## Description
Support additional model values for the case of IP filtering in combination with a lambda authorizer. Also rename the config attribute to more correctly reflect its purpose in controlling authorization rather than authentication.

## Motivation and Context
When used with a lambda authorizer, the default value of "IP" incorrectly provides an explict ALLOW  rather than relying on it to come from the policy provided by the authorizer. By providing explicit values to be used with the authorizer, the configuration can be validated as appropriate.

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/apigateway/id.ftl
+++ b/providers/shared/components/apigateway/id.ftl
@@ -107,9 +107,7 @@ object.
                 "Types" : STRING_TYPE,
                 "Values" : [
                     "IP",
-                    "SIG4ORIP", "SIG4_OR_IP",
                     "AUTHORISER_OR_IP", "AUTHORIZER_OR_IP",
-                    "SIG4ANDIP", "SIG4_AND_IP",
                     "AUTHORISER_AND_IP", "AUTHORIZER_AND_IP"
                 ],
                 "Default" : "IP"
--- a/providers/shared/components/apigateway/id.ftl
+++ b/providers/shared/components/apigateway/id.ftl
@@ -102,9 +102,15 @@ object.
                 "Default" : []
             },
             {
-                "Names" : "Authentication",
+                "Names" : ["AuthorisationModel", "AuthorizationModel", "Authentication"],
                 "Types" : STRING_TYPE,
-                "Values" : ["IP", "SIG4ORIP", "SIG4ANDIP"],
+                "Values" : [
+                    "IP",
+                    "SIG4ORIP", "SIG4_OR_IP",
+                    "AUTHORISER_OR_IP", "AUTHORIZER_OR_IP",
+                    "SIG4ANDIP", "SIG4_AND_IP",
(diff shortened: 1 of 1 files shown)


providers/shared/components/apigateway/id.ftl
```diff
@@ -102,9 +102,15 @@ object.
                 "Default" : []
             },
             {
-                "Names" : "Authentication",
+                "Names" : ["AuthorisationModel", "AuthorizationModel", "Authentication"],
```
roleyfoley, line 105: Should probably include a description if we are making things easier to follow

providers/shared/components/apigateway/id.ftl
```diff
@@ -102,9 +102,15 @@ object.
...
                 "Types" : STRING_TYPE,
-                "Values" : ["IP", "SIG4ORIP", "SIG4ANDIP"],
+                "Values" : [
+                    "IP",
+                    "SIG4ORIP", "SIG4_OR_IP",
```
roleyfoley, line 109: Should SIG4 move to the aws provider as an extension attribute? Its very specific to AWS

roleyfoley (changes requested):

providers/shared/components/apigateway/id.ftl
ml019: Probably but it has been like that so long removing it would require a breaking change.

providers/shared/components/apigateway/id.ftl
ml019: Will do

ml019: Done

providers/shared/components/apigateway/id.ftl
ml019: I could make the others I'm adding AWS specific and leave the unhyphenated ones as is?

ml019: I've moved the new `SIG4_*` values to the AWS provider and prefixed them. We can plan to remove the shared values in a future breaking change - https://github.com/hamlet-io/engine/issues/1997

roleyfoley: You can also add shared:<value> in the aws provider to keep them using the existing value while also adding the prefixed version 

https://github.com/hamlet-io/engine-plugin-aws/blob/bb0c68eb47dd6bc8f275808979d655af1fb2685e/aws/components/ecs/id.ftl#L101 

Has an example of that for the awsvpc network mode on containers

roleyfoley (approved):

PR #2008, opened 2022-06-07
fix(dn): Test for missing domain name

### LINKED ISSUE #1665: [BUG] Handle missing domain names with error
## Expected Behaviour

When generating a template which includes certificate configuration and the specified domain can't be found an error should be raised that it can't be found and let me know what it was looking for 

## Current Behaviour

When generating a template with a domain specified on a component the following exception is raised 

```
13:50:00 (Info) Generating outputs:
13:50:06 [*] entrance: deployment | output: contract | subset: generationcontract
13:50:15 [*] entrance: deployment | output: script | subset: prologue
13:50:17 Apr 28, 2021 3:50:17 AM freemarker.log._JULLoggerFactory$JULLogger error
13:50:17 SEVERE: Error executing FreeMarker template
13:50:17 FreeMarker template error:
13:50:17 When calling function "splitDomainName", required parameter "name" (parameter #1) was specified, but had null/missing value.
13:50:17 
13:50:17 ----
13:50:17 Tip: If the parameter value expression on the caller side is known to be legally null/missing, you may want to specify a default value for it with the "!" operator, like paramValue!defaultValue.
13:50:17 ----
13:50:17 
13:50:17 ----
13:50:17 FTL stack trace ("~" means nesting-related):
13:50:17 	- Failed at: #function splitDomainName(name)  [in template "shared/components/name.ftl" in function "splitDomainName" at line 11, column 1]
```

## Possible Solution

When formatting the domain name validate that the values are present before trying to format the details

## Context

Creating and managing custom domains for components

ken9ross:
## Intent of Change

## Description
Raise non-stopping fatal error when domain name does not include at least one dot (returning empty array rather than erroring)

## Motivation and Context
Part of #1665 

## How Has This Been Tested?

Couldn't see how to set up a failure case to test - any advice

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


providers/shared/components/name.ftl
```diff
@@ -9,7 +9,12 @@
 [/#function]
 
 [#function splitDomainName name]
-    [#return name?split(".")]
+    [#if !name?contains(".") ]
+        [@fatal message="Missing domain name for API docs" context={ "splitDomainName" : name } /]
```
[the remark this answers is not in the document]
ken9ross: Fair call. Done

providers/shared/references/Certificate/id.ftl
```diff
@@ -54,7 +54,12 @@
...
-            [#return domain ]
+            [#if !domain.Name?contains(".") ]
+                [@fatal message="Missing or invalid primary domain name" context=domain /]
+                [#return {} ]
+            [#else]
+                [#return domain ]
+            [/#if]
             [#break]
         [/#if]
     [/#list]
```
roleyfoley, line 65: Again I think this isn't quite the right spot to perform this validation. 
This function is specifically for validating a primary domain, What about the other domains that are generated in the certificateObject? 

Probably best to do this validation when the certificates object is created. That way it applies whenever a domain name is formed.

roleyfoley (changes requested):

providers/shared/references/Certificate/id.ftl
ken9ross: I think this is now in the correct place

roleyfoley (approved):

PR #2009, opened 2022-06-10
feat: add support for dynamic value evaulation

roleyfoley:
## Intent of Change

## Description

Extends the dynamic value evaluation from solution configuration on runbooks to a more generic function which can be used across all components and their solution configuration 

Dynamic values can be used in solution configuration to represent a value that should be resolved at run time rather than an explicit configuration defined in the solution file. 

They can be used anywhere a string is accepted and can be used at any point in the string. They follow the format of 

`__<type>:<attributes>...__`

Where type is the type attribute assigned to a Dynamic Value Provider and the attributes section represents one or more attributes separated by `:` These attributes are used by the provider to determine the value to replace. When a replacement occurs all of the dynamic value from __ to __ ( inclusive ) are replaced with the resolved value. 

Supported dynamic value providers in this initial PR are 

- setting: looks for a setting key in the environment variable syntax for Settings and provides the setting value for the replacement 
- attribute: given the id of a link configured under the component Solution.Links Object and the name of an attribute of that component, return the attribute as the replacement
- input: used for runbooks only and returns a RunBookInput command line value for replacement
- output: used for rubooks only and is used to pass outputs between contract steps 

## Motivation and Context

There were a couple of motivations for adding support for this feature 

- Use runbook inputs on runbook solution configuration. With moving dynamic value provides up to the occurrence level it enables the use of inputs on the runbook solution configuration. This is useful when you might want to run a runbook across multiple components ( like emptying S3 buckets ), using this approach you can make the Component parameter of the link a dynamic input value and during the occurrence processing the value will be replaced. 
- Longer or complicated values can be stored as settings and referenced using a dynamic placeholder. Some components require details like long URLS, certificates or other long string values. With dynamic values these can be stored in a settings file and then updated at runtime

With the ability for providers to add their own dynamic value providers this could be used to substitute secrets encrypted values that should be decrypted or require specific reference configuration ( [AWS CF Dynamic references](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html) to be used whenever a user wants to use them

## How Has This Been Tested?

Tested locally and with additional test cases

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/provider.ftl
+++ b/engine/provider.ftl
@@ -358,7 +358,7 @@
                 /]
             [/#list]
 
-            [#-- Determine the tasks for the provider --]
+            [#-- Determine the dynamic value providers for the provider --]
             [#local directories =
                 internalGetPluginFiles(
                     [providerMarker.Path, "dynamicvalues"],
--- a/engine/provider.ftl
+++ b/engine/provider.ftl
@@ -358,6 +358,22 @@
                 /]
             [/#list]
 
+            [#-- Determine the tasks for the provider --]
+            [#local directories =
+                internalGetPluginFiles(
+                    [providerMarker.Path, "dynamicvalues"],
+                    [
+                        ["[^/]+"]
+                    ]
+                )
+            ]
+            [#list directories as directory]
+                [@internalIncludeTemplatesInDirectory
+                    directory,
+                    ["id", "name", "dynamicvalue"]
+                /]
+            [/#list]
+
             [#-- Determine the output writers for the provider --]
             [#local directories =
                 internalGetPluginFiles(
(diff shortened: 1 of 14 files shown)


engine/provider.ftl
```diff
@@ -358,6 +358,22 @@
                 /]
             [/#list]
 
+            [#-- Determine the tasks for the provider --]
```
ml019, line 361:
```suggestion
            [#-- Determine the dynamic value providers for the provider --]
```

ml019 (changes requested):

ml019:
@roleyfoley wondering why you dismissed the comment change I proposed? The comment looks like a cut/paste duplication.

roleyfoley:
Was a rebase with a force push must have missed that change syncing it locally

ml019 (approved):

PR #2018, opened 2022-06-22
feat(entrance): stackoutput entrance

roleyfoley:
## Intent of Change

## Description

- Adds the stackoutput entrance which generates a stack output file based on the provided stack output inputs
- Adds a task to call the stackoutput entrance from a runbook
- Adds the CMDB Root dir as a common input parameter

## Motivation and Context

This allows for runbooks to contribute stack outputs for components. This also supports the migration to deployment contracts in the future with the potential to collect all stack outputs generated during a deployment and write them to a single file. 

This would remove the need for multiple pseduo stack outputs and have a single stack output state file.

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/inputseeders/shared/id.ftl
+++ b/providers/shared/inputseeders/shared/id.ftl
@@ -109,6 +109,7 @@
                 [#-- Input data control --]
                 "Input" : {
                     "Source" : inputSource!"composite",
+                    "RootDir": (rootDir)!"",
                     "Filter" :
                         attributeIfContent("DistrictType", districtType!SEGMENT_DISTRICT_TYPE) +
                         attributeIfContent("Tenant", tenant!"") +
--- a/engine/output.ftl
+++ b/engine/output.ftl
@@ -635,6 +635,7 @@
     [#switch entrance ]
         [#case "deployment" ]
         [#case "deploymenttest" ]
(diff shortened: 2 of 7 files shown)


providers/shared/inputseeders/shared/id.ftl
```diff
@@ -109,6 +109,7 @@
                 [#-- Input data control --]
                 "Input" : {
                     "Source" : inputSource!"composite",
+                    "RootDir": (rootDir)!"",
```
ml019, line 112: Not sure this should live in the core of the engine - we added the input pipeline process partly to isolate such details in plugins like the cmdb plugin.

roleyfoley: I think until that has been fully incorporated into the engine we need to include this value. Otherwise its not possible to describe an entrance invocation as part of a contract.

ml019:
@roleyfoley I found the other PR which had the definition of the task type and merged it so not sure why this is still failing.

I did notice that maybe the task type name should have `_TASK_TYPE` on the end to match all the other definitions but that doesn't explain why this one is failing...

ml019 (approved):

PR #2034, opened 2022-08-29
feat: add support for base attribute set expansion

roleyfoley:
## Intent of Change

## Description

- Adds support for adding an attribute which only lists an attribute set. When evaluating this attribute it will be expanded and the attribute set will replace the attribute where it was listed
- Migrates all existing shared configuration defined in global vars to this approach

## Motivation and Context

This is useful when you have shared configuration that doesn't need to live under a child object and should instead be included at the base level of a configuration section. 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019 (commented):
All the usage of expandCompositeConfiguration() in the changes sets childSets to false. Given that the previous behaviour was only childSets, wondering how that is achieved e.g. use of the function without explicitly setting the attributes?

roleyfoley:
> All the usage of expandCompositeConfiguration() in the changes sets childSets to false. Given that the previous behaviour was only childSets, wondering how that is achieved e.g. use of the function without explicitly setting the attributes?

OOps! that's the wrong way around, dang positional arguments. WIll fix that up.

roleyfoley:
> > All the usage of expandCompositeConfiguration() in the changes sets childSets to false. Given that the previous behaviour was only childSets, wondering how that is achieved e.g. use of the function without explicitly setting the attributes?
> 
> OOps! that's the wrong way around, dang positional arguments. WIll fix that up.

No sorry it is the right way, I don't want it to expand the child objects during expansion of attributes in extend because doing that would break the schema generation that uses refs across to the different attribute sets.

ml019:
OK - so how is the current behaviour maintained? I'm assuming there is another call to `expandCompositeConfiguration()` somewhere that hasn't been touched so the attribute defaults will apply including handling baseSets?

roleyfoley:
> OK - so how is the current behaviour maintained? I'm assuming there is another call to `expandCompositeConfiguration()` somewhere that hasn't been touched so the attribute set expansion will apply based on the attribute defaults (including handling baseSets)?

Yep thats right.

PR #2046, opened 2022-09-29
fix: tier network configuration

ml019:
## Intent of Change

## Description
A network must now be explicitly enabled on a tier rather than defaulting to being enabled.

The `external` tier has also been added to the master data given its common use in solutions.

## Motivation and Context
The new behaviour aligns with the behaviour prior to the addition of the `Solutions` feature. It should not 
cause a breaking change as the network is explicitly enabled/disabled for all tiers except the `external` tier.

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


providers/shared/solution/tiers/solution.ftl
```diff
@@ -26,7 +26,7 @@
                         {
                             "Names" : "Enabled",
                             "Types" : BOOLEAN_TYPE,
-                            "Default" : true
```
roleyfoley, line 29: I think we should keep this as the updated default to true, otherwise it adds a collection of steps to get networking going, for anyone enabling a network they would just expect it to create them.

roleyfoley (changes requested):

providers/shared/solution/tiers/solution.ftl
ml019: But don't you need to explicitly add the network link and the route and ACL config?

roleyfoley: Even more reason then not to have another thing I have to do...

ml019: There is also the tier order and network tier order to adjust so it is not an inconsequential addition. If we leave the enable value as true by default, then whether you want a network for the tier or not, either way you need to add some extra explicit config.

How about I switch to it being true, but I might add a fatal error if the tier is enabled and `Tier` and `Component` attributes are missing in the link with a message flagging what is needed in each case (i.e. if you want a network, link, acl and route table need to be explicitly added to the tier, if no network is required, explicitly disable it)?

ml019: Probably a reasonable check to add to the network component anyway...

roleyfoley: The orders are actually no longer required, there is some sneaky stuff in the newer solution level config that replaces it. So you just need to set a unique index number for the tier.

PR #2049, opened 2022-10-02
feat(datacatalog): add new component

roleyfoley:
## Intent of Change

## Description

- Adds the datacatalog component which represents the idea of a catalog of data that has been normalised using a datastore process
- Currently includes tables for column based data representation

## Motivation and Context

Data catalogs can be used to map data from different sources into a centralised repository. The initial implementation aims at adding AWS Glue into hamlet. 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/datacatalog/id.ftl
+++ b/providers/shared/components/datacatalog/id.ftl
@@ -1,0 +1,266 @@
+[#ftl]
+
+[@addComponent
+    type=DATACATALOG_COMPONENT_TYPE
+    properties=
+        [
+            {
+                "Type" : "Description",
+                "Value" : "Defines a catalog of data sources"
+            },
+            {
+                "Type" : "Providers",
+                "Value" : [ "aws" ]
+            }
+        ]
+    attributes=
+        [
+            {
+                "Names" : "Links",
+                "SubObjects" : true,
+                "AttributeSet" : LINK_ATTRIBUTESET_TYPE
+            }
+        ]
+/]
+
+[@addComponentDeployment
+    type=DATACATALOG_COMPONENT_TYPE
+    defaultGroup="solution"
+/]
+
+
+[@addChildComponent
+    type=DATACATALOG_TABLE_COMPONENT_TYPE
+    parent=DATACATALOG_COMPONENT_TYPE
+    childAttribute="Tables"
+    linkAttributes=["Table"]
+    properties=
+        [
+            {
+                "Type"  : "Description",
+                "Value" : "A Tabular data store in the catalog"
+            }
+        ]
+    attributes=
+        [
+            {
+                "Names" : "Links",
+                "SubObjects" : true,
+                "AttributeSet" : LINK_ATTRIBUTESET_TYPE
+            },
+            {
+                "Names" : [ "Extensions", "Fragment", "Container" ],
+                "Description" : "Extensions to invoke as part of component processing",
+                "Types" : ARRAY_OF_STRING_TYPE,
+                "Default" : []
+            },
+            {
+                "Names" : "Parameters",
+                "Description" : "Additional key/value parameters to configure the table - Key of the object is used as key by default",
+                "SubObjects" : true,
+                "Children": [
+                    {
+                        "Names": "Enabled",
+                        "Default" : "Include the parameter",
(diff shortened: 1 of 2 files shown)


providers/shared/components/datacatalog/id.ftl
```diff
@@ -0,0 +1,266 @@
...
+                            },
+                            {
+                                "Names" : "Type",
+                                "Description": "The data type of the column",
+                                "Types": STRING_TYPE,
+                                "Default" : ""
+                            },
+                            {
+                                "Names": "Description",
+                                "Description": "A descrpition of the column",
```
ml019, line 132:
```suggestion
                                "Description": "A description of the column",
```

providers/shared/components/datacatalog/id.ftl
```diff
@@ -0,0 +1,266 @@
...
+                    }
+                ]
+            },
+            {
+                "Names": "Layout",
+                "Description" : "The layout of the data in the table",
+                "Children" : [
+                    {
+                        "Names": "Partitioning",
+                        "Description" : "How data is partitioned within the layout - name is the key of the object by default",
```
ml019, line 109: Name is a mandatory attribute?

providers/shared/components/datacatalog/id.ftl
```diff
@@ -0,0 +1,266 @@
...
+                                "Names": "Description",
+                                "Description": "A descrpition of the column",
+                                "Types": STRING_TYPE,
+                                "Default" : ""
+                            }
+                        ]
+                    },
+                    {
+                        "Names": "Columns",
+                        "Description" : "The base columns in the data article - name is the key of the object by default",
```
ml019, line 140: Same comment re Names?

providers/shared/components/datacatalog/id.ftl
```diff
@@ -0,0 +1,266 @@
...
+                        ]
+                    }
+                ]
+            },
+            {
+                "Names": "Format",
+                "Description" : "Defines the format of the data",
+                "Children": [
+                    {
+                        "Names": "SerDe",
```
ml019, line 176: Name a bit cryptic - maybe just use "Serialisation". I know it doesn't explicitly mention deserialisation but the two are normally referred to together anyway...

providers/shared/components/datacatalog/id.ftl
```diff
@@ -0,0 +1,266 @@
...
+                "Children" : [
+                    {
+                        "Names" : "Enabled",
+                        "Description": "Enable the use of a Crawler for the table",
+                        "Types" : BOOLEAN_TYPE,
+                        "Default": false
+                    },
+                    {
+                        "Names": "Schedule",
+                        "Description" : "The schedule to run the crawler on",
```
ml019, line 233: CRON format? Maybe add a bit more detail in the description

ml019 (changes requested):

providers/shared/components/datacatalog/id.ftl
roleyfoley: Yeah fair call, I was a bit confused at first but it does seem to be the indurstry term. However Serialisation does cover both sides I think. Updated across the PR's

providers/shared/components/datacatalog/id.ftl
roleyfoley: Fixed up, I said that it should be in AWS rate format which should cover it.

providers/shared/components/datacatalog/id.ftl
```diff
@@ -0,0 +1,264 @@
...
+                "Children" : [
+                    {
+                        "Names" : "Enabled",
+                        "Description": "Enable the use of a Crawler for the table",
+                        "Types" : BOOLEAN_TYPE,
+                        "Default": false
+                    },
+                    {
+                        "Names": "Schedule",
+                        "Description" : "The schedule to run the crawler on - Format should in in the AWS Rate format",
```
ml019, line 231:
```suggestion
                        "Description" : "The schedule to run the crawler on - Format should in the AWS Rate format",
```

ml019 (changes requested):

PR #2065, opened 2022-10-21
fix(waf): set default WAFProfile to default

roleyfoley:
## Intent of Change

## Description

- Adds a default WAF profile which allows for open access by default
- Aligns with other security profiles and allows for creating rules without the provided OWASP rulesets

## Motivation and Context

This aligns with the updated behaviour in https://github.com/hamlet-io/engine-plugin-aws/pull/668 which brings the WAF security configuration more in line with how the rest of the profiles work across other components. In the AWS provider if the OWASP WAF configuration isn't enable the user could not use the profile to configure the WAF rules. Creating a default open profile and using that as the default aligns with the current behaviour in AWS 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- https://github.com/hamlet-io/engine-plugin-aws/pull/668 

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/inputseeders/shared/masterdata.json
+++ b/providers/shared/inputseeders/shared/masterdata.json
@@ -1800,6 +1800,10 @@
     }
   },
   "WAFProfiles": {
+    "default": {
+      "Rules": [],
+      "DefaultAction": "ALLOW"
+    },
     "OWASP2017": {
       "Rules": [
         {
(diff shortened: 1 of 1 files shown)


ml019:
This is going to break things if I understand correctly? A project would need to explicitly select a profile for OWASP whereas previously it would have used 2017 by default if OWASP was enabled?

roleyfoley:
It shouldn't require changes for anyone who has enabled the OWASP as the changes in the AWS provider in https://github.com/hamlet-io/engine-plugin-aws/pull/668 handle the use of the OWASP Configuration settings. 

My thoughts would be to deprecate the OWASP option as its a limited version of the SecurityProfile option anyway.

ml019 (approved):

PR #2076, opened 2022-12-08
fix(account_s3): remove s3 registry initialisation

roleyfoley:
## Intent of Change

## Description

- Remove the setup of registry markers in the account registry bucket
- Removes the code bucket sync

## Motivation and Context

With the move to fixed registry prefixes to store images the need to initialise the registry isn't really there any more. This is currently causing issues as the s3RegistryIds variable was removed in other image updates
We no longer have components which rely on the code sync process so this has been removed as well

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/legacy/account/account_s3.ftl
+++ b/legacy/account/account_s3.ftl
@@ -268,17 +268,7 @@
                     [
                         "function sync_code_bucket() {"
                     ] +
-                        scriptSyncContent +
-                    [
-                        "}",
-                        "#",
-                        "case $\{STACK_OPERATION} in",
-                        "  create|update)",
-                        "    sync_code_bucket || return $?",
-                        "    initialise_registries ${s3RegistryIds} || return $?",
-                        "    ;;",
-                        " esac"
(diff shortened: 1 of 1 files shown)


legacy/account/account_s3.ftl
```diff
@@ -268,17 +268,7 @@
...
-                        "}",
-                        "#",
-                        "case $\{STACK_OPERATION} in",
-                        "  create|update)",
-                        "    sync_code_bucket || return $?",
-                        "    initialise_registries ${s3RegistryIds} || return $?",
-                        "    ;;",
-                        " esac"
-                    ]
+                        scriptSyncContent
```
RossMurr4y, line 271: It's been a while so I may be reading this wrong, but it looks like `scriptSyncContent` will need a `}` somewhere in there? 

> `scriptSyncContent + [ "}" ]`

PR #2082, opened 2023-02-14
feat(account): remove bootstrap credential buckets and enable TLS for s3

roleyfoley:
## Intent of Change

## Description

- Removes the credentials and code buckets from the account level s3 deployments
- Removes the ScriptStore reference data type which used the code buckets and is no longer required
- Adds bucket policy on all S3 account buckets to require HTTPS connections to s3

## Motivation and Context

- Removes buckets which are rarely used and can cause concerns ( especially with a bucket called credentials)
- Reduces the various features for Ec2 based tasks that have been replaced by compute tasks
- Improves security stance from a reporting and auditing side.

## How Has This Been Tested?

Tested locally and with deployment

### Prerequisite PRs:

- https://github.com/hamlet-io/engine-plugin-aws/pull/697

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019:
@roleyfoley the only use of the code bucket was to support ecs ec2 instances for things like associating them to a cluster. Has that all moved to runbooks?

roleyfoley:
> @roleyfoley the only use of the code bucket was to support ecs ec2 instances for things like associating them to a cluster. Has that all moved to runbooks?

Yeah all of the ec2 based configuration was moved into compute tasks as part of the introduction. The main reason for that was to make sure the ec2 instance setup was isolated as updating the old bootstrap scripts would cause issues if templates hadn't been updated in some cases

PR #2090, opened 2023-03-01
refactor(waf): remove version control option

roleyfoley:
## Intent of Change

## Description

- Removes setting the version on WAF configuration

## Motivation and Context

Reduces the configuration and maintenance required for working with WAF. This probably shouldn't have been included in the core in the first place as the verisons were AWS only. 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/attributesets/waf/id.ftl
+++ b/providers/shared/attributesets/waf/id.ftl
@@ -14,13 +14,6 @@
             "Default" : true
         },
         {
-            "Names" : "Version",
-            "Description" : "Version of WAF to use",
-            "Types"  : STRING_TYPE,
-            "Values" : [ "v1","v2"],
-            "Default" : "v1"
-        },
-        {
             "Names" : "IPAddressGroups",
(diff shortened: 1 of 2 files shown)


ml019:
Given they've changed it once before, should we simply leave the version attribute in place and make the only valid value and the default `v2`?

The simplified code is good so I guess it just depends on whether we think it will change again any time soon...

roleyfoley:
> Given they've changed it once before, should we simply leave the version attribute in place and make the only valid value and the default `v2`?
> 
> The simplified code is good so I guess it just depends on whether we think it will change again any time soon...

I really doubt it, they are continuously pushing new features into it and can't see any sign of it going into the quiet death march that they give aws services. If it does come back then it should move out of the engine and into a vendor specific option anyway.

ml019 (approved):

PR #2104, opened 2023-05-05
refactor(s3): object ownership support

ml019:
## Intent of Change

## Description
Add support for control of the object ownership for the bucket.

Also, per https://docs.aws.amazon.com/AmazonS3/latest/userguide/enable-server-access-logging.html, switch from using an ACL to using bucket policy to permit logging to the audit bucket.

## Motivation and Context
- align with AWS best practices
- avoid deployment failures where an SCP enforces bucket owner object ownership.
- 
## How Has This Been Tested?
Local generation. Use in customer site will also be undertaken once part of the unicycle release.

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/s3/id.ftl
+++ b/providers/shared/components/s3/id.ftl
@@ -220,6 +220,12 @@
                         "Default" : "Daily"
                     }
                 ]
+            },
+            {
+                "Names" : "ObjectOwnership",
+                "Values" : [ "owner", "ownerpreferred", "writer"],
+                "Types" : STRING_TYPE,
+                "Description" : "Explicitly control the way in which object ownership should be treated"
             }
         ]
 /]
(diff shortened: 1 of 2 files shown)


providers/shared/components/s3/id.ftl
```diff
@@ -220,6 +220,12 @@
                         "Default" : "Daily"
                     }
                 ]
+            },
+            {
+                "Names" : "ObjectOwnership",
```
roleyfoley, line 225: Do we need a default here?

roleyfoley (changes requested):

providers/shared/components/s3/id.ftl
ml019: I was leaving it off because I didn't want to trigger adding config for existing deployments. In reality, the default behaviour is "writer" so if you think it is better, we could always add the object ownership config, in which case we'd add that as the default.

roleyfoley (approved):

PR #2106, opened 2023-05-15
feat: private endpoint services

ml019:
## Intent of Change

## Description
Extend the processing of endpoint services to support privately defined endpoints, as opposed to the public ones offered by cloud providers. 

These endpoints are assumed to be required in all zones supported by the network of the tier in which the gateway is defined.

## Motivation and Context
Support the use of customer defined service endpoints in AWS.

## How Has This Been Tested?
Local template generation. I'll also apply on customer site.

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/engine/common.ftl
+++ b/engine/common.ftl
@@ -416,20 +416,17 @@
 [/#function]
 
 [#function getNetworkEndpoints endpointGroups zone region ]
-    [#local services = []]
+    [#local services = [] ]
+    [#local allZoneServiceEndpoints = {} ]
     [#local networkEndpoints = {}]
 
     [#local regionObject = getRegions()[region]]
     [#local zoneNetworkEndpoints = (regionObject.Zones[zone].NetworkEndpoints)![] ]
 
     [#list endpointGroups as endpointGroup ]
-        [#if networkEndpointGroups[endpointGroup]?? ]
-            [#list networkEndpointGroups[endpointGroup].Services as service ]
-                [#if !services?seq_contains(service) ]
-                    [#local services += [ service ]]
-                [/#if]
-            [/#list]
-        [/#if]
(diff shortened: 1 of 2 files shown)


ml019:
Not sure if this is the best way to handle this, or whether I should instead use links from a gateway destination to an external service component which has the serviceName as an attribute. Maybe both make sense - thoughts welcome.

roleyfoley:
You can include private ones in the current implementation. 

Under the region config

```
    "Regions": {
        "ap-southeast-2": {
            "Locality": "Sydney",
            "Zones": {
                "a": {
                    "NetworkEndpoints": [
                        {
                            "Type": "Interface",
                            "ServiceName": "com.amazonaws.vpce.ap-southeast-2.vpce-svc-abc123"
                        }
```


Then under the network endpoint groups 

```
    "NetworkEndpointGroups": {
        "privateService": {
            "Services": [
                "com.amazonaws.vpce.ap-southeast-2.vpce-svc-abc123"
            ]
        },
```

Then when creating the gateway endpoint you can specifiy privateService like you would for the storage groups etc.

ml019:
Yeah - I didn't want to fiddle with the regional data as that naturally comes from the masterdata for the cloud provider. I looked at doing it the way you suggested initially, but the arrays and the overriding of provider data felt wrong,.

The other thing I wanted to avoid what having to do it for every zone in the regional data so basically always using every zone in the solution. For VPC endpoints associated with customer defined service I think that will generally be the case, and if not, then the fallback is the way you suggest.

ml019:
@roleyfoley  are we good to merge this or do you still have concerns? I'm at the point where I want to test it with the customer.

ml019:
BTW I suspect we may have wait issues with customer endpoint services as connections need to be accepted. I want to get to the point of being able to try one to work out next steps.

PR #2108, opened 2023-05-17
refactor(waf): remove rule tuples and add error

roleyfoley:
## Intent of Change

## Description

- Removes the shared tuple lookups as they were only used for AWS and aren't required anymore
- Add error for missing rule in WAF profile

## Motivation and Context

Removes extra processing that isn't required in the shared provider 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


providers/shared/references/WAFRule/reference.ftl
```diff
@@ -29,97 +29,3 @@
         } ]
     [#return result]
 [/#function]
-
-[#function formatWAFByteMatchTuples filter={} valueSet={} ]
```
ml019, line 33: Are these all WAF things? Or just code refactoring that has rendered them unnecessary?

roleyfoley: They were left overs from removing WAF v1 support. They were missed as they weren't obvious in the first round of cleanup

ml019 (approved):

PR #2115, opened 2023-07-01
feat(apigateway): private APIs

ml019:
## Intent of Change

## Description
Add the option of private APIs. For these APIS, traffic filtering is done on the basis of VPC endpoints rather than IP addresses. The VPC endpoints are assumed to be configured via inbound links to external services, which should provide a VPC_ENDPOINTS attribute containing a command separated list of vpc endpoint names.

Adjust the available values for the authorisation model to support either IPs or VPC Endpoints.

## Motivation and Context
- support new AWS feature
- support more security conscious deployment

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/providers/shared/components/apigateway/id.ftl
+++ b/providers/shared/components/apigateway/id.ftl
@@ -83,6 +83,7 @@ object.
             },
             {
                 "Names" : "Links",
+                "Description" : "For a private API, provide VPC Endpoints via an inbound link from an external service with a \"VPC_ENDPOINTS\" attribute",
                 "SubObjects" : true,
                 "AttributeSet" : LINK_ATTRIBUTESET_TYPE
             },
@@ -93,7 +94,7 @@ object.
             {
                 "Names" : "EndpointType",
                 "Types" : STRING_TYPE,
-                "Values" : ["EDGE", "REGIONAL"],
+                "Values" : ["EDGE", "REGIONAL", "PRIVATE"],
                 "Default" : "EDGE"
             },
             {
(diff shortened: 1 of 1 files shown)


providers/shared/components/apigateway/id.ftl
```diff
@@ -93,24 +93,25 @@ object.
             {
                 "Names" : "EndpointType",
                 "Types" : STRING_TYPE,
-                "Values" : ["EDGE", "REGIONAL"],
+                "Values" : ["EDGE", "REGIONAL", "PRIVATE"],
                 "Default" : "EDGE"
             },
             {
-                "Names" : "IPAddressGroups",
+                "Names" : ["TrafficSources", "IPAddressGroups", "VPCEndpoints"],
```
roleyfoley, line 100: How do I define VPC endpoints? based on their aws arn/id? Given IP address groups work off the abstraction within hamlet we probably either need to say how to use it or figure out what the abstraction should be in this case?

roleyfoley (changes requested):

providers/shared/components/apigateway/id.ftl
ml019: It is just a DNS name so a string. I played with having a separate entry for them but all the policy logic works pretty well the same whether it is IP or VPC endpoint or overall it kind of felt like the same thing. 

You are correct that the abstraction is a bit different, which is why I allowed for the use of the explicit name depending on what the situation was.

Do you think it deserves its own specific config attribute? Happy to switch to that if you think we should. Technically you could IP whitelist as well as endpoint whitelist, but given the way the private APIs work, I can't see much call for that.

ml019: Reverted this change and now using inbound links from external service to define consuming vpc endpoints

roleyfoley (approved):

